PR #4, opened 2015-01-30
Random cleanup

keyneston:
This branch is all sorts of changes. Some are minor, hygienic, and others are things needed in order to handle changes slack has made in their api.

### DIFF
--- a/channels.go
+++ b/channels.go
@@ -1,7 +1,6 @@
 package slack
 
 import (
-	"errors"
 	"net/url"
 	"strconv"
 )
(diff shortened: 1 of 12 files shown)


nlopes:
@tarrant I don't particularly like/agree with https://github.com/tarrant/slack/commit/8ee19fedf85fd5a9d29958cf65db2cd2abadb7ed

I know golint complains about it but I am not convinced it should be uppercase. 
I will gladly accept the commits if you remove that one.

Thanks for the cleanup, much appreciated!

keyneston:
I've removed the ID change. I find the standard for initialisms being consistently capitalized or not makes for cleaner functions but I'm happy to conform to local customs.

Thanks for your quick responses, this is an awesome codebase. It has been a pleasure both writing against it and making changes to the code base.

I hope that I have the opportunity to continue to contribute, but we will see which way the winds blow in life and work.

PR #6, opened 2015-03-02
Fix data polution problem in the events processing

tcolgate:
The second patch fixes a seious problem where events end up with old event data being mixed in with new event data.


nlopes:
Thanks!

PR #8, opened 2015-04-17
Fix PostMessage not taking AsUser into account

Wiwiweb:
Seems like AsUser was meant to be turned into json when PostMessageParameters gets Marshaled, however the PostMessage() method doesn't Marshall and instead creates it's own array of URL values.

I've tried fixing this the simplest way possible. I'm new to Go so please feel free to point out stuff I missed.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -129,6 +129,9 @@ func (api *Slack) PostMessage(channelId string, text string, params PostMessageP
 	if params.Username != DEFAULT_MESSAGE_USERNAME {
 		values.Set("username", string(params.Username))
 	}
+	if params.AsUser != false {
+		values.Set("as_user", "true")
+	}
 	if params.Parse != DEFAULT_MESSAGE_PARSE {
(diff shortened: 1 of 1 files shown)


chat.go
```diff
@@ -129,6 +129,9 @@ func (api *Slack) PostMessage(channelId string, text string, params PostMessageP
 	if params.Username != DEFAULT_MESSAGE_USERNAME {
 		values.Set("username", string(params.Username))
 	}
+	if params.AsUser != false {
```
nlopes: It would be nice to have this false defined as a DEFAULT_MESSAGE_ASUSER. the rest looks good. Thanks for the patch

Wiwiweb:
How does it look now?

I removed the json tag of AsUser since it didn't seem to be used.

nlopes:
Looks good.

Wiwiweb:
Cheers! :)

PR #9, opened 2015-04-19
Add markdown support

dojineko:
Add markdown flag to PostMessageParameters and Attachment.
https://api.slack.com/docs/formatting#message_formatting


nlopes:
Looks good.

patcon:
Anyone know if this assumes the slackdown (alternative markdown that slack uses)?

If so, we should probably rename it to slackdown, which seems to be the preferred term :)

If there's agreement, I am happy to create a PR!

PR #10, opened 2015-04-19
Fix bug in merge #8 from wrong conversion of bool to string

Wiwiweb:
Actually I messed up, converting a bool to string is not as simple as `string(bool)`. Terribly sorry about that :(

I ended up just copying the way you did it for `UnfurlLinks` and `UnfurlMedia` which is to just hardcode write the opposite bool value of the default.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -132,7 +132,7 @@ func (api *Slack) PostMessage(channelId string, text string, params PostMessageP
 		values.Set("username", string(params.Username))
 	}
 	if params.AsUser != DEFAULT_MESSAGE_ASUSER {
-		values.Set("as_user", string(params.AsUser))
+		values.Set("as_user", "true")
 	}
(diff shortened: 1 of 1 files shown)


mediocregopher:
+1

kentaro:
:+1:

dojineko:
:+1:

yuya-takeyama:
:+1:

kyokomi:
:+1:

bfontaine:
It’d be great to have this merged, since the bug breaks the build and thus prevents people from installing the library:

```
$ go get github.com/nlopes/slack
# github.com/nlopes/slack
../../nlopes/slack/chat.go:135: cannot convert params.AsUser (type bool) to type string
```

nlopes:
Glorious vacations without cell reception :)
/merged

PR #11, opened 2015-04-29
Add escapeMessage option for PostMessage and UpdateMessage

Wiwiweb:
It is not very clear from [the formatting docs](https://api.slack.com/docs/formatting), but bots can add hyperlinks and special commands by sending strings such as `<http://www.foo.com|this is a link>` or `<!channel>`.

However by default, PostMessage and UpdateMessage escape these characters, which makes the bot send literally the characters `<` and `>`.

I added an option to make this escaping optional so that bots can send links and commands again. For PostMessage the escaping is on by default, as previous behavior. For UpdateMessage I had to add a parameter to the function.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -182,14 +182,11 @@ func (api *Slack) PostMessage(channelId string, text string, params PostMessageP
 }
 
 // UpdateMessage updates a message in a channel
-func (api *Slack) UpdateMessage(channelId, timestamp, text string, escape bool) (string, string, string, error) {
-	if escape {
-		text = escapeMessage(text)
-	}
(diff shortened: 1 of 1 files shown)


Wiwiweb:
I've removed the changes from UpdateMessage. It turns out the `chat.update` API call doesn't support formatted links, so this change of code was useless.

A quote from Slack support:

> Only the `chat.postMessage` API method and incoming webhooks support formatted links at the moment. Clients and the RTM API do not, though it's possible we'll add this capability in the future

nlopes:
Looks good. And this one isn't a breaking change. Merging.

PR #12, opened 2015-04-30
messages: ReplyTo is a string timestamp, not an int

bpowers:
This is different from the reply_to on websocket ACKs - reply_to in a
message refers to the globally-unique message timestamp of another
message represented as a string.  I don't see this explicit in the
docs, but I get unmarshal errors on a specific object:

{
    "reply_to": "1430355230.67",
    "type": "message",
    "channel": "D04JWB1TW",
    "user": "U03A807KY",
    "text": "...",
    "ts": "1430355230.000009"
}

FWIW this message came from a client using the IRC bridge on a "direct
message" channel - could be why other people don't seem to be
experiencing this.

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -33,7 +33,7 @@ type Msg struct {
 	Hidden           bool         `json:"bool,omitempty"`
 	DeletedTimestamp string       `json:"deleted_ts,omitempty"`
 	Attachments      []Attachment `json:"attachments,omitempty"`
-	ReplyTo          int          `json:"reply_to,omitempty"`
+	ReplyTo          string       `json:"reply_to,omitempty"`
 	Upload           bool         `json:"upload,omitempty"`
 }
 


bpowers:
ugh, it sometimes _does_ return an int.  what hell

nlopes:
Yeah, I know. Frustrating. I haven't figured out a way to have this properly researched.

nlopes:
I am really not sure what to do here. Unless more people complain, I will leave this open.

dvdplm:
Here's what we did to tackle this:

```
type IntOrString struct {
    Kind   IntstrKind
    IntVal int
    StrVal string
}
// IntstrKind represents the stored type of IntOrString.
type IntstrKind int

const (
    IntstrInt    IntstrKind = iota // The IntOrString holds an int.
    IntstrString                   // The IntOrString holds a string.
)

type SlackRtmMessage struct {
    ReplyTo util.IntOrString `json:"reply_to"`
        …
        …
}
```

The rest of the implementation we borrowed from https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/util/util.go#L122

It has worked fine for us, might be a "fix" here too?

james-lawrence:
due to age.

PR #14, opened 2015-04-30
BaseChannel: channels, groups and IMs all share common properties

bpowers:
In seeing I needed to add some fields to IM, it was clear that these 3
types of objects all share common fields.  `BaseChannel` is the part
of the objects common to all.  Open to better a better name - I went
with `BaseChannel` because the IDs of IMs and Groups are both used in
the 'channel' field of messages.

### DIFF
--- a/channels.go
+++ b/channels.go
@@ -30,12 +30,21 @@ type ChannelPurpose struct {
 	LastSet JSONTime `json:"last_set"`
 }
 
+type BaseChannel struct {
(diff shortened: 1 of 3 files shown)


nlopes:
Thanks. This should be fine but I haven't had time to merge this locally and test. Once I do, I will let you know.

nlopes:
Did not have time to look into this yet. Will do at my first chance.

PR #16, opened 2015-05-08
Fix JSON Roundtrip Issues

joefitzgerald:
This fixes a few issues preventing roundtrips (writing JSON back out to disk) from matching the original JSON coming back from `api.slack.com`.
- `omitempty` won't omit fields like `Channel`>`Latest` unless it's a pointer
- `is_starred` isn't present for `channels`
- `is_member` isn't present for `groups`
- `is_general` isn't present for `groups` (only `channels` can be general)

### DIFF
--- a/channels.go
+++ b/channels.go
@@ -40,7 +40,6 @@ type Channel struct {
 	IsArchived         bool           `json:"is_archived"`
 	IsGeneral          bool           `json:"is_general"`
 	IsGroup            bool           `json:"is_group"`
-	IsStarred          bool           `json:"is_starred"`
 	Members            []string       `json:"members"`
 	Topic              ChannelTopic   `json:"topic"`
 	Purpose            ChannelPurpose `json:"purpose"`
--- a/channels.go
+++ b/channels.go
@@ -46,7 +46,7 @@ type Channel struct {
 	Purpose            ChannelPurpose `json:"purpose"`
 	IsMember           bool           `json:"is_member"`
 	LastRead           string         `json:"last_read,omitempty"`
-	Latest             Message        `json:"latest,omitempty"`
+	Latest             *Message       `json:"latest,omitempty"`
 	UnreadCount        int            `json:"unread_count,omitempty"`
 	NumMembers         int            `json:"num_members,omitempty"`
 	UnreadCountDisplay int            `json:"unread_count_display,omitempty"`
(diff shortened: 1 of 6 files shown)


nlopes:
Thanks for the patch. Let me take a deeper look into this. Will probably happen this weekend.

nlopes:
Did not have time to look into this yet. Will do at my first chance.

joefitzgerald:
Bump @nlopes.

channels.go
```diff
@@ -39,14 +39,12 @@ type Channel struct {
 	Creator            string         `json:"creator"`
 	IsArchived         bool           `json:"is_archived"`
 	IsGeneral          bool           `json:"is_general"`
-	IsGroup            bool           `json:"is_group"`
-	IsStarred          bool           `json:"is_starred"`
```
nlopes, line 43: I am pretty sure this is part of a Channel structure. Are you sure we can remove this?

joefitzgerald, line 43: Yes; go and export all your channels and observe no IsStarred in the JSON, even with starred channels.

nlopes, line 43: I just connected to the websocket api and the first info payload return contains channels with the is_starred property.

joefitzgerald, line 43: That's definitely an inconsistency in their API then. I'd almost think of the websockets info payload as a separate use case (with potentially a separate struct).

nlopes, line 43: I know. It isn't the only one by the way. They also have inconsistencies with the way they send back timestamps. Really annoying.

nlopes:
@joefitzgerald I decided to merge it anyway. Hopefully they move towards a better API and I'll work a bit harder on the inconsistencies.

dvdplm:
This change has brought with it other inconsistencies such as those pointed out in #30 
There are other similar inconsistencies in the Slack response, e.g. the `User` data returned by RTM Start contains the `presence` key, while the `User` returned by events does not.
I wonder what the real value of preserving roundtrip consistency is, @joefitzgerald. Perhaps it's better to ensure that the hosting data structures on the library side are consistent? I.e. always use pointers and always contain all the fields that may or may not be present in the data?

joefitzgerald:
@dvdplm I posited in #30 that we could use pointers for optional fields, and not use pointers for required fields. I think you're right that the Slack API itself is inconsistent. I've written a bunch of tests to try to get group and channel history to function correctly with all the supported events. And this certainly highlights some of the issues.

dvdplm:
@joefitzgerald yeah, saw that (thanks for the feedback). I've pinged slackhq on twitter about the `presence` key issue the last few days but so far they've replied evasively.

PR #21, opened 2015-05-23
Support groups.info API

spiderpug:
Adds support for the "groups.info" method.

API description:  https://api.slack.com/methods/groups.info

Feedback is welcome :smile:


nlopes:
Thanks!

PR #22, opened 2015-05-26
Add admin InviteGuest support

krishicks:
This uses the undocumented users.admin.invite method. This doesn't use
the regular api.slack.com endpoint, but instead uses teamname.slack.com
with the method and an epoch tacked on the end:

https://teamname.slack.com/api/users.admin.invite?t=1416723927

The idea for this came from @levelsio: https://levels.io/slack-typeform-auto-invite-sign-ups/

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -1,0 +1,52 @@
+package slack
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+)
+
+type adminResponse struct {
+	OK    bool   `json:"ok"`
+	Error string `json:"error"`
+}
+
+func adminRequest(method string, teamName string, values url.Values, debug bool) (*adminResponse, error) {
+	adminResponse := &adminResponse{}
+	err := parseAdminResponse(method, teamName, values, adminResponse, debug)
+	if err != nil {
+		return nil, err
+	}
+
+	if !adminResponse.OK {
+		return nil, errors.New(adminResponse.Error)
+	}
+
+	return adminResponse, nil
+}
+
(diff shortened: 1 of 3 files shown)


nlopes:
If I am adding this, can you make it so that we can pass a list of channels? From what I can gather from levels.io it actually supports multiple channels separated by commas. Thanks!

krishicks:
Sounds good. I updated it to take a slice of strings for channelIDs.

joefitzgerald:
A single channel guest only can only be a member of a single channel. That's the whole point of it. This shouldn't take a slice of strings, but rather a single string.

joefitzgerald:
A restricted account can take a slice of strings, but that's another proposition altogether.

krishicks:
Good point. Back to where we were.

nlopes:
InviteGuest doesn't really transpire the fact that it's a single channel to me. Only the error in that function and the fact that you're only passing one channel do. Inviting a guest to a channel seems a subset of inviting a guest to multiple channels. The point of the api call seems to be to invite the guest to N amount of channels (assuming what is described in levelsio is correct). Am I wrong in assuming all this?

joefitzgerald:
@nlopes there will be an alternative method - invite restricted - which is coming in a subsequent PR. This invite method is for single channel guests.

nlopes:
Alright, makes sense then.

krishicks:
To close the loop: https://github.com/nlopes/slack/pull/24

PR #25, opened 2015-06-03
WIP: Add Missing Fields For Message Subtypes

joefitzgerald:
:rotating_light: WIP - Still Adding Tests :rotating_light: 

This:
- Adds missing fields for message subtypes
- Adds tests for message mapping, based on actual API examples as of June 3rd, 2015 and the examples at https://api.slack.com/events/message
- Removes redundant and mis-cased `Id` usage across the entire library (fixing a vast number of lint issues in both this library and consumers of it: https://github.com/golang/lint/blob/master/lint.go#L712)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@
 			return
 		}
 		for _, group := range groups {
-			fmt.Printf("Id: %s, Name: %s\n", group.Id, group.Name)
+			fmt.Printf("ID: %s, Name: %s\n", group.ID, group.Name)
 		}
 	}
 
@@ -48,7 +48,7 @@
 		    fmt.Printf("%s\n", err)
 		    return
 	    }
-	    fmt.Printf("Id: %s, Fullname: %s, Email: %s\n", user.Id, user.Profile.RealName, user.Profile.Email)
+	    fmt.Printf("ID: %s, Fullname: %s, Email: %s\n", user.ID, user.Profile.RealName, user.Profile.Email)
     }
 
 ## Why?
--- a/admin.go
+++ b/admin.go
@@ -96,6 +96,23 @@ func (api *Slack) InviteRestricted(
 	return nil
 }
 
+// SetRegular enables the specified user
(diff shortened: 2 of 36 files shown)


nlopes:
This is a great and much needed change. I love it.

I am definitely good with User/Channel instead of UserId/ChannelId (I've gone back and forth locally on that myself) _but_ I don't like the ID change from Id. I know golint complains about it (I've mentioned that before in https://github.com/nlopes/slack/pull/4) but IMO it shouldn't. Nothing in the spec says it should complain and I prefer Id to ID. Golint seems to prefer ID (https://github.com/golang/lint/issues/124) even though I feel Id is a contraction, not an acronym.

If you feel strongly about this, I will accept that change (especially  because this is the second time being brought up) but I won't like it :D

joefitzgerald:
> If you feel strongly about this

It really has nothing to do with how I feel (although if feelings are in question, it's probably those of the go team that need to be inspected). It is about the go in this project resembling most other go in the wild.

Tooling like gofmt / golint exist to end pitchfork battles like tabs vs spaces (or Id vs ID :fork_and_knife:?). It's just not worth getting upset over. You will continue to have people attempt to "fix" the code in perpetuity because most people are running these tools (quite rightly) when editing any go code. So even if it doesn't pass your correctness filter, perhaps just accept that idiomatic go code uses `ID` instead of `Id` (or doesn't use Id/ID at all) and that alone may be reason to use it too.

joefitzgerald:
Thanks for the encouraging words on the rest of the changes! I'm shooting for 100% fidelity when round-tripping any message type or subtype so that exporting private group or channel history to a Slack compatible import archive is possible.

nlopes:
"perhaps just accept that idiomatic go code uses ID instead of Id"
Not quite. I just grep'd through the code base and the go codebase actually uses Id and ID. _sigh_
That being said, it won't help this codebase to discuss this stuff. Feel free to assume ID is fine.

PR #37, opened 2015-08-07
Add SetRestricted/SetUltraRestricted, update DisableUser godoc

krishicks:


admin.go
```diff
@@ -129,3 +129,45 @@ func (api *Slack) SendSSOBindingEmail(teamName string, user string) error {
 
 	return nil
 }
+
+// SetUltraRestricted converts a user into a single-channel guest
+func (api *Slack) SetUltraRestricted(
+	teamName string,
+	uid string,
+	channel string,
+) error {
```
nlopes, line 138: This would probably look better as func(api *Slack) SetUltraRestricted(teamName, uid, channel string) error { instead of the multiline.

admin.go
```diff
@@ -129,3 +129,45 @@ func (api *Slack) SendSSOBindingEmail(teamName string, user string) error {
...
+	}
+
+	return nil
+}
+
+// SetRestricted converts a user into a restricted account
+func (api *Slack) SetRestricted(
+	teamName string,
+	uid string,
+) error {
```
nlopes, line 159: Same here.

nlopes:
Just noticed we have the same stuff in other parts of the code so I'll merge and fix later on.

PR #42, opened 2015-08-20
GroupConversation: open and centralize some boolean flags.

abourget:
In https://github.com/abourget/slick .. I keep references to those channels and conversations, and I sometimes tweak them... I need to be able to create some and keep them updated..

any reasons to keep those internal ?

This is just a proposition.. nothing fixed in stone here..

### DIFF
--- a/channels.go
+++ b/channels.go
@@ -19,9 +19,6 @@ type channelResponseFull struct {
 // Channel contains information about the channel
 type Channel struct {
 	GroupConversation
-	IsChannel bool `json:"is_channel"`
-	IsGeneral bool `json:"is_general"`
-	IsMember  bool `json:"is_member"`
 }
 
 func channelRequest(path string, values url.Values, debug bool) (*channelResponseFull, error) {
--- a/channels.go
+++ b/channels.go
@@ -18,7 +18,7 @@ type channelResponseFull struct {
 
 // Channel contains information about the channel
 type Channel struct {
-	groupConversation
+	GroupConversation
 	IsChannel bool `json:"is_channel"`
 	IsGeneral bool `json:"is_general"`
 	IsMember  bool `json:"is_member"`
--- a/conversation.go
+++ b/conversation.go
@@ -16,6 +16,10 @@ type GroupConversation struct {
 	Conversation
 	Name       string   `json:"name"`
 	Creator    string   `json:"creator"`
+	IsChannel  bool     `json:"is_channel"`
+	IsGroup    bool     `json:"is_group"`
+	IsGeneral  bool     `json:"is_general"`
+	IsMember   bool     `json:"is_member"`
 	IsArchived bool     `json:"is_archived"`
 	Members    []string `json:"members"`
 	NumMembers int      `json:"num_members,omitempty"`
--- a/conversation.go
+++ b/conversation.go
@@ -1,7 +1,7 @@
 package slack
 
 // Conversation is the foundation for IM and BaseGroupConversation
-type conversation struct {
+type Conversation struct {
 	ID                 string   `json:"id"`
 	Created            JSONTime `json:"created"`
 	IsOpen             bool     `json:"is_open"`
@@ -12,8 +12,8 @@ type conversation struct {
 }
 
 // GroupConversation is the foundation for Group and Channel
-type groupConversation struct {
-	conversation
+type GroupConversation struct {
+	Conversation
 	Name       string   `json:"name"`
 	Creator    string   `json:"creator"`
 	IsArchived bool     `json:"is_archived"`
--- a/groups.go
+++ b/groups.go
@@ -8,7 +8,7 @@ import (
 
 // Group contains all the information for a group
 type Group struct {
-	groupConversation
+	GroupConversation
 	IsGroup bool `json:"is_group"`
 }
 
--- a/groups.go
+++ b/groups.go
@@ -9,7 +9,6 @@ import (
 // Group contains all the information for a group
 type Group struct {
 	GroupConversation
-	IsGroup bool `json:"is_group"`
 }
 
 type groupResponseFull struct {
--- a/im.go
+++ b/im.go
@@ -22,7 +22,7 @@ type imResponseFull struct {
 
 // IM contains information related to the Direct Message channel
 type IM struct {
-	conversation
+	Conversation
 	IsIM          bool   `json:"is_im"`
 	User          string `json:"user"`
 	IsUserDeleted bool   `json:"is_user_deleted"`
--- a/websocket_managed_conn.go
+++ b/websocket_managed_conn.go
@@ -112,7 +112,7 @@ func (rtm *RTM) startRTMAndDial() (*Info, *websocket.Conn, error) {
 		return nil, nil, err
 	}
 
-	conn, err := websocket.Dial(url, "", "http://api.slack.com")
+	conn, err := websocketProxyDial(url, "http://api.slack.com")
 	if err != nil {
 		return nil, nil, err
 	}
--- a/websocket_proxy.go
+++ b/websocket_proxy.go
@@ -1,0 +1,83 @@
+package slack
+
+import (
+	"crypto/tls"
+	"errors"
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"os"
+	"strings"
+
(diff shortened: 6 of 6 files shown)


abourget:
so for simplicty of the bot, I like to have a "Channel" being able to represent a Group.

Why have `IsChannel` on a Channel.. if there's no way it is not a channel.. why have `IsGroup` on a Group, if there's no way it's not a group.

I feel we should have a `Channel` with "IsGroup" and distinguish it like that.. shouldn't we ? would be way simpler..

abourget:
@mlbvsnba was that unified Channel model something that you found useful in my fork ?

abourget:
Ok, this last commit re-similarizes Group and Channel.. through `GroupConversation`.

abourget:
Cool.. I just re-migrated to your breaking changes branch.. couple of quick fixes and I'm back on track (with this PR in).

Thanks for merging all of this !

abourget:
Whoops, added yet another feature. Support for Websockets through HTTP_PROXYs !

nlopes:
I agree with the channel/group changes as I thought of changing it myself to make things easier. The websocket proxy change maintains backwards compatibility so I am good with it as well. Thanks!

im.go
```diff
@@ -22,7 +22,7 @@ type imResponseFull struct {
 
 // IM contains information related to the Direct Message channel
 type IM struct {
-	conversation
+	Conversation
```
joefitzgerald, line 25: This shouldn't be exported. The fields within conversation are exported and can be accessed that way.

groups.go
```diff
@@ -8,8 +8,7 @@ import (
 
 // Group contains all the information for a group
 type Group struct {
-	groupConversation
-	IsGroup bool `json:"is_group"`
+	GroupConversation
```
joefitzgerald, line 11: I think you may have missed the point of adding GroupConversation. It includes the common members that are shared by both Channel and Group. You cannot be a member (`IsMember`) of a Channel. You cannot have a group that `IsGeneral`. `IsChannel` and `IsGroup` are mutually exclusive. Channels and Groups never come back in a unified list. Consequently, I'm not sure why you made the change?

Further, groupConversation shouldn't be exported.

conversation.go
```diff
@@ -12,10 +12,14 @@ type conversation struct {
 }
 
 // GroupConversation is the foundation for Group and Channel
-type groupConversation struct {
-	conversation
+type GroupConversation struct {
+	Conversation
 	Name       string   `json:"name"`
 	Creator    string   `json:"creator"`
+	IsChannel  bool     `json:"is_channel"`
```
joefitzgerald, line 19: Not valid in a Group.

conversation.go
```diff
@@ -12,10 +12,14 @@ type conversation struct {
...
+	IsGroup    bool     `json:"is_group"`
```
joefitzgerald, line 20: Not valid in a Channel.

conversation.go
```diff
@@ -12,10 +12,14 @@ type conversation struct {
...
+	IsGeneral  bool     `json:"is_general"`
```
joefitzgerald, line 21: Not valid in a Group.

conversation.go
```diff
@@ -12,10 +12,14 @@ type conversation struct {
...
+	IsMember   bool     `json:"is_member"`
```
joefitzgerald, line 22: Not valid in a Channel.

channels.go
```diff
@@ -18,10 +18,7 @@ type channelResponseFull struct {
 
 // Channel contains information about the channel
 type Channel struct {
-	groupConversation
-	IsChannel bool `json:"is_channel"`
-	IsGeneral bool `json:"is_general"`
-	IsMember  bool `json:"is_member"`
+	GroupConversation
```
joefitzgerald, line 21: I think you may have missed the point of adding GroupConversation. It includes the common members that are shared by both Channel and Group. You cannot be a member (`IsMember`) of a Channel. You cannot have a group that `IsGeneral`. `IsChannel` and `IsGroup` are mutually exclusive. Channels and Groups never come back in a unified list. Consequently, I'm not sure why you made the change?

Further, groupConversation shouldn't be exported.

conversation.go
```diff
@@ -1,7 +1,7 @@
 package slack
 
 // Conversation is the foundation for IM and BaseGroupConversation
-type conversation struct {
```
joefitzgerald, line 4: This shouldn't be exported. The fields within conversation are exported and can be accessed when embedded.

conversation.go
```diff
@@ -12,10 +12,14 @@ type conversation struct {
 }
 
 // GroupConversation is the foundation for Group and Channel
-type groupConversation struct {
-	conversation
+type GroupConversation struct {
```
joefitzgerald, line 15: This shouldn't be exported. The fields within groupConversation are exported and can be accessed when embedded.

conversation.go
```diff
@@ -12,10 +12,14 @@ type conversation struct {
...
+	Conversation
```
joefitzgerald, line 16: This shouldn't be exported. The fields within conversation are exported and can be accessed that way.

joefitzgerald:
@abourget I'm trying to figure out why you wanted these changes? Is it simply because you want to represent a single list of all channels and groups and be able to perform actions based on that list and the type of the group conversation? These changes create JSON that is not valid when interacting with `channel` and `group` resources.

abourget:
Ah I see. Yes the goal is to work with a unified list. It simplifies a lot
of things in a statically typed world. When reading having "IsGroup" on a
Group doesn't provide much value.. where it could never be false. However
having IsGroup being false on a channel is helpful.

I understand that when writing, things are different. Could we make these
precise fields be pointers to bool, and add "omitempty" for when we write ?
Does the API fail when we send an `"is_group": false` in a channel context ?

Also regarding the Conversation being exported, if it isn't,  you can't
create those objects.. only modify its fields.. so you need to have huge
copy functions.  So I ask you, why not export them ?

Le lun. 31 août 2015 17:39, Joe Fitzgerald notifications@github.com a
écrit :

nlopes:
`Does the API fail when we send an`"is_group": false`in a channel context ?` - I am going to test this because this is something I was thinking as well.
On the other hand, I really don't like giving the ability to people to set IsGroup to true on a Channel by mistake. @joefitzgerald brings some valid points to the conversation and they're probably a bit saner (from a developer perspective) so I am going to re-analyse this merge and might revert it.

abourget:
In the same vein, I need ReactionEvent to be opened up.. for the same
reason. I want to be able to deal with both reactions with a single API.

Le jeu. 10 sept. 2015 14:48, Norberto Lopes notifications@github.com a
écrit :

abourget:
In which case do people create Channel objects ? Those objects are read into.. and in that case theirs fields are appropriate with my changes. We always refer to channels with their ID in API calls.

When I create Channel objects.. it is to avoid duplication of my code, by considering a Group as a Channel, which it _is_ conceptually  (and most probably in Slack's own implementation!).

I'm sorry to say, I don't see the point being made by "groupConversation shouldn't be exported". I'm presenting a use case, "it shouldn't" is not a compelling reason enough to dismiss it.

Again, having IsChannel and IsGroup is absolutely useless in Go if we only have it on their respective objects, because the _type_ of the object already tells us that. IsChannel and IsGroup is only useful if it can discriminate _against_ something else, which it can help do in a GroupConversation.

I'd prefer not to maintain my own fork because of issues like this. Can we settle on this matter ? Especially in the light of a similar change regarding ReactionEvent I'm about to propose.

abourget:
Ok folks, I'm willing to revert the groupConversation changes if you want.. I needed to have the same support for `IM`and `IsIM` but making it clean in the `slack` codebase was becoming very ugly.. so I reverted to create my own abstraction in my bot.. This means I wouldn't need the GroupConversation to be open (as I'm not going to use the `slack.Channel` for the abstraction), it can be closed up, and we can put back the `IsChannel` only on the `Channel` object.  Just beware that this PR included the HTTP_PROXY support too.. so maybe we want to revert only the two concerned commits.

nlopes:
Awesome, thank you for this. I'll be reverting that either today or tomorrow.

nlopes:
Argh, I ended up spending more time away from this than I wished.
The two commits have been reverted, the change for HTTP_PROXY has been left in.

Thank you so much @abourget

PR #44, opened 2015-09-12
Add pinning support

nlopes:
This is temporary support for pinning. Once I have this fleshed out (shouldn't be too far out - maybe this PR will even do the trick), I will merge.

### DIFF
--- a/examples/pins/pins.go
+++ b/examples/pins/pins.go
@@ -1,0 +1,123 @@
+package main
+
+import (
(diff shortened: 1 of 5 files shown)


nlopes:
@joefitzgerald I would love to have an extra pair of eyes here if you don't mind.

nlopes:
@abourget If you don't mind also taking a look at this, I would really appreciate it.

PR #45, opened 2015-09-24
Implemented a Time() function on the JSONTimeString and JSONTime objects

abourget:
Took out a potential panic too!


nlopes:
Will look into this today or tomorrow. Thank you so much!

PR #46, opened 2015-10-29
adding function that invites a user to a team

selfcontained:
Wanted the ability to be able to send an invitation to a user to join the slack team, but not restrict them to a specific channel.
- Added  `InviteToTeam(teamName, string, firstName string, lastName string, email string)`


nlopes:
Thanks!

PR #52, opened 2016-01-11
Add missing fields to the File

dvdplm:
Starting on Jan 4 the Slack API wrt files changed substantially. The blog post outlining the changes is here: https://medium.com/slack-developer-blog/important-changes-to-files-in-the-web-api-eb38f2a9c1e7#.dwnc9pp7o
The information in the blog post is incomplete and in reality the data that comes in over the wire is actually different still. For example there are numerous new thumbnail sizes (is a 1024px file still a "thumbnail"? :).
The PR leaves the now removed `URL` and `URLDownload` members in the `File` struct. I'm happy to remove them if you prefer. It depends on what you think the best way of dealing with such breaking changes is.

The closest equivalent to `URL` is the new `PermalinkPublic` field. It's worth knowing that when such a link is accessed for the first time, the slackbot will mention it in a private message to the user, giving them the option to revoke the permissions on the file (accessing it again yields a 404).

### DIFF
--- a/files.go
+++ b/files.go
@@ -23,12 +23,13 @@ type File struct {
 	Created   JSONTime `json:"created"`
 	Timestamp JSONTime `json:"timestamp"`
 
-	Name       string `json:"name"`
-	Title      string `json:"title"`
-	Mimetype   string `json:"mimetype"`
-	Filetype   string `json:"filetype"`
-	PrettyType string `json:"pretty_type"`
-	User       string `json:"user"`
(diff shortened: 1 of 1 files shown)


dvdplm:
@nlopes gentle poke! :)

nlopes:
Sorry @dvdplm. Was on vacation :D

I agree with what was said, so I'm merging this for now.

dvdplm:
Great, thanks! :D

PR #53, opened 2016-01-13
Support full OAauth response (for adding hooks and bot users via OAuth)

johnboiles:
Because `GetOAuthToken`'s signature is unchanged, this should be fully compatible with existing code that uses `oauth.go`

### DIFF
--- a/oauth.go
+++ b/oauth.go
@@ -5,27 +5,50 @@ import (
 	"net/url"
 )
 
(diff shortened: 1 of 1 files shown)


dvdplm:
+1

nlopes:
Thanks!

johnboiles:
Woohoo!

(Sent from my iPhone)

PR #54, opened 2016-01-30
In other events, the User is not a pointer

dvdplm:
For consistency and Least Surprise it's nice if the TeamJoinEvent also returns a `User`.
See https://github.com/nlopes/slack/issues/30


nlopes:
Mind removing the `omitempty` as well? Doesn't make much sense to be there.

dvdplm:
good pt, fixing

nlopes:
Thanks!

PR #55, opened 2016-01-30
Remove JSONTimeString

dvdplm:
The Slack API docs explicitly states that their funny timestamps are to be treated as a sort of message IDs and are guaranteed to be unique in the channel. In update and delete calls it is necessary to know the timestamp-string or the call will fail. Client code will use these timestamp-strings for things like checking for duplicate messages.

Instead of casting incoming timestamp-strings to (english speaking) human-readable strings like "Sun Jan 31" I think it's wise to let the client do its own formatting and leave the timestamp-strings be.

[Slack docs](https://api.slack.com/rtm):

```
All messages within a single channel are guaranteed to have a unique timestamp which is 
ASCII sortable. Given the precision of the timestamp, clients should treat these timestamps 
as strings, not floats/doubles.
```

### DIFF
--- a/websocket_channels.go
+++ b/websocket_channels.go
@@ -4,7 +4,7 @@ package slack
 type ChannelCreatedEvent struct {
 	Type           string             `json:"type"`
 	Channel        ChannelCreatedInfo `json:"channel"`
-	EventTimestamp JSONTimeString     `json:"event_ts"`
+	EventTimestamp string             `json:"event_ts"`
 }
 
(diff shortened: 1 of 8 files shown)


nlopes:
Let me take a closer look at this soon™

dvdplm:
+1

dvdplm:
Gentle poke: any news on this? :)

nlopes:
Haven't forgotten about this, just been pretty busy. Apologies!

nlopes:
Thank you!

PR #58, opened 2016-02-09
Handle "reconnect_url" event

fujimoto:
as of Feb 2016 (or so), slack started to send "reconnect_url" event via RTM API, and this causes unmasharling_error event (which is not desirable). and this will simply handle the event (Slack API documentation still [says nothing](https://api.slack.com/events/reconnect_url) about this event, so just handling the event, not anymore.


fujimoto:
thanks! @nlopes

PR #59, opened 2016-02-16
slack/team: GetTeamInfo for Slack Client is added

mehmetalisavas:
hey @nlopes 
I've searched for TeamInfo method for slack but i couldn't find.
Then I've used this and this may be helpful for others also..

By the way, good work.. !!

### DIFF
--- a/team.go
+++ b/team.go
@@ -1,0 +1,46 @@
+package slack
+
+import (
+	"errors"
+	"net/url"
+)
+
+type TeamResponse struct {
+	Team TeamInfo `json:"team"`
+	SlackResponse
+}
+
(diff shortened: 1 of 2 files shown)


nlopes:
Do you mind adding tests? I know they are currently lacking in a lot of these calls but I am thinking it is about time we all start adding them (that goes for me as well). 
Thanks for this!

mehmetalisavas:
I've added a simple test for it with incoming data. 
Also parsing is working as well ..

nlopes:
Will merge once those commits are squashed ;)

mehmetalisavas:
@nlopes thx for ur attention :)

PR #64, opened 2016-03-17
Posting message as user shouldn't unfurls links

AlphaB:
[Unfurling documentation](https://api.slack.com/docs/unfurling) says that message posted with `chat.postMessage` won't unfurl links:

> For messages posted via webhooks or the chat.postMessage API method, we will unfurl links to media, but not other links.

but posting message `as_user` will unfurl links by default:

> By default we unfurl all links in any messages posted by users. 

I've wrote a message to Slack support, and they said this: 

> I apologize for the confusion in the documentation here. The `By default we unfurl all links in any messages posted by users. referenced in (https://api.slack.com/docs/unfurling)` appears to be overriding the `chat.postMessage` exception when the `as_user` parameter is set to true.

So we should explicitly send `unfurl_links: false` when `as_user: true`. This patch fixes this behaviour.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -125,6 +125,11 @@ func (api *Client) PostMessage(channel, text string, params PostMessageParameter
 	if params.UnfurlLinks != DEFAULT_MESSAGE_UNFURL_LINKS {
 		values.Set("unfurl_links", "true")
 	}
+	// I want to send a message with explicit `as_user` `true` and `unfurl_links` `false` in request.
+	// Because setting `as_user` to `true` will change the default value for `unfurl_links` to `true` on Slack API side.
+	if params.AsUser != DEFAULT_MESSAGE_ASUSER && params.UnfurlLinks == DEFAULT_MESSAGE_UNFURL_LINKS {
+		values.Set("unfurl_links", "false")
+	}
 	if params.UnfurlMedia != DEFAULT_MESSAGE_UNFURL_MEDIA {
 		values.Set("unfurl_media", "false")
 	}


nlopes:
I am not convinced this is the right thing to do (at least until they change the defaults as shown [here](https://api.slack.com/methods/chat.postMessage)). The defaults for as_user is ~True~ False and unfurl_link is ~True~ False as well. You can always override these when calling the function so I am leaving it as it is. If they end up changing their documentation defaults (and API defaults) then I'll change these as well.

Does that make sense?

AlphaB:
The problem is that [this](https://github.com/nlopes/slack/blob/master/chat.go#L125) line of code doesn't allow explicitly sending `unfurl_links: false` because [default value](https://github.com/nlopes/slack/blob/master/chat.go#L15) for this flag is `false`. So I can't send a message with clear `as_user: true` and `unfurl_links: false`.

nlopes:
`PostMessage("channel", "message", PostMessageParameters{ AsUser: true })` should do exactly what you need. Am I wrong?

You are right in that the defaults are false, not true. I changed my comment above to reflect that.

AlphaB:
I'm sorry, seems like my main goal wasn't formulated properly. I want to send a message with explicit `as_user: true` and `unfurl_links: false` in request, because setting `as_user` to `true` will change the default value for `unfurl_links` to `true` on Slack API side.
This code 

> `PostMessage("channel", "message", PostMessageParameters{ AsUser: true, UnfurlLinks: false })`

won't send `unfurl_links: false` because of the line of code that I've referenced in my previous comment. And Slack will unfurl all non-media links in this message, even if I don't want to.

nlopes:
Do you mind adding 
"I want to send a message with explicit as_user: true and unfurl_links: false in request, because setting as_user to true will change the default value for unfurl_links to true on Slack API side."
as a comment before those entries in your commit? I think it will help other people understand the issue.

Thank you for catching this! Will merge after comment.

AlphaB:
All done!

nlopes:
Awesome! Thank you again!

PR #68, opened 2016-04-14
Fix for issue 27, "Allow externally-provided *http.Client"

joelgwebber:
This is a very straightforward implementation, and I'm open to doing it differently if you prefer.

In particular, it might be a bit simpler just to pass the entire `api *Client` reference straight to all the `*Request()` functions, rather than passing `httpClient` and `debug` separately.

Also, it might be easier (though a bit more subtle) to allow `nil` as the `*http.Client` parameter to `New()`, so that people don't have to know to grab `http.DefaultClient`. It would be nicer with optional function parameters, but I'd also be fine with separating that into a second `NewWithClient()` or similar function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: go
 
 go:
     - 1.4
+    - 1.5
+    - 1.6
     - tip
 
 before_install:
--- a/admin.go
+++ b/admin.go
@@ -3,6 +3,7 @@ package slack
 import (
 	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 )
 
@@ -11,9 +12,9 @@ type adminResponse struct {
 	Error string `json:"error"`
 }
 
-func adminRequest(method string, teamName string, values url.Values, debug bool) (*adminResponse, error) {
+func adminRequest(client *http.Client, method string, teamName string, values url.Values, debug bool) (*adminResponse, error) {
 	adminResponse := &adminResponse{}
-	err := parseAdminResponse(method, teamName, values, adminResponse, debug)
+	err := parseAdminResponse(client, method, teamName, values, adminResponse, debug)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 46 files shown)


nlopes:
Sorry, just got back to this.
Thanks for all the work you put in this.

Everything looks good. I have a preference for something like:

```
    api := slack.New("YOUR_TOKEN_HERE") # this would set the httpClient to the default of course
    api.SetHttpClient(yourHttpClient)
    channels, err := api.GetChannels(false)
```

versus

```
    api := slack.New("YOUR_TOKEN_HERE", yourHttpClient)
    channels, err := api.GetChannels(false)
```

The reason is to keep backwards compatibility. passing the client to `.New` makes sense but would break all the clients out there.

I am aware that the change to `GetOAuthToken` would still break clients but it wouldn't be as pernicious as the change to `New`

What do you think?

joelgwebber:
Oh, right, of course. Don't know why I didn't think about the compatibility issue. I'd be fine with either your suggestion of adding a mutator for the `http.Client`, or adding a second `NewWithClient()` or similar function. I have a very slight allergy to the mutator, in that generally you'll want to set it only once -- I don't see anything in the current code that would _break_ if you re-set the client multiple times, but it could become a bit of a constraint on future changes.

Let me know which you prefer, and I'll update the PR.

flexd:
A good solution for this would be to implement functional options. That way the default could work as you want it, and you could add whatever options you want as functions that modify the config.

See http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis :-)

joelgwebber:
@flexd Good point, and that would be source-compatible with the existing API. @nlopes That work for you?

nlopes:
Indeed. Good shout.

dlsniper:
There's now an ` HTTPClient ` package level variable, which can be changed so that it can use a custom http.Client fixed in #73. I'll close this for now.

PR #71, opened 2016-04-25
Set presence=1 in GetUsers to get presence data

flexd:
This should probably be something you can specify when calling `GetUsers()`, but I'm in the web editor at the moment. Tested locally and works for me :-)


nlopes:
This seems to be the right thing to do. Just gives a bit more data and doesn't break compatibility, which introducing a parameter would.

Thanks!

flexd:
Awesome, thanks! :-)!

PR #73, opened 2016-05-17
Make the HTTP client accessible from the outside world

lestrrat:
This allows this library to be used within Google App Engine by saying

```
  ctx := appengine.NewContext(r)
  slack.HTTPClient.Transport = &urlfetch.Transport{Context: ctx}
```

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -16,6 +16,8 @@ import (
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


nlopes:
I missed this for some reason. Apologies.

PR #75, opened 2016-05-27
Let slack use an external logger

parsley42:
I wanted to supply a logger to your lib so I could fold in api debug logs with my chatbots logs. If you'd rather it be a struct member of api, I could do that; there are only a couple of places in the code that don't use api.DebugX.

Thanks for the great lib, made it pretty easy to get gopherbot going.


nlopes:
Thanks!

PR #84, opened 2016-06-29
Fix debugging and logging

dlsniper:
Hi,

This fixes calling SetDebug without having a logger set. Right now doing so breaks the apps / examples.
Thank you.

### DIFF
--- a/slack.go
+++ b/slack.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"log"
 	"net/url"
+	"os"
 )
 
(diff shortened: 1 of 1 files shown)


parsley42:
I feel bad that my earlier pull request introduced this bug, sorry. I've seen a few patches go by, but this looks like the best approach.

dlsniper:
@parsley42 heh, all cool, it happens ;)

nlopes:
I didn't catch it either. Much appreciated @dlsniper

PR #85, opened 2016-06-30
Avoid using globals to expose http.Client

lestrrat:
This is an expansion of my previous PR, https://github.com/nlopes/slack/pull/73.
Previously I was mucking with globals, but in many cases we need to create http.Client instances that refer to different contexts.

While the gist of the change is fairly simple, but since this is a rather big PR, I understand if you have concerns about merging.

---

Now if you want to use this in appengine, you should do:

``` go
  c := slack.New(...)
  c.HTTPClient = &http.Client{
    Transport: &urlfetch.Transport{
      Context: appengine.NewContext(r),
    },
  }
```

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -3,6 +3,7 @@ package slack
 import (
 	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 )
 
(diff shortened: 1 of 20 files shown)


james-lawrence:
a fix for this has been merged, closing.

PR #93, opened 2016-08-08
Add UnmarshalJSON to Message and params to UpdateMessage

erizocosmico:
This PR adds `UnmarshalJSON` to Message so you can unmarshal an attachment action callback directly. Otherwise it fails due it having the original message as a quoted string and a string can be unmarshalled in `Message` struct.

It also adds parameters to `UpdateMessage` so more complex messages can be used.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -60,6 +60,24 @@ func NewPostMessageParameters() PostMessageParameters {
 	}
 }
 
+// UpdateMessageParameters contains all the parameters necessary (including the optional ones) for an UpdateMessage() request
+type UpdateMessageParameters struct {
+	AsUser      bool
+	Parse       string
+	LinkNames   int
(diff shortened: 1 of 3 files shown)


nlopes:
This breaks compatibility. I am in the process of doing some other changes that may be breaking changes, so will merge this then. Shouldn't be long. I am now back on this (had a bit of a hiatus there).

Thanks for submitting and apologies for the late response!

edwardIshaq:
Hello @nlopes,
I'm a big fan of your repo, and been trying to find ways to contribute. 
I was wondering when will this be reopened or added as a feature ? 
it seems like it could be worked around by allowing options on a function similar to 
```
func (api *Client) UpdateMessage(channelID, timestamp, text string) (string, string, string, error) {
```
What do you think ?

PR #96, opened 2016-08-30
Tells JSON interpreter that TS is a string encoded int64

mehlium:


tobert:
The field was never an int64. It's a string that happens to hold a value that looks like a float. While this may paper over the issue for now, it probably is not correct.

dvdplm:
Note that these funny looking float-strings have special meaning in a Slack context: they work as message IDs (for edits/deletions) so it is probably safest to leave them as strings and cast&parse only when needed.

PR #105, opened 2016-09-30
Attachment action confirm field is a hash not an array

AlphaB:
Hi, according to [slack doc](https://api.slack.com/docs/message-buttons#action_fields) `confirm` field for attachment action is a hash not an array. This causes 2 problems:
1. Unmarshalling of button webhook payload is falling with error `cannot unmarshal object into Go value of type []slack.ConfirmationField`.
2. Slack ignores provided `confirm` field and uses default values for confirmation window instead.

### DIFF
--- a/attachments.go
+++ b/attachments.go
@@ -12,12 +12,12 @@ type AttachmentField struct {
 // using message buttons and otherwise not useful. A maximum of 5 actions may be
 // provided per attachment.
 type AttachmentAction struct {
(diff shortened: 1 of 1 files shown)


wiyr (approved):

AlphaB:
@nlopes Hi! Any chances this PR will pass?

nlopes:
Looking into this now. Thanks!

PR #106, opened 2016-10-06
Add support for the "team.accessLogs" Slack API

ewestfal:
Included this in the "team.go" source file. I also wrote a test that is consistent with the style in the rest of the project.

Implements support in this client for the team.accessLogs API:

https://api.slack.com/methods/team.accessLogs

I have also manually verified this code functions properly when connecting to the actual Slack API


nlopes:
Thank you! I'm merging this one but would be nice if you could add the optional `before` argument as well ^^

Issue #80, opened 2016-06-23
how to access users.identity


aeneasr:
I need to see what user the token was granted for. Slack has the [users.identity](https://api.slack.com/methods/users.identity) endpoint for that, but it does not seem to be implemented here

tyndyll:
Doing this requires adding the identity scope to the OAuth request

nlopes:
Hey @arekkas. It has now been added. Thanks!

PR #110, opened 2016-10-18
adds users.identity api call and tests

selfcontained:
This adds support for the Slack `users.identity` api method: 

https://api.slack.com/methods/users.identity

It ends up being a pretty different response structure than other `user.*` methods, so I went the route of using dedicated structs.  You can request a few different `identity.*` scopes, and the fields I added to the structs encompass everything you could possibly get if you have all of the `identity` related scopes.

I also added a basic test for it.

This addresses #80 as well I believe.

### DIFF
--- a/users.go
+++ b/users.go
@@ -57,6 +57,39 @@ type UserPresence struct {
 	LastActivity    JSONTime `json:"last_activity,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


nlopes:
Thanks!

PR #112, opened 2016-10-28
check the HTTP status code and content type headers

tobert:
This change checks the response code and Content-Type header to make
sure the server returned a 200 first, then another check for the correct
content type.

There are various conditions where the Slack service will return a 504
(in my case) on file upload. Prior to this change those errors would
come back as a parsing error along the lines of: "invalid character '<'
looking for beginning of value", which is a bit confusing. Now a clear
error is returned and there are two layers of checking to make sure HTML
isn't passed to the parseResponseBody() function.

Also added a logResponse() function that uses net/http/httputil to dump
the response when in debug mode. net/http/httputil is already used in
websocket_proxy.go so it does not create a new dependency.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"mime/multipart"
 	"net/http"
+	"net/http/httputil"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


tobert:
The content-type check is broken. Making a fix now.

PR #120, opened 2016-11-28
added app user access token response struct

code0x9:
if using 'app user', oauth.access response struct has "user_id" and "incoming_webhook.channel_id". added them.


nlopes:
Thanks!

PR #122, opened 2016-12-06
simplify websocket example

jorinvo:
I'm not sure if I missed something but I think the example could be simplified to this.

I changed it to directly loop over the channel
and I replaced the "break with label" with a "return".

And thanks for creating this helpful library!


nlopes:
Cheers!

PR #127, opened 2017-01-17
Add function that returns rtm.IncomingEvents for external testing

oldmantaiter:
- When writing some tests that utilize this library, I need to mock
  out slack to test message handling.
- To create my testing interface I need a function to return the
  `IncomingEvents` channel


oldmantaiter:
Actually, don't need this, just realized I can do it in a cleaner way. Sorry for the noise!

PR #129, opened 2017-01-20
Add API to retrieve user's billing active status

kdamara:
Based on the API https://api.slack.com/methods/team.billableInfo

### DIFF
--- a/team.go
+++ b/team.go
@@ -51,7 +51,7 @@ type BillableInfoResponse struct {
 }
 
 type BillingActive struct {
-	BillingActive map[string]bool `json:"billing_active"`
+	BillingActive bool `json:"billing_active"`
 }
 
 // AccessLogParameters contains all the parameters necessary (including the optional ones) for a GetAccessLogs() request
--- a/team.go
+++ b/team.go
@@ -83,7 +83,7 @@ func teamRequest(path string, values url.Values, debug bool) (*TeamResponse, err
 	return response, nil
 }
 
-func billableInfoRequest(path string, values url.Values, debug bool) (*BillableInfoResponse, error) {
+func billableInfoRequest(path string, values url.Values, debug bool) (map[string]BillingActive, error) {
 	response := &BillableInfoResponse{}
 	err := post(path, values, response, debug)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


team.go
```diff
@@ -73,6 +83,20 @@ func teamRequest(path string, values url.Values, debug bool) (*TeamResponse, err
...
+	err := post(path, values, response, debug)
+	if err != nil {
+		return nil, err
+	}
+
+	if !response.Ok {
+		return nil, errors.New(response.Error)
+	}
+
+	return response, nil
```
nlopes: You could do `return response.BillableInfo, nil` here and adjust per below comments.

team.go
```diff
@@ -117,3 +141,31 @@ func (api *Client) GetAccessLogs(params AccessLogParameters) ([]Login, *Paging,
...
 }
 
+// GetBillableInfoForUser returns the billing_active status of a user by ID.
+func (api *Client) GetBillableInfoForUser(user string) (map[string]BillingActive, error) {
+	values := url.Values{
+		"token": {api.config.token},
+		"user": {user},
+	}
+
+	response, err := billableInfoRequest("team.billableInfo", values, api.debug)
```
nlopes: Instead of
```
+	response, err := billableInfoRequest("team.billableInfo", values, api.debug)
 +	if err != nil {
 +		return nil, err
 +	}
 +	return response.BillableInfo, nil
```

You can just do `return billableInfoRequest("team.billableInfo", values, api.debug)` if you adjust according to the previous comment.

team.go
```diff
@@ -117,3 +141,31 @@ func (api *Client) GetAccessLogs(params AccessLogParameters) ([]Login, *Paging,
...
+
+}
+
+// GetBillableInfoForTeam returns the billing_active status of all users on the team.
+func (api *Client) GetBillableInfoForTeam() (map[string]BillingActive, error) {
+	values := url.Values{
+		"token": {api.config.token},
+	}
+
+	response, err := billableInfoRequest("team.billableInfo", values, api.debug)
```
nlopes: Same here.

nlopes:
Hi @kdamara.

Thank you for the contribution!

One of the commits above has a token (hopefully it isn't valid, but still, I'd invalidate that token and rewrite your commit log and re-push)

I've also made a few comments that may simplify the whole PR a bit.

kdamara:
Hi @nlopes,

Thanks for your comments. I'm working on amending the PR, and will have a new one shortly.

Thanks,
Karen

kdamara:
@nlopes I've updated my pull request with your suggested changes. Please review, and once this pull request is merged, I will submit another one with an example. Thank you!

nlopes:
Thanks!

PR #133, opened 2017-01-29
Add thread_ts parameter when posting messages

afjoseph:
Allow sending `thread_ts` parameter when posting a message. Sending a message that has a `thread_ts` parameter that's the same as a the `ts` parameter of a message will thread the sent message. 

So, if message X had a `ts` value of `1485051909.018632`, any new messages that have a `thread_ts` value of `1485051909.018632` will become a child of message X and be listed under its thread.

More info [here](https://api.slack.com/docs/message-threading)

Quick code example:
```
func SendMessageToBot(message string) {
	if myBotStuff.channelId == "" {
		log.Println("channelId is not available")
		return
	}

	params := slack.PostMessageParameters{}
	params.Markdown = true

// >>>>>>>>>>>
// params.ThreadTimestamp maps to `thread_ts`
	if myBotStuff.lastTimestamp != "" {
		params.ThreadTimestamp = myBotStuff.lastTimestamp
		log.Printf("sending message with ts: %s", params.ThreadTimestamp)
	}
// <<<<<<<<<<<<
	_, _, err := myBotStuff.api.PostMessage(myBotStuff.channelId, message, params)
	if err != nil {
		fmt.Printf("%s\n", err)
		return
	}
}

```

The code above is tested and working :)

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -8,16 +8,17 @@ import (
 )
 
 const (
-	DEFAULT_MESSAGE_USERNAME     = ""
-	DEFAULT_MESSAGE_ASUSER       = false
-	DEFAULT_MESSAGE_PARSE        = ""
-	DEFAULT_MESSAGE_LINK_NAMES   = 0
-	DEFAULT_MESSAGE_UNFURL_LINKS = false
-	DEFAULT_MESSAGE_UNFURL_MEDIA = true
-	DEFAULT_MESSAGE_ICON_URL     = ""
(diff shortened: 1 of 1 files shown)


LouisTrezzini:
Btw we also need a `ThreadTS` field in the `Msg` struct

khalilovcmd:
@nlopes any updates on merging this?

nlopes:
Merging and then looking into the `Msg` one.

PR #135, opened 2017-02-02
Use UNIX timestamp for files.list API

najeira:
Parameters `ts_from` and `ts_to` should be UNIX timestamp.

See examples of https://api.slack.com/methods/files.list.


dlsniper:
Thank you for this!

PR #137, opened 2017-02-14
Add fields to the EmojiChangedEvent

stefafafan:
As seen in the API documents of the emoji_changed event, we should be able to distinguish whether or not it is an `add` or a `remove`. Also, we should be able to know which emojis are added/removed.
https://api.slack.com/events/emoji_changed

The `EmojiChangedEvent` struct did not have these fields ( `subtype` , `name` , `names` ) so I added them.

### DIFF
--- a/websocket_misc.go
+++ b/websocket_misc.go
@@ -80,6 +80,7 @@ type EmojiChangedEvent struct {
 	SubType        string   `json:"subtype"`
 	Name           string   `json:"name"`
 	Names          []string `json:"names"`
+	Value          string   `json:"value"` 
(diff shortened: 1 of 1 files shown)


dlsniper (changes requested):
@stefafafan Thank you for the PR.
Would you be interested to also add the ` Value ` field to this so it can be a complete event? Otherwise I can merge it as it is and add it after.

Thank you!

stefafafan:
@dlsniper I added `Value` to the struct, does this look ok?
Thanks for the review!

dlsniper (approved):
This looks good. Thank you for the PR!

PR #138, opened 2017-02-15
Add threaded timestamp to messages

tso:
Correct me if I'm mistaken but there didn't appear to be a way to determine if a received message was within a thread or not without this.


dlsniper:
@tristan-ohlson I'm so so sorry I've missed your PR when I've merged #149. I'll try to be more careful next time.
Please don't let that discourage you and do send more PRs in the future.
Thank you!

PR #145, opened 2017-03-21
JSON error when trying to parse plain text response body

mrf:
I was hitting rate limits with api requests to getUsers and getting back the following:
`invalid character 'Y' looking for beginning of value` as the error message.

Since we can't rely on a JSON response in all cases this bugs out if we see anything but a 200 and returns a more useful status code instead `Slack server error: 429 Too Many Requests.`


james-lawrence:
https://github.com/nlopes/slack/pull/121

PR #121, opened 2016-11-30
refactor PostMessage/UpdateMessage, add SendMessage.

james-lawrence:
- allows providing attachments when updating a message. (via SendMessage)
- makes the usage for sending a message more general.

Fixes #145, #146

PR #146, opened 2017-03-25
Implement method UpdateMessageWithAttachments

mkungla:
Added UpdateMessageParameters which  contains all the parameters necessary (including the optional ones) for a UpdateMessageWithAttachments() request which updates a message in a channel with attachments. I added new method so that it would not become breaking change for others.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -45,6 +45,16 @@ type PostMessageParameters struct {
 	EscapeText      bool         `json:"escape_text"`
 }
 
+// UpdateMessageParameters contains all the parameters necessary (including the optional ones) for a UpdateMessage() request
+type UpdateMessageParameters struct {
+	Timestamp   string       `json:"ts"`
+	Text        string       `json:"text"`
+	Attachments []Attachment `json:"attachments"`
+	Parse       string       `json:"parse"`
+	LinkNames   int          `json:"link_names"`
+	AsUser      bool         `json:"as_user"`
+}
+
 // NewPostMessageParameters provides an instance of PostMessageParameters with all the sane default values set
 func NewPostMessageParameters() PostMessageParameters {
 	return PostMessageParameters{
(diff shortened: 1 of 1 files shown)


james-lawrence:
see https://github.com/nlopes/slack/pull/121

mkungla:
@james-lawrence yours will introduce braking changes right? Anyhow seems that @nlopes is not up to maintain this repository or does not have enough time for it which forces me to develop further under my fork or re design that API Client from scratch since I find that API client should be more consistent with the endpoints scema instead of inventing new method names and parameter stucts which do not match with endpoint. Here are some example What I mean:

| actual endpoint | method here | better would be |
| --- | --- | --- |
| `bots.info` | `GetBotInfo` | `BotsInfo` |
| `channels.archive` | `ArchiveChannel` | `ChannelsArchive | Channels.Archive` |
| `channels.unarchive` | `UnarchiveChannel` | `ChannelsUnarchive | Channels.Unarchive` |
| dnd.endDnd | EndDND | `DndEndDnd | Dnd.EndDnd` |

and so on.

james-lawrence:
mine shouldn't introduce any api changes, if I did they were unintentional. I added a new method that the old ones delegate down to.

I'm in general not opposed to breaking changes if it forwards a library, as long as the library is versioned. In my own libraries I would have changed the signature of PostMessage instead of adding the new method.

As for the naming, I'm fairly indifferent long as they are clear. My issues main issues revolve around
- RTM not having a safe shutdown/startup sequence.
- the http client being completely inaccessible. (it should be something that can be configured and passed in.
- private member structs requiring setting by struct.field = instead of struct{field:}

dlsniper:
I'll have a review of this and #121 soon (hopefully by the end of the weekend).
As far as I can see #121 doesn't introduce any API changes but it's a pretty decent sized one and I'd rather be safe than sorry.

dlsniper:
Fixed in #121, sorry for the delay.

PR #148, opened 2017-04-02
Add users.setPhoto and users.deletePhoto

joshheinrichs:
This adds basic support for users.setPhoto and users.deletePhoto. I was debating whether the image should be provided as a file path or an [image.Image](https://golang.org/pkg/image/#Image). A file path has the benefit of being more agnostic. I don't think that we should be responsible for how the image an gets encoded before being uploaded.


dlsniper:
Hi @joshheinrichs,

Thank you for the PR.
I'll try and review it and get it merged this weekend.

PR #76, opened 2016-06-06
Groundwork for supporting `usergroups` and `usergroups.users` methods

YoSmudge:
RE #63 

Not yet complete but base for the methods
- `usergroups.list`
- `usergroups.update`
- `usergroups.users.list`
- `usergroups.users.update`

I needed these specific functions so prioritised them, but will try to look at adding the rest at some point this week if I get chance, but wanted to shoot this back over anyway.

PR #150, opened 2017-04-06
Usergroups

### LINKED ISSUE #63: Add support for usergroups methods

whiteley:
Trying to continue from the previous work to support usergroups

Closes #63 #76

### DIFF
--- a/usergroups.go
+++ b/usergroups.go
@@ -3,17 +3,21 @@ package slack
 import (
 	"errors"
 	"net/url"
+	"strings"
 )
 
-// User contains all the information of a user group
+// UserGroup contains all the information of a user group
 type UserGroup struct {
 	ID          string `json:"id"`
-	Name        string `json:"name"`
-	TeamId      string `json:"team_id"`
+	TeamID      string `json:"team_id"`
 	IsUserGroup bool   `json:"is_usergroup"`
+	Name        string `json:"name"`
 	Description string `json:"description"`
 	Handle      string `json:"handle"`
 	IsExternal  bool   `json:"is_external"`
+	DateCreate  int    `json:"date_create"`
+	DateUpdate  int    `json:"date_update"`
+	DateDelete  int    `json:"date_delete"`
 	AutoType    string `json:"auto_type"`
 	CreatedBy   string `json:"created_by"`
 	UpdatedBy   string `json:"updated_by"`
@@ -23,7 +27,7 @@ type UserGroup struct {
 		Groups   []string `json:"groups"`
 	} `json:"prefs"`
 	Users     []string
-	UserCount int `json:"user_count"`
+	UserCount int `json:"user_count,string"`
 }
 
 type userGroupResponseFull struct {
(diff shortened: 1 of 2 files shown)


usergroups.go
```diff
@@ -0,0 +1,172 @@
...
+	Description string `json:"description"`
+	Handle      string `json:"handle"`
+	IsExternal  bool   `json:"is_external"`
+	DateCreate  int    `json:"date_create"`
+	DateUpdate  int    `json:"date_update"`
+	DateDelete  int    `json:"date_delete"`
+	AutoType    string `json:"auto_type"`
+	CreatedBy   string `json:"created_by"`
+	UpdatedBy   string `json:"updated_by"`
+	DeletedBy   string `json:"deleted_by"`
```
dlsniper: According to the documentation here: https://api.slack.com/types/usergroup ` deleted_by ` can be ` null ` so I guess we need to use ` *string ` as type here.

usergroups.go
```diff
@@ -0,0 +1,172 @@
...
+// UserGroup contains all the information of a user group
+type UserGroup struct {
+	ID          string `json:"id"`
+	TeamID      string `json:"team_id"`
+	IsUserGroup bool   `json:"is_usergroup"`
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Handle      string `json:"handle"`
+	IsExternal  bool   `json:"is_external"`
+	DateCreate  int    `json:"date_create"`
```
dlsniper: This should be of type ` JSONTime `

usergroups.go
```diff
@@ -0,0 +1,172 @@
...
+	DateUpdate  int    `json:"date_update"`
```
dlsniper: This should be of type ` JSONTime `

usergroups.go
```diff
@@ -0,0 +1,172 @@
...
+	DateDelete  int    `json:"date_delete"`
```
dlsniper: This should be of type ` JSONTime `

usergroups.go
```diff
@@ -0,0 +1,172 @@
...
+	AutoType    string `json:"auto_type"`
+	CreatedBy   string `json:"created_by"`
+	UpdatedBy   string `json:"updated_by"`
+	DeletedBy   string `json:"deleted_by"`
+	Prefs       struct {
```
dlsniper: This should be extracted in its own type I think so that it's easier to reference / maintain in the future.

dlsniper (changes requested):
Hi @whiteley, 

Thanks a lot for your PR.
There are a few changes that should be made before this can be merged.
I really appreciate your PR and the presence of tests.

Thank you.

usergroups.go
whiteley: `deleted_by` can be null or an ID but not an empty string as well, it seemed like that would only be necessary if distinguishing between all three?

usergroups.go
```diff
@@ -0,0 +1,175 @@
...
+	IsExternal  bool           `json:"is_external"`
+	DateCreate  JSONTime       `json:"date_create"`
+	DateUpdate  JSONTime       `json:"date_update"`
+	DateDelete  JSONTime       `json:"date_delete"`
+	AutoType    string         `json:"auto_type"`
+	CreatedBy   string         `json:"created_by"`
+	UpdatedBy   string         `json:"updated_by"`
+	DeletedBy   string         `json:"deleted_by"`
+	Prefs       UserGroupPrefs `json:"prefs"`
+	Users       []string
```
dlsniper: I'm not sure if this is needed by the looks of it as ` userGroupResponseFull ` already has another ` Users ` and it's tagged while this isn't used anywhere.

usergroups.go
dlsniper: As we don't need to encode those I guess it's ok, fair point.

dlsniper (changes requested):
Hi @whiteley, really sorry for this one last thing, I've missed in my original review. Thank you so much for this.

dlsniper:
Turns out I can use a cool new Github feature to help on this.
Thanks a lot for the PR @whiteley, awesome job!

dlsniper (approved):

PR #154, opened 2017-04-16
Add ThreadTimestamp to OutgoingMessage

kechako:
I want to reply to a thread using the (*RTM)SendMessage method.
But the OutgoingMessage struct doesn't have a field of ThreadTimestamp.


dlsniper:
Thank you!

PR #155, opened 2017-04-17
add support for setting and unsetting Custom Status/Emoji

### LINKED ISSUE #153: Support setting custom status messages on User's profile
This issue is to track supporting the new Custom Status messages as recently unveiled by the Slack team:

- https://slackhq.com/set-your-status-in-slack-28a793914b98

This is implemented by setting `status_text` and `status_emoji` values on the User's profile[1], which could maybe be easily implemented by adding two new struct fields to `UserProfile`. However, the status messages (and Emojis) are unset by setting both fields to empty string (`""`), which poses an interesting challenge with the `omitempty` needed to avoid accidentally unsetting the status if the fields are intentionally omitted.

[1] https://api.slack.com/docs/presence-and-status#custom_status

theckman:
This change introduces two new methods on *Client for setting and unsetting the
Custom Status and Status Emoji for the currently authenticated User. This also
adds `StatusText` and `StatusEmoji` fields to the `UserProfile` struct. The
Custom Status and Status Emoji functionality was recently added as part of the
User's profile[1].

The first method added is `SetUserCustomStatus()` which takes two string
arguments. This method uses an anonymous struct to generate the JSON required to
update a User's Custom Status and Status Emoji. If both arguments are set to
empty-string (`""`) the Slack API will unset the Custom Status/Emoji.

The second method is `UnsetUserCustomStatus()`, which is a convenience method
that takes zero arguments and wraps `SetUserCustomStatus()`. It unsets the
custom status.

Integration tests were added to ensure that that setting and unsetting the
Custom Status should work as-expected.

Closes #153.

[1] https://slackhq.com/set-your-status-in-slack-28a793914b98

Signed-off-by: Tim Heckman <t@heckman.io>

### DIFF
--- a/users.go
+++ b/users.go
@@ -1,6 +1,7 @@
 package slack
 
 import (
+	"encoding/json"
 	"errors"
 	"net/url"
 )
@@ -29,6 +30,8 @@ type UserProfile struct {
 	Title              string `json:"title"`
 	BotID              string `json:"bot_id,omitempty"`
 	ApiAppID           string `json:"api_app_id,omitempty"`
+	StatusText         string `json:"status_text,omitempty"`
+	StatusEmoji        string `json:"status_emoji,omitempty"`
 }
 
 // User contains all the information of a user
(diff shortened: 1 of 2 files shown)


JulienBreux:
Very nice work!

A small rebase from `master` and let's go :)

theckman:
Thanks! Just rebased and re-signed the commit.

dlsniper:
Thank you!

Issue #142, opened 2017-03-03
using api.postMessage doesn't post from authenticated user


jhebb-lm:
Using the API with a slackbot, If I send a message using `rtm.SendMessage(rtm.NewOutgoingMessage("Hello World", channel))` it posts from my bot as expected but if I try and use an attachment with `api.PostMessage(channel, "Hello World", params)` it posts from a generic bot user - is there something else I need to do post from my own bot using an attachment?

I followed the example here:
https://github.com/nlopes/slack/blob/master/examples/messages/messages.go

Davejkane:
You need to set params.AsUser as true.

params := slack.NewPostMessageParameters
params.AsUser = true
api.PostMessage(channel, "Hello World", params)

gemina:
@Davejkane I tried like this also and got the same error, message displayed as "bot"

Davejkane:
Is that perhaps because I accidentally left the parenthesis off? should be
    params := slack.NewPostMessageParameters()

gemina:
Same, I tried that one also

Davejkane:
Is it at all possible to share your code? Obviously not your bot token.

dlsniper:
Hi,

I've merged #146 without noticing a regression on the AsUser option.
Luckily @vrischmann was quick to identify it and fix it in #156 so please try this again. I've verified it using https://github.com/gopheracademy/gopher/blob/master/bot/bot.go#L121 which should also serve as an example on how to use this.
Not sure if it solves the initial issue but I think it should as well.
Thank you and I apologize for the issue, testing is something I know I need to improve in the future for this package.

gemina:
Thank you

james-lawrence:
sounds like this was fixed. closing.

PR #156, opened 2017-04-21
correctly apply the AsUser parameter

vrischmann:
This correctly applies the AsUser MsgOption.

I noticed this because my bot always appeared in slack as "bot" without an icon, despite AsUser being set. I added some log and saw that `as_user` was never set in the `config.values` map.

I believe #142 is the same bug too.


dlsniper (approved):
Thank you!

james-lawrence:
oops, my bad.

PR #157, opened 2017-04-21
Add ConnectRTM call to use the rtm.connect endpoint

dvdplm:
Slack now provides `rtm.connect` for faster WS connection (details: https://api.slack.com/changelog/2017-04-start-using-rtm-connect-and-stop-using-rtm-start)

This implements `ConnectRTM` much in the same vein as `StartRTM` with the difference of passing back the `WebResponse` also. This is useful to be able to know when Slack is returning a 429 rate limitation response (not very elegant though... what's a better way?).

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -17,7 +17,7 @@ const (
 // RTM represents a managed websocket connection. It also supports
 // all the methods of the `Client` type.
 //
-// Create this element with Client's NewRTM().
+// Create this element with Client's NewRTM() or NewRTMWithOptions(*RTMOptions)
 type RTM struct {
 	idGen IDGenerator
 	pings map[int]time.Time
@@ -38,23 +38,23 @@ type RTM struct {
 
 	// UserDetails upon connection
 	info *Info
+
+	// useRTMStart should be set to true if you want to use
+	// rtm.start to connect to Slack, otherwise it will use
+	// rtm.connect
+	useRTMStart bool
 }
 
-// NewRTM returns a RTM, which provides a fully managed connection to
-// Slack's websocket-based Real-Time Messaging protocol.
-func newRTM(api *Client) *RTM {
-	return &RTM{
-		Client:           *api,
-		IncomingEvents:   make(chan RTMEvent, 50),
-		outgoingMessages: make(chan OutgoingMessage, 20),
-		pings:            make(map[int]time.Time),
-		isConnected:      false,
-		wasIntentional:   true,
-		killChannel:      make(chan bool),
-		forcePing:        make(chan bool),
-		rawEvents:        make(chan json.RawMessage),
-		idGen:            NewSafeID(1),
-	}
+// RTMOptions allows configuration of various options available for RTM messaging
+//
+// This structure will evolve in time so please make sure you are always using the
+// named keys for every entry available as per Go 1 compatibility promise adding fields
+// to this structure should not be considered a breaking change.
+type RTMOptions struct {
+	// UseRTMStart set to true in order to use rtm.start or false to use rtm.connect
+	// As of 11th July 2017 you should prefer setting this to false, see:
+	// https://api.slack.com/changelog/2017-04-start-using-rtm-connect-and-stop-using-rtm-start
+	UseRTMStart bool
 }
 
 // Disconnect and wait, blocking until a successful disconnection.
--- a/misc.go
+++ b/misc.go
@@ -108,6 +108,7 @@ func postWithMultipartResponse(path, filepath, fieldname string, values url.Valu
 }
 
 func postForm(endpoint string, values url.Values, intf interface{}, debug bool) error {
+	const retryAfterHeader = "Retry-After"
 	resp, err := HTTPClient.PostForm(endpoint, values)
 	if err != nil {
 		return err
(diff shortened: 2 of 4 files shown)


dlsniper:
Hi,

Thanks for the PR.
I've been reading: https://api.slack.com/changelog/2017-04-start-using-rtm-connect-and-stop-using-rtm-start from what I can there, this is supposed to be a replacement for ` rtm.start`.
As such, I believe that we also need to add a different way to start the connection now, by calling this method that you've added in this PR which would replace ` startRTMAndDial() ` method (from ` websocket_managed_conn.go `).
I'll work on adding those changes in before merging but I'm not sure just yet about the BC break from the previous commit. I want to work on the library and fix all the outstanding issues that the current static analysis tools find and that would be a proper moment to introduce such a break. Ideally I'd like this to support only the current and previous releases of Go as a model of support, but I've not decided yet and it's certainly not something I want to think now, sorry.

Thanks a lot for your hard work on this one, I'll make sure it's put to good use.

dvdplm:
Hi, thank you for your kind words. I agree with you that handling rate-limiting errors with a string in the error message is not good. 

How about a custom error type? Something like:

```
type SlackErrorTooManyRequests struct {
	RetryAfter int
	Err        error
}
```

To give some context, Slacks rate limiting became very much more aggressive just the last few days and we (all?) will need to deal with it soon.

dlsniper:
Hmm, I'd rather build a more generic Slack error than this one and make it implement the standard ` error ` interface so that people can assert for the error, thus not having to break the API for them. I understand that this is something that you, and others as well, are going to needs this sooner rather than later, so I'll try and think on this as a top priority.

Meanwhile, I've split this PR into two PRs and opened #159 for the error specific bits, and if I've done this correctly, with the help of a bit of magic from Github, this will be updated with the code that enables the ` rtm.connect ` call.

dlsniper:
I've finished the refactoring and now, from the client point of view, it looks like this:

```
-       slackBotRTM := slackBotAPI.NewRTM()
+       slackBotRTM := slackBotAPI.NewRTMWithOptions(&slack.RTMOptions{})
        go slackBotRTM.ManageConnection()
```

Using the old ` NewRTM() ` works and keeps the default ` rtm.start ` connection. passing a ` nil ` instead of a pointer to ` RTMOptions ` also keeps the old default in place.

I'll merge this tomorrow if it looks ok

websocket.go
```diff
@@ -38,23 +38,23 @@ type RTM struct {
...
+// RTMOptions allows configuration of various options available for RTM messaging
+//
+// This structure will evolve in time so please make sure you are always using the
+// named keys for every entry available as per Go 1 compatibility promise adding fields
+// to this structure should not be considered a breaking change.
+type RTMOptions struct {
+	// UseRTMStart set to true in order to use rtm.start or false to use rtm.connect
+	// As of 11th July 2017 you should prefer setting this to false, see:
+	// https://api.slack.com/changelog/2017-04-start-using-rtm-connect-and-stop-using-rtm-start
+	UseRTMStart bool
```
dvdplm, line 57: Possibly add bools to handle the new `no_latest=1` param? That is what Slack promotes as the migration path for users currently stuck on `rtm.start`.

dlsniper: There's no need for that as we allow the usage of the ` rtm.start ` right now and as far as I can understand this is the only difference, for now.

jyap808:
Hmm, having some issues when I use this. It has some endless ping loop.

```
Event Received: nlopes/slack2017/11/03 18:22:13 misc.go:82: parseResponseBody: {"ok":true,"url":"wss:\/\/mpmulti-xxxx.lb.slack-msgs.com\/websocket\/<REMOVED>=\/2?dp=1","team":{"id":"<REMOVED>","name":"<REMOVED>","domain":"<REMOVED>"},"self":{"id":"<REMOVED>","name":"<REMOVED>-bot"}}
nlopes/slack2017/11/03 18:22:13 slack.go:86: Using URL: wss://mpmulti-xxxx.lb.slack-msgs.com/websocket/<REMOVED>=/2?dp=1
Event Received: Infos: &{wss://mpmulti-xxxx.lb.slack-msgs.com/websocket/<REMOVED>=/2?dp=1 0xc42025dcc0 0xc42055e0f0 [] [] [] [] []}
Connection counter: 1
nlopes/slack2017/11/03 18:22:14 slack.go:86: Incoming Event: {}
Event Received: nlopes/slack2017/11/03 18:22:14 slack.go:86: Sending PING  1
nlopes/slack2017/11/03 18:22:14 slack.go:86: Sending PING  2
nlopes/slack2017/11/03 18:22:14 slack.go:86: Sending PING  3
nlopes/slack2017/11/03 18:22:14 slack.go:86: Sending PING  4
nlopes/slack2017/11/03 18:22:14 slack.go:86: Sending PING  5
nlopes/slack2017/11/03 18:22:14 slack.go:86: Sending PING  6
nlopes/slack2017/11/03 18:22:14 slack.go:86: Sending PING  7
nlopes/slack2017/11/03 18:22:14 slack.go:86: Sending PING  8
nlopes/slack2017/11/03 18:22:14 slack.go:86: Sending PING  9
```

PR #159, opened 2017-04-24
Extract Retry-After header value and return specific error

dlsniper:


dvdplm:
@dlsniper let me know if I can help out with this one in any way. In the past we've done something like this:

```go
type SlackErrorTooManyRequests struct {
	RetryAfter int
	Err        error
}

func (connErr *SlackErrorTooManyRequests) Error() string {
	if connErr == nil {
		return "<nil>"
	}
	if connErr.RetryAfter == 0 {
		return connErr.Err.Error()
	}
	return connErr.Err.Error() + fmt.Sprintf(" Retry after %d", connErr.RetryAfter)
}
```
…and then used it like this:

```go
	err := ConnectToSlackCode(param, param2, param3)
	if err != nil {
		if connErr, ok := err.(*SlackErrorTooManyRequests); ok {
			time.Sleep(time.Duration(connErr.RetryAfter) * time.Second)
                        // retry here
                 }
                 // handle other errors here
          }
```

dvdplm:
@dlsniper @nlopes  this stuff is continuing to bite us. Anything we can do to help move this along? :/

dlsniper:
The reason I don't want to push this is because it should be done better to flag the issue and not have to match / parse the error text. Unfortunately I'm busy with GopherCon and that's why the late reply.

dvdplm:
@dlsniper sure I get that this is not the right path. I am not sure what the alternative plan is though. If you could share your thoughts on what the proper way to do this is, perhaps I can come up with a PR that have a chance at getting merged?

catherinetcai:
@dlsniper @dvdplm Is this still a WIP? I think we want to start closing out our PRs, so if there isn't any movement on this front I'm going to close this out.

dvdplm:
Imo this is better than nothing and I'd prefer it get merged as-is rather than closed. Currently there is no way for the caller to know if they're being rate-limited and for how long. At least with this, interested clients can choose to do string matching and take action.

james-lawrence:
closing due to #186 getting merged.

PR #161, opened 2017-05-04
Add support of channels.replies API method to channels

megane42:
### Overview

This PR adds support of [channels.replies](https://api.slack.com/methods/channels.replies) Web API method. 
Note that the sample response written in the above link is wrong. The correct example is shown at [here](https://api.slack.com/docs/message-threading#retrieving_message_replies).

### Demo

```golang
package main

import (
	"fmt"
	"github.com/megane42/slack"
	"github.com/k0kubun/pp"
)

func main() {
	api := slack.New("YOUR_TOKEN_HERE")
	channels, err := api.GetChannelReplies("CHANNEL_ID_HERE", "PARENT_THREAD_TS_HERE")
	if err != nil {
		fmt.Printf("%s\n", err)
		return
	}

	pp.Print(channels)
}
```

The result of the above code is [here](https://gist.github.com/megane42/5e81e009e392eb16acdc13c7e4634f38).

### DIFF
--- a/channels.go
+++ b/channels.go
@@ -259,3 +259,17 @@ func (api *Client) SetChannelTopic(channel, topic string) (string, error) {
 	}
 	return response.Topic, nil
 }
+
+// GetChannelReplies gets an entire thread (a message plus all the messages in reply to it).
+func (api *Client) GetChannelReplies(channel, thread_ts string) ([]Message, error) {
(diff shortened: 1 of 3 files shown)


dlsniper:
Thank you for the PR and sorry for the delay to merge this.

PR #162, opened 2017-05-05
Fix errors reported by staticcheck

kevinburke:
Specifically:

examples/team/team.go:20:25: this value of err is never used (SA4006)
misc.go:94:7: this value of err is never used (SA4006)
websocket_proxy.go:34:8: httputil.NewProxyClientConn is deprecated: Use the Client or Transport in package net/http instead.  (SA1019)
websocket_proxy.go:36:26: httputil.ErrPersistEOF is deprecated: No longer used.  (SA1019)

I wasn't sure how to replace the NewProxyClientConn so I left it
alone. It's possible we can just call http.NewRequest and then use
a Transport that implements http.ProxyURL.

### DIFF
--- a/users.go
+++ b/users.go
@@ -198,10 +198,7 @@ func (api *Client) SetUserAsActiveContext(ctx context.Context) error {
 		"token": {api.config.token},
 	}
 	_, err := userRequest(ctx, "users.setActive", values, api.debug)
-	if err != nil {
-		return err
-	}
-	return nil
+	return err
 }
 
 // SetUserPresence changes the currently authenticated user presence
@@ -245,7 +242,7 @@ func (api *Client) GetUserIdentityContext(ctx context.Context) (*UserIdentityRes
 }
 
 // SetUserPhoto changes the currently authenticated user's profile image
-func (api *Client) SetUserPhoto(ctx context.Context, image string, params UserSetPhotoParams) error {
+func (api *Client) SetUserPhoto(image string, params UserSetPhotoParams) error {
 	return api.SetUserPhotoContext(context.Background(), image, params)
 }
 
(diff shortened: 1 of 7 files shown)


websocket_proxy.go
```diff
@@ -33,7 +33,7 @@ func websocketHTTPConnect(proxy, urlString string) (net.Conn, error) {
 
 	cc := httputil.NewProxyClientConn(p, nil)
 	cc.Do(&req)
```
dvdplm: I think you need to change this to `response, err := cc.Do(&req)` or else the error check is just a repetition of line 23 right?

catherinetcai:
@kevinburke Is this still a WIP or can I close this out?

kevinburke:
Apologies, this sliped off my radar, please take another look.

users.go
```diff
@@ -246,7 +243,7 @@ func (api *Client) GetUserIdentityContext(ctx context.Context) (*UserIdentityRes
 
 // SetUserPhoto changes the currently authenticated user's profile image
 func (api *Client) SetUserPhoto(ctx context.Context, image string, params UserSetPhotoParams) error {
-	return api.SetUserPhoto(context.Background(), image, params)
```
kevinburke: This is definitely broken, anyone who currently calls this method will get an infinite loop. For consistency the API *should* be:

```
SetUserPhoto(image string, params UserSetPhotoParams)
```

but that would be a breaking change.

dvdplm: I think we can safely assume nobody is using this, so please go ahead and make the suggested change. After that this looks ready to merge imo. @catherinetcai agree?

kevinburke: Done

Issue #27, opened 2015-06-23
Enhancement: Allow externally-provided *http.Client


joelgwebber:
I need to use this library from an AppEngine frontend, which necessitates the use of Google's urlfetch API, which provides a custom `*http.Client`. I have it patched on my end (`slack.Slack` gets an extra `config *http.Client` field that's initialized in `slack.New()`), but I wanted to be sure you were receptive to such a change before submitting a patch.

joelgwebber:
And of course if you'd prefer a different approach, I'm happy to adapt to whatever style you prefer.

derekperkins:
If he uses the http.DefaultClient, you could modify that on init and then it should work with urlfetch.

joelgwebber:
True enough, but that doesn't work in environments like AppEngine, where you can't rely on setting http.DefaultClient.

nlopes:
Let me take a look at this. I had lost track of this ticket, apologies for that.

nlopes:
@joelgwebber I've added some breaking changes to the library today so I'll merge this in if you feel generous enough to send a patch.

joelgwebber:
Sure thing. Let me pull in your changes to my fork and get a patch together. Hopefully over the next day or two.

dvdplm:
Did this merge happen? We'd love to be able to use something like facebooks httpcontrol (https://godoc.org/github.com/facebookgo/httpcontrol) to get retries and stats. :)

john-difool:
I would love to hear if any progress has been made on this one?

nlopes:
It didn't. I never got back a patch for it afaict. Will look at adding a patch myself this week.

joelgwebber:
So sorry this completely slipped my mind for ages. I'm updating our fork now and will send a PR by the end of the week. It's a pretty simple, but pervasive change to push the `http.Client` parameter everywhere, so the change will look larger than it actually is, conceptually.

joelgwebber:
https://github.com/nlopes/slack/pull/68

dlsniper:
There's now an ` HTTPClient ` package level variable, which can be changed so that it can use a custom http.Client fixed in #73. I'll close this for now.

PR #163, opened 2017-05-12
remove global http client requirement

james-lawrence:
cleaned up version of #68 and #85 for #27. the package level scope isn't enough. we run a number of clients simultaneously and sometimes need to enable debugging of the http req/resp on a per client basis.

- allows for per instance http client configuration, useful for setting a client with a customized transport/timeouts etc.
- backwards compatible.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -17,20 +17,6 @@ import (
 	"time"
 )
 
-// HTTPRequester defines the minimal interface needed for an http.Client to be implemented.
-//
-// Use it in conjunction with the SetHTTPClient function to allow for other capabilities
-// like a tracing http.Client
-type HTTPRequester interface {
-	Do(*http.Request) (*http.Response, error)
-}
-
-var customHTTPClient HTTPRequester
-
-// HTTPClient sets a custom http.Client
-// deprecated: in favor of SetHTTPClient()
-var HTTPClient = &http.Client{}
-
 type WebResponse struct {
 	Ok    bool      `json:"ok"`
 	Error *WebError `json:"error"`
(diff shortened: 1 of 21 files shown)


dlsniper:
Sorry for the delay.
This is technically an API breakage and I'd rather not do this now.
I plan to tag this as a 0.9 at some point and then refactor everything out to have a better handling of the http.Client

misc.go
```diff
@@ -16,8 +16,6 @@ import (
 	"time"
 )
 
-var HTTPClient = &http.Client{}
```
dlsniper: Removing this line causes an API break.

james-lawrence: look again sorry I moved it to slack.go. sorry

oauth.go
```diff
@@ -45,7 +46,7 @@ func GetOAuthResponse(clientID, clientSecret, code, redirectURI string, debug bo
 		"redirect_uri":  {redirectURI},
 	}
 	response := &OAuthResponse{}
-	err = post("oauth.access", values, response, debug)
```
james-lawrence: should probably change this to HTTPClient instead of http.DefaultClient

james-lawrence:
@dlsniper I disgree that it is an api breakage. I simply moved the HTTPClient to slack.go. misc.go felt like a random place to put it originally and I had trouble finding it.

I'm happy to rebase this again if you're willing to accept the change.

dlsniper:
Hi, I'll have to look at it again, let me come back to you at the end of the week please.

james-lawrence:
updated against latest master.

not a huge fan of the new customHTTPRequester adds in another global but it didn't change the PR much.
I did move the customHTTPRequester next to the HTTPClient in slack.go so they are near each other so the interaction is obvious from the code.

I also removed the getHTTPRequester() method as it is no longer necessary.

ernestoalejo:
This will help running this package in App Engine too. We need to pass a custom client with a context different for each request; we can't pass or initialize a global HTTP client in App Engine.

james-lawrence:
@dlsniper any update?

james-lawrence:
github isnt detecting my updates fixing the conflicts. closing and reopening.

PR #164, opened 2017-05-19
file.upload: fix error handling for multipart

dennwc:
handle error correctly in postWithMultipartResponse


dlsniper:
Nice catch! Thank you for the PR and sorry for the delay.

PR #165, opened 2017-05-20
file.upload: allow uploads with io.Reader

dennwc:
This change is [Reviewable](https://reviewable.io/reviews/nlopes/slack/165)

### DIFF
--- a/files.go
+++ b/files.go
@@ -2,6 +2,7 @@ package slack
 
 import (
 	"errors"
+	"io"
 	"net/url"
 	"strconv"
 	"strings"
@@ -90,6 +91,7 @@ type File struct {
 type FileUploadParameters struct {
 	File           string
 	Content        string
+	Reader         io.Reader
 	Filetype       string
 	Filename       string
 	Title          string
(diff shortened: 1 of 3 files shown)


dlsniper:
Thanks for the PR! Unfortunately it's not mergeable anymore due to a previous PR (I suspect it was yours to fix another error). Would you mind fixing the issues here so that this can be merged again?
Also please add a bit of documentation on how the user is expected to set the ` Reader ` field instead of the ` File ` field.
Thank you !

dlsniper:
As a small heads-up, I'll push shortly an update to allow setting of context for the API which should allow full usage of the https://godoc.org/cloud.google.com/go/trace#HTTPClient and other clients.

dennwc:
Done. Please take a look.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlopes/slack/165#-:-KlErxKUwkr1HrJkbzxo:b-29aia4)*

dlsniper:
LGTM, thank you for the quick reply.

PR #169, opened 2017-06-29
Added exclude_members option to channel list

brentahughes:
As slack teams grow to large levels of channels and members the api call to get the list of channels is getting slower and slower. A very large part of this slow down is it returns the list of members for each channel. 

In order to prevent timeouts from the slack api we have to use exclude_members in the channels.list call to cut down on the large response.

**It should be noted that this change will break any existing use of slack.Channels.GetChannels()**

### DIFF
--- a/channels.go
+++ b/channels.go
@@ -228,17 +228,20 @@ func (api *Client) KickUserFromChannelContext(ctx context.Context, channel, user
 }
 
 // GetChannels retrieves all the channels
-func (api *Client) GetChannels(excludeArchived bool) ([]Channel, error) {
-	return api.GetChannelsContext(context.Background(), excludeArchived)
+func (api *Client) GetChannels(excludeArchived bool, excludeMembers bool) ([]Channel, error) {
+	return api.GetChannelsContext(context.Background(), excludeArchived, excludeMembers)
 }
 
 // GetChannelsContext retrieves all the channels with a custom context
-func (api *Client) GetChannelsContext(ctx context.Context, excludeArchived bool) ([]Channel, error) {
+func (api *Client) GetChannelsContext(ctx context.Context, excludeArchived bool, excludeMembers bool) ([]Channel, error) {
 	values := url.Values{
 		"token": {api.config.token},
 	}
 	if excludeArchived {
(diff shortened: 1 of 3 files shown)


dvdplm:
This is breaking the API, so that's a bit troublesome. I think we need a general way of passing in options to API calls. The issue you address here is very real and I totally agree we should fix it, I'm just not sure adding arguments like this is the right way (or rather: I think we should add a way to add API options to all calls).
Perhaps something like:

```golang
opts := url.Values{"exclude_members": {"false"}}
api.GetChannels(opts)
```

We're having a similar issue with the `GetUsers` call, where there's currently no way of setting `presence=0`.

brentahughes:
@dvdplm I fully agree. I can make the changes needed to use an options struct. How do we want to handle values that are not used by the service? Should it be up to each service to validate the struct and return a friendly error if a value is set that isn't allowed for the service? Or should unused values be ignored? I worry that ignoring them could cause confusion for users if they set a value like pagination limit or offset and it doesn't do that because the service they are calling doesn't actually support it.

dvdplm:
My two cents: the Go type system doesn't really let us deal with this in a clean, machine-checked way. We'd have to add/remove/change arguments to methods, breaking the API at every turn. The alternative is to accept a grab-bag of `options` and have each method validate the parts they need and can use.
I agree that seemingly accepting params that are not actually used (e.g. pagination) is bad, but can be partially fixed with good API docs. I think it's fair to expect the user of a Slack library to read the Slack API docs.

:)

dvdplm:
@bah2830 any progress on this on your end?

brentahughes:
@dvdplm Sorry, I've switched companies so I no longer maintain a slack system for using this. If I get some free cycles this holiday season I will setup a simple system to test some changes and push them.

dvdplm:
@bah2830 sounds great! Do let us know if you can't make it happen and we'll see about getting it some other way. :)

james-lawrence:
@dvdplm SendMessage options are an example of how to do what you're describing in go its not really a problem with the type system =)

james-lawrence:
closing due to age.

PR #170, opened 2017-06-30
Print readable names for example 👀

mbbroberg:
Happy [Open Source Friday](https://opensourcefriday.com/)!!! ❤️❤️❤️

Thinking that a shift from `ID` to `Name` will make this first example friendlier to anyone exploring the Slack API at first. Also think that pointing to available methods would be handy for those hacking their way through Go.


nlopes:
Happy OSF! Thank you for sending this in!

PR #172, opened 2017-07-12
Add support for postMessage's reply_broadcast option

monder:
Add an optional parameter described here:
https://api.slack.com/methods/chat.postMessage#reply_broadcast

> Used in conjunction with thread_ts and indicates whether reply should be made visible to everyone in the channel or conversation. Defaults to false.


eklein:
👍 I would really appreciate having this, any reason this couldn't be merged? 😄

dvdplm:
lgtm

PR #173, opened 2017-07-24
Ignoring desktop_notification (fixes #125)

### LINKED ISSUE #125: no desktop_notification at eventMapping
no desktop_notification at eventMapping

jBugman:
There is no documentation on `desktop_notification` event (at least I could not find it) and it spams log with unmarshal errors.
I do not see any use for this event now, so I propose to ignore it instead of creating a new mapping.


dlsniper:
Thank you!

PR #176, opened 2017-08-03
chat: support chat.command for sending a command

kalbasit:
The `chat.command` method allows calling internal and custom commands in a slack team. It's worth to mention that I can't see `chat.command` in the [doc for methods](https://api.slack.com/methods#chat), and I found it being used in [wee-slack](https://github.com/wee-slack/wee-slack/blob/40c937f894d2968774dab41ddff02fd6a8c0dd5f/wee_slack.py#L3025).

Tested with https://github.com/erroneousboat/slack-term/pull/57

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -174,6 +174,7 @@ const (
 	chatUpdate      sendMode = "chat.update"
 	chatPostMessage sendMode = "chat.postMessage"
 	chatDelete      sendMode = "chat.delete"
+	chatCommand     sendMode = "chat.command"
 )
 
 type sendConfig struct {
@@ -228,6 +229,17 @@ func MsgOptionText(text string, escape bool) MsgOption {
 		if escape {
 			text = escapeMessage(text)
 		}
+		config.values.Add("text", text)
+		return nil
+	}
+}
+
+// MsgOptionCommand executes a command
+func MsgOptionCommand(command, text string) MsgOption {
+	return func(config *sendConfig) error {
+		config.mode = chatCommand
+		config.values.Del("ts")
+		config.values.Add("command", command)
 		config.values.Add("text", text)
 		return nil
 	}


dvdplm:
@kalbasit do you mind asking the Slack API support about this? Is it an undocumented feature they'd rather people didn't use or is it simply something they forgot to document? Also a usage example + test would be great so we know if/when it breaks. Thanks!

james-lawrence:
@kalbasit @dvdplm I'd be willing to accept this in the form of a method that lets you explicity set the mode.
```
// MsgOptionMode used to set a custom mode, useful for modes like 'chat.command' which exist but
// are not officially supported.
func MsgOptionMode(mode string) MsgOption {
	return func(config *sendConfig) error {
		config.mode = mode
		return nil
	}
}
```

dvdplm:
@james-lawrence can you elaborate on what the allowed values for `mode` are? Sounds like we'd want to lock down the allowed values.

bee-keeper:
Is there any interest in getting this committed?  Happy to take it on if you're willing to merge the feature?  I know it's an undocumented API but without we can't write bots which submit commands.

james-lawrence:
@dvdplm I say we lock down support for official values which we do via the MsgOptions, but there is a usecase here that I've also run into w/ ephemeral messages.

@bee-keeper yes there is interest on my part I have a related [PR](https://github.com/nlopes/slack/compare/master...james-lawrence:response-url-hacks#diff-1f9eebfe78e8930de1a3795966f172b3R338) which also requires some internal changes to SendMessage API.

In particular I think we should consider changing config.mode to config.endpoint and updating the relevant MsgOption. I also believe we should be very clear in the PR that the method is unsupported officially. Then this PR changes to:

```golang
// UnsafeMsgOptionEndpoint deliver the message to the specified endpoint.
// NOTE: USE AT YOUR OWN RISK: No issues relating to the use of this Option
// will be supported by the library, it is subject to change without notice that
// may result in compilation errors or runtime behaviour changes.
func MsgOptionEndpoint(endpoint string, update func(url.Values)) MsgOption {
	return func(config *sendConfig) error {
		config.endpoint = endpoint
		update(config.values)
		return nil
	}
}
```

PR #178, opened 2017-08-08
Update User mention funcPostMessageTo

sinramyeon:
Added PostMessageTo, because I couldn't find effective func for send a mention to specific user.
If there is anything to fix, please do not hesitate to reject my pull and tell me.
thx.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -102,6 +102,22 @@ func (api *Client) PostMessageContext(ctx context.Context, channel, text string,
 	return respChannel, respTimestamp, err
 }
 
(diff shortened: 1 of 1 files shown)


james-lawrence:
this doesn't look like a change that belongs in the library. mainly looks very specific to something you were trying to do, and it also just uses all the stuff provided by the library, anyone who wants a function like this can write it very quickly and easily.

but an example/test demonstrating how to do this would seem more acceptable.

james-lawrence:
going to close unless someone objects/wants to discuss further.

PR #179, opened 2017-08-09
SetLogger accepts generic interface, backwards-compatible with stdlib

jBugman:
`func SetLogger(l stdLogger) ` where `stdLogger` is a logger interface compatible with both stdlib and some 3rd party loggers such as logrus.

### DIFF
--- a/slack.go
+++ b/slack.go
@@ -8,7 +8,7 @@ import (
 	"os"
 )
 
-var logger *log.Logger // A logger that can be set by consumers
+var logger stdLogger // A logger that can be set by consumers
 /*
   Added as a var so that we can change this for testing purposes
 */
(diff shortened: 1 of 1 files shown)


bjhaid:
I am super interested in this, I have had to pass in an implementation io writer that called glog into a new stdlib logger to get by this. It will be better if I didn't have to do the hack

james-lawrence:
not sure how many of those methods we actually use. but +1.

slack.go
```diff
@@ -41,12 +41,29 @@ type Client struct {
 	debug bool
 }
 
+// stdLogger is a logger interface compatible with both stdlib and some
+// 3rd party loggers such as logrus.
+type stdLogger interface {
```
dvdplm, line 46: Don't we need https://golang.org/pkg/log/#Logger.Output here too?

PR #186, opened 2017-08-23
Return a special error when rate limited.

jmhobbs:
Instead of having to parse the error string provided, for the special case of status code 429, we can return a different error type to allow for intelligent retry.

See also: https://api.slack.com/docs/rate-limits

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -121,7 +121,7 @@ func postWithMultipartResponse(ctx context.Context, path, name, fieldname string
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode == 429 {
+	if resp.StatusCode == http.StatusTooManyRequests {
 		retry, err := strconv.ParseUint(resp.Header.Get("Retry-After"), 10, 64)
 		if err != nil {
 			retry = 1
@@ -153,7 +153,7 @@ func postForm(ctx context.Context, endpoint string, values url.Values, intf inte
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode == 429 {
+	if resp.StatusCode == http.StatusTooManyRequests {
 		retry, err := strconv.ParseUint(resp.Header.Get("Retry-After"), 10, 64)
 		if err != nil {
 			retry = 1
--- a/misc.go
+++ b/misc.go
@@ -130,7 +130,7 @@ func postWithMultipartResponse(ctx context.Context, path, name, fieldname string
 	}
 
 	// Slack seems to send an HTML body along with 5xx error codes. Don't parse it.
-	if resp.StatusCode != 200 {
+	if resp.StatusCode != http.StatusOK {
 		logResponse(resp, debug)
 		return fmt.Errorf("Slack server error: %s.", resp.Status)
 	}
@@ -162,7 +162,7 @@ func postForm(ctx context.Context, endpoint string, values url.Values, intf inte
 	}
 
 	// Slack seems to send an HTML body along with 5xx error codes. Don't parse it.
-	if resp.StatusCode != 200 {
+	if resp.StatusCode != http.StatusOK {
 		logResponse(resp, debug)
 		return fmt.Errorf("Slack server error: %s.", resp.Status)
 	}
--- a/misc.go
+++ b/misc.go
@@ -28,6 +28,9 @@ type HTTPRequester interface {
 
 var customHTTPClient HTTPRequester
 
+// Default duration for rate limiting if Retry-After header does not exist/parse
+const defaultRetrySeconds int64 = 60
+
 // HTTPClient sets a custom http.Client
 // deprecated: in favor of SetHTTPClient()
 var HTTPClient = &http.Client{}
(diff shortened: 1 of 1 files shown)


jmhobbs:
This can then be leveraged like so;

```go
history, err := client.GetChannelHistory(channel_id, slack.NewHistoryParameters())
if err != nil {
	rlerr, ok := err.(*slack.RateLimitedError)
	if ok {
		// Retry logic here instead
		log.Fatalf("Rate limited. Retry in %d seconds.", rlerr.RetryAfter)
	}
	log.Fatal(err)
}
```

dvdplm:
This is probably a dupe of https://github.com/nlopes/slack/pull/159. There's also https://github.com/nlopes/slack/tree/slack-error where work on this was begun. Unsure what the status is on either though (and to my eyes, this here looks fine too).

jmhobbs:
I've got this in production, and #159 looks good too. Should definitely be handled though, excellent library otherwise.

misc.go
```diff
@@ -42,6 +43,14 @@ func (s WebError) Error() string {
 	return string(s)
 }
 
+type RateLimitedError struct {
```
james-lawrence, line 46: lets add a method Backoff() time.Duration and change RetryAfter to time.Duration.

then people can do the following:
```
type backoff interface {
  Backoff() time.Duration
}

switch cause := err.(type) {
case backoff:
  time.Sleep(cause.Backoff())
}
```

james-lawrence (commented):
would like some other maintainers thoughts.

@catherinetcai @dvdplm

misc.go
```diff
@@ -112,6 +121,14 @@ func postWithMultipartResponse(ctx context.Context, path, name, fieldname string
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode == 429 {
```
catherinetcai: Let's use the `http.StatusTooManyRequests` in here https://godoc.org/net/http#pkg-constants

james-lawrence:
Normally I'd agree but we'd lose the duration that is in the response if
we use the standard library's error. We want to be able to pass that back
to the user.

catherinetcai:
@james-lawrence Was just referencing the status code constant itself, not the error that gets passed back

catherinetcai:
Oops, did not mean to close this when I commented. Fat fingered

misc.go
dvdplm: @james-lawrence I think @catherinetcai is referring to using naked status code numerics, not returning the standard library's error. I agree using `http.StatusTooManyRequests` is better here.

misc.go
```diff
@@ -136,6 +153,14 @@ func postForm(ctx context.Context, endpoint string, values url.Values, intf inte
...
```
dvdplm: `http.StatusTooManyRequests` here too

misc.go
jmhobbs: [`backoff.go` already has a `backoff` struct](https://github.com/nlopes/slack/blob/master/backoff.go#L15), should we consider a different name?

misc.go
```diff
@@ -112,6 +121,14 @@ func postWithMultipartResponse(ctx context.Context, path, name, fieldname string
...
+		retry, err := strconv.ParseUint(resp.Header.Get("Retry-After"), 10, 64)
+		if err != nil {
+			retry = 1
```
dvdplm: I doubt 1 sec is going to be enough. When Slack rate limits they set the `Retry-After` to 60 seconds so if we ever get here I suggest we return 60 (which is crazily aggressive, I know, but not sure what other value would assure waiting enough time?).

misc.go
dvdplm: I for one like the "RetryAfter" name. It's very hard to misunderstand and is consistent with the header name.

misc.go
james-lawrence: Ha that's what I get for just reading the email.

misc.go
```diff
@@ -27,6 +28,9 @@ type HTTPRequester interface {
 
 var customHTTPClient HTTPRequester
 
+// Default duration for rate limiting if Retry-After header does not exist/parse
+const defaultRetrySeconds int64 = 60
```
jmhobbs: @dvdplm added this as a constant up here, and changed it to 60.

jmhobbs: However, in my usage I mostly see 1 second delays, I wonder if instead of a default we should just raise a different error because we couldn't parse the retry-after header?

dvdplm: You know what, I think you're right. There is really no "good" default here and if Slack sends a 429 but no retry-after value we can parse, then perhaps it's best to return an error and let applications decide what they want to do (retry after 1, 10 or a 100 secs, call the police or whatever fits their setup).

james-lawrence: I tend to agree, I personally have a strong policy of not lying or surprising the users in API design.

we could do two things here:
1) return the parse error directly and let the application developers deal with it.
2) return a different implementation of the retry error with a default value specified, and the underlying error that caused it to occur.

either way works, I'm partial to just return the error as its less code, and if it ever comes up in practice we can deal with it later via method 2.

jmhobbs: I went with option 1, just return the error.

misc.go
```diff
@@ -136,8 +153,16 @@ func postForm(ctx context.Context, endpoint string, values url.Values, intf inte
...
+		retry, err := strconv.ParseInt(resp.Header.Get("Retry-After"), 10, 64)
+		if err != nil {
+			return err
+		}
+		return &RateLimitedError{time.Duration(retry) * time.Second}
+	}
+
 	// Slack seems to send an HTML body along with 5xx error codes. Don't parse it.
-	if resp.StatusCode != 200 {
+	if resp.StatusCode != http.StatusOK {
```
jmhobbs, line 165: I went ahead and changed these as well to be consistent.

misc.go
dvdplm: @james-lawrence is it ok by you if we go ahead and merge this as-is, i.e. can we PR the `Backoff()`/`RetryAfter()` in a separate PR?

james-lawrence: absolutely

jmhobbs:
Thanks all!

PR #192, opened 2017-09-26
Remove Text from PostMessageParameters

mavidser:
It is already a required argument in PostMessage(), and isn't used anywhere.
Also, initalize ThreadTimestamp with the default value


james-lawrence:
@mavidser do you have time to rebase this? and I'll get it merged.

Issue #196, opened 2017-10-12
GetUsers should use pagination


jschauma:
When calling GetUsers on a slack instance with a lot of users, you will receive a 500 error, as noted in https://api.slack.com/methods/users.list.  This call should probably use pagination to fill up the members array in https://github.com/nlopes/slack/blob/master/users.go#L183

PR #199, opened 2017-10-14
use pagination for users.list

jschauma:
Issue #196

Use pagination to try to get all users.
The API docs recommend to request 200 users each time, but it
there is also rate-limiting in effect, so if we loop too often
too quickly, we'll never get all users.  We try to work around
that by sleeping for 2 seconds in between each iteration (API
docs say "send no more than one message per second", so we
want to play it safe), but even then we get rate-limited when
trying to fetch several K users, so we try to get as many as
we can on each call (i.e. 999).

This is all very suboptimal, but getting, say, 8K users out of
20K is better than getting 0, so I suppose it's a start.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -145,7 +145,13 @@ func postForm(ctx context.Context, endpoint string, values url.Values, intf inte
 	if resp.StatusCode != 200 {
 		retryAfter := 0
 		if resp.StatusCode == 429 {
-			retryAfter, _ = strconv.Atoi(resp.Header.Get("retry-after"))
+			retryAfter, err = strconv.Atoi(resp.Header.Get("retry-after"))
+			if err != nil {
+				if debug {
+					logger.Printf("Got a 429 without a retry-after: %s", err)
+				}
+				retryAfter = 10
+			}
 		}
 		logResponse(resp, debug)
 		return WebError{
--- a/misc.go
+++ b/misc.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 )
@@ -36,10 +37,14 @@ type WebResponse struct {
 	Error *WebError `json:"error"`
 }
 
-type WebError string
+type WebError struct {
+	Status       int
+	RetryAfter   int
+	ErrorMessage string
+}
 
 func (s WebError) Error() string {
-	return string(s)
+	return string(s.ErrorMessage)
 }
 
 func fileUploadReq(ctx context.Context, path, fieldname, filename string, values url.Values, r io.Reader) (*http.Request, error) {
@@ -138,8 +143,16 @@ func postForm(ctx context.Context, endpoint string, values url.Values, intf inte
 
 	// Slack seems to send an HTML body along with 5xx error codes. Don't parse it.
 	if resp.StatusCode != 200 {
+		retryAfter := 0
+		if resp.StatusCode == 429 {
+			retryAfter, _ = strconv.Atoi(resp.Header.Get("retry-after"))
+		}
 		logResponse(resp, debug)
-		return fmt.Errorf("Slack server error: %s.", resp.Status)
+		return WebError{
+			Status:       resp.StatusCode,
+			RetryAfter:   retryAfter,
+			ErrorMessage: fmt.Sprintf("Slack server error: %s.", resp.Status),
+		}
 	}
 
 	return parseResponseBody(resp.Body, &intf, debug)
--- a/users.go
+++ b/users.go
@@ -199,7 +199,10 @@ pagination:
 			response, err = userRequest(ctx, "users.list", values, api.debug)
 			if err != nil {
 				if e, ok := err.(WebError); ok && e.Status == 429 {
-					time.Sleep(time.Duration(e.RetryAfter+1) * time.Second)
+					if api.debug {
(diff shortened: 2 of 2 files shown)


jschauma:
Note: PR updated with fixes from @pdbogen

misc.go
```diff
@@ -138,8 +143,16 @@ func postForm(ctx context.Context, endpoint string, values url.Values, intf inte
 
 	// Slack seems to send an HTML body along with 5xx error codes. Don't parse it.
 	if resp.StatusCode != 200 {
+		retryAfter := 0
+		if resp.StatusCode == 429 {
+			retryAfter, _ = strconv.Atoi(resp.Header.Get("retry-after"))
```
dvdplm: Discarding the error here is not so great as it leaves `retryAfter` set to `0`. Perhaps check for the error and set it to something like `60` by default to ensure that you don't end up in an eternal rate limiting loop?

users.go
```diff
@@ -169,22 +171,55 @@ func (api *Client) GetUserInfoContext(ctx context.Context, user string) (*User,
...
+// rate limits on `users.list`.
+func (api *Client) GetUsersContext(ctx context.Context) (users []User, err error) {
 	values := url.Values{
+		// API docs say to use 200, but rate limiting kicks in too
+		// quickly, so we'll have to be bad little citizens and
+		// use the max. :-/
+		"limit":    {"1000"},
+		"presence": {"true"},
 		"token":    {api.config.token},
-		"presence": {"1"},
```
dvdplm, line 181: ~~Removing this breaks the API so I think it should be made optional.~~ Sorry, seeing you replace it a few lines above.

users.go
```diff
@@ -169,22 +171,55 @@ func (api *Client) GetUserInfoContext(ctx context.Context, user string) (*User,
...
+	for {
+		var response *userResponseFull
+	retry:
+		for {
+			var err error
+			response, err = userRequest(ctx, "users.list", values, api.debug)
+			if err != nil {
+				if e, ok := err.(WebError); ok && e.Status == 429 {
+					if api.debug {
+						logger.Printf("GetUsersContext: users.list rate limited, sleeping for %d seconds", e.RetryAfter)
```
dvdplm: There's `api.Debugf()` and `api.Debugln()`.

misc.go
pdbogen: Ya, it's a good point. Arguably we should actually return an error in this case; if Slack responds with a 429 and doesn't set a retry-after error header, it's a malformed/unexpected response. That said, I think `10` is a reasonably sane failsafe, `60` is probably a bit too much. (Most of the rate limits are 1/second, but in this specific use case, it's 6/minute.)

dvdplm: You are right of course, 60sec is a pretty insane `RetryAfter` value, but it's what Slack defaulted to last year. That said, I'm perfectly ok with a 10sec default too. :)

kneal:
Where is this one at for getting merged?

users.go
```diff
@@ -101,9 +102,10 @@ type TeamIdentity struct {
 }
 
 type userResponseFull struct {
-	Members      []User                  `json:"members,omitempty"` // ListUsers
-	User         `json:"user,omitempty"` // GetUserInfo
-	UserPresence                         // GetUserPresence
+	Members          []User                  `json:"members,omitempty"` // ListUsers
+	User             `json:"user,omitempty"` // GetUserInfo
+	UserPresence                             // GetUserPresence
+	ResponseMetadata map[string]string       `json:"response_metadata,omitempty"`
```
james-lawrence, line 108: Generally I'm against using maps like this due to how they interact with the heap. use a type for the metadata with the explicit fields you want.

not to mention magic strings everywhere.

users.go
```diff
@@ -169,22 +171,51 @@ func (api *Client) GetUserInfoContext(ctx context.Context, user string) (*User,
...
 func (api *Client) GetUsers() ([]User, error) {
 	return api.GetUsersContext(context.Background())
 }
 
-// GetUsersContext returns the list of users (with their detailed information) with a custom context
-func (api *Client) GetUsersContext(ctx context.Context) ([]User, error) {
+// GetUsersContext returns the list of users (with their detailed information)
+// with a custom context. This call can take a long time for large teams, due to
+// rate limits on `users.list`.
+func (api *Client) GetUsersContext(ctx context.Context) (users []User, err error) {
```
james-lawrence, line 183: since You're adding a new method, return the pagination id and accept it as a parameter. don't want to restrict users of the library from being constrainted to storing all the users in memory.

james-lawrence (changes requested):
needs some changes, but overall lg.

kneal:
@jschauma Were you still working on this?

jschauma:
Neal <notifications@github.com> wrote:
> @jschauma Were you still working on this?

I'm sorry, I haven't had time to look into this.  (This is, of course,
directly related to the fact that the current patch works just fine for
me, so my itch has been scratched.)

james-lawrence:
if someone wants to clean this pr up, I'll be happy to fast track it.

james-lawrence:
closing in favour of #296, thanks @jschauma used your branch as the base.

jschauma:
Great, thanks a lot!

(Sorry to have left it hanging incomplete...)

james-lawrence:
Np! thanks for the initial pass. =)

PR #203, opened 2017-10-23
renamed channel arg to either channelId or channelName

### LINKED ISSUE #202: Consistency - channel argument is in some cases channel name, in others channel id
```
rtm := slack.New(token).NewRTM()
...
rtm.JoinChannel(channel string) // expects channel name
rtm.LeaveChannel(channel string) // expects channel id
rtm.GetChannelInfo(channel string) // expects channel id
```
would be good to make this consistent (change join channel to accept channel id instead of name) and also rename arguments so it is clear if it is channel name or id: `LeaveChannel(channelId string)`

pete911:
this closes #202 I have only renamed fields that will not break anyones code

### DIFF
--- a/channels.go
+++ b/channels.go
@@ -38,15 +38,17 @@ func channelRequest(ctx context.Context, path string, values url.Values, debug b
 }
 
 // ArchiveChannel archives the given channel
-func (api *Client) ArchiveChannel(channel string) error {
-	return api.ArchiveChannelContext(context.Background(), channel)
+// see https://api.slack.com/methods/channels.archive
+func (api *Client) ArchiveChannel(channelId string) error {
+	return api.ArchiveChannelContext(context.Background(), channelId)
 }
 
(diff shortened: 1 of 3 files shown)


dvdplm:
I think this is a good change, but it should be `channelID` not `channelId` (as per Go's formatting standard).

pete911:
@dvdplm updated (sorry didn't notice `ID` is used instead of `Id` everywhere else in the code)

mkungla:
@pete911 perhaps squash your commits

pete911:
@mkungla don't want to sound rude, but the last commit (and pull request) has been merged on 25th of July, this pull request has been here for 22 days ... So I feel like we should first address #207

mkungla:
@pete911 I do agree with that

james-lawrence:
Looks good, merging. Thanks @pete911 =)

PR #204, opened 2017-10-30
Allow for lowercase http proxy

gingersnapz:


dvdplm:
Hmm, I'm skeptical about this, not sure it's the Slack lib's responsibility to account for non-standard environment vars. What's the use-case exactly?

PR #208, opened 2017-11-14
Open websocket over HTTPS

lewellyn:
The websocket should be opened over HTTPS, per the Slack API documentation. The connection is currently upgraded from HTTP to HTTPS, which causes an extra connection and potential data leakage to a man in the middle. This change will prevent the automatic upgrade by doing the right thing instead.

### DIFF
--- a/websocket_managed_conn.go
+++ b/websocket_managed_conn.go
@@ -124,6 +124,7 @@ func (rtm *RTM) startRTMAndDial(useRTMStart bool) (*Info, *websocket.Conn, error
 		return nil, nil, err
 	}
 
+	// Only use HTTPS for connections to prevent MITM attacks on the connection.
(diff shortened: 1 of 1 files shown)


dvdplm:
@lewellyn I am very much inclined to merge this. Can you add a comment with link to the relevant Slack docs so our future selves know what happened here in case anything breaks?

dvdplm:
@catherinetcai do you mind taking a look at this one as well? Thanks!

catherinetcai:
@dvdplm LGTM. Agreed on having a comment to the docs

james-lawrence:
I'll just point out that its common practice to never use HTTP when HTTPS is available due to the security risks. I don't believe slack specfic docs are necessary.

Just going to add a comment inline on why https in general and merge it.

PR #212, opened 2017-11-21
Fix bug in users.SetUserPhoto

bdragon:
`SetUserPhoto` was calling itself recursively instead of passing through to `SetUserPhotoContext`.


dvdplm:
lgtm

PR #214, opened 2017-11-29
Extra options for menu callback (aka Options Load URL payload)

sazarkin:
Add extra fields to AttachmentActionCallback to decode Slack callback when message menus dynamically populated

https://api.slack.com/docs/interactive-message-field-guide#options_load_url


sazarkin:
Another options is add extra fields to `AttachmentActionCallback`

james-lawrence:
LGTM.

PR #218, opened 2017-12-06
Adding usergroup example and usergroups param options

kneal:
* Adding a usergroup example

* Need param options for endpoint.

### DIFF
--- a/usergroups.go
+++ b/usergroups.go
@@ -122,14 +122,25 @@ func (api *Client) EnableUserGroupContext(ctx context.Context, userGroup string)
 }
 
 // GetUserGroups returns a list of user groups for the team
-func (api *Client) GetUserGroups() ([]UserGroup, error) {
-	return api.GetUserGroupsContext(context.Background())
+func (api *Client) GetUserGroups(includeCount bool, includeDisabled bool, includeUsers bool) ([]UserGroup, error) {
+	return api.GetUserGroupsContext(context.Background(), includeCount, includeDisabled, includeUsers)
 }
 
 // GetUserGroupsContext returns a list of user groups for the team with a custom context
-func (api *Client) GetUserGroupsContext(ctx context.Context) ([]UserGroup, error) {
+func (api *Client) GetUserGroupsContext(ctx context.Context, includeCount bool, includeDisabled bool, includeUsers bool) ([]UserGroup, error) {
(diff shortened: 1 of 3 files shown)


usergroups.go
```diff
@@ -122,15 +122,26 @@ func (api *Client) EnableUserGroupContext(ctx context.Context, userGroup string)
...
 }
 
 // GetUserGroupsContext returns a list of user groups for the team with a custom context
-func (api *Client) GetUserGroupsContext(ctx context.Context) ([]UserGroup, error) {
+func (api *Client) GetUserGroupsContext(ctx context.Context, includeCount bool, includeDisabled bool, includeUsers bool) ([]UserGroup, error) {
 	values := url.Values{
 		"token": {api.config.token},
 	}
+	if includeCount {
+		values.Add("include_count", "true")
```
dvdplm, line 135: @nac2791 What are the API default values here? I.e. if these three params are omitted, are count/disabled/user list not included in the API response? Can you please document the defaults in the docs-comment?
I think we ideally should strive to use param builder methods to avoid hard to read method signatures, and imo `GetUserGroups(true, false, true)` is borderline. Do you mind looking into adding a param builder method? Thanks!

kneal: Do you have any examples of a param builder method I can look at? Not really sure what you mean. I was looking at some already merged code to model this off

kneal: Would it make more sense to do something like

```
// UserGroupParams struct to set params when passing to Slack API
type UserGroupListParams struct {
	includeCount bool
	includeDisabled bool
	includeUsers bool
}

// NewUserGroupParams provides an instance of UserGroupParams with all the sane default values set
func NewUserGroupListParams() UserGroupListParams {
	return UserGroupParams{
		includeCount: false,
		includeDisabled:  false,
		includeUsers: false
	}
}
```

Then pass the param struct to the method. It's still a breaking change but it looks like most other endpoints are using a similar structure.

kneal: I could create a conts to like 

```
const(
DEFAULT_INCLUDE_COUNT = false
...
)
```
Then reference the const vs pass false in the return

james-lawrence: the structure version would be fine, no need for consts.

james-lawrence:
closing in favor of #392

PR #221, opened 2017-12-12
adding more support for ephemeral messages

sebito91:
Thank you for opening this repo!

Adding extra support for the `chat.postEphemeral` workflow which allows us to build posts similar to existing `MsgOption...` workflow.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -44,12 +44,17 @@ type PostMessageParameters struct {
 	IconEmoji       string       `json:"icon_emoji"`
 	Markdown        bool         `json:"mrkdwn,omitempty"`
 	EscapeText      bool         `json:"escape_text"`
+
+	// chat.postEphemeral support
+	Channel string `json:"channel"`
+	User    string `json:"user"`
 }
 
 // NewPostMessageParameters provides an instance of PostMessageParameters with all the sane default values set
 func NewPostMessageParameters() PostMessageParameters {
 	return PostMessageParameters{
 		Username:    DEFAULT_MESSAGE_USERNAME,
+		User:        DEFAULT_MESSAGE_USERNAME,
 		AsUser:      DEFAULT_MESSAGE_ASUSER,
 		Parse:       DEFAULT_MESSAGE_PARSE,
 		LinkNames:   DEFAULT_MESSAGE_LINK_NAMES,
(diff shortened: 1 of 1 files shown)


chat.go
```diff
@@ -46,6 +46,17 @@ type PostMessageParameters struct {
...
 }
 
+// PostEphemeralParameters contains all necessary params for sending chat.postEphemeral
+type PostEphemeralParameters struct {
+	Channel     string       `json:"channel"`
+	Text        string       `json:"text"`
+	User        string       `json:"user"`
+	AsUser      bool         `json:"as_user"`
+	Attachments []Attachment `json:"attachments"`
+	LinkNames   int          `json:"link_names"`
```
andyone: `link_names` has a boolean type - https://api.slack.com/methods/chat.postEphemeral

andyone: Hmm... It looks like in the previous version of API this property was an integer. Don't mind about my comment.

VojtechVitek (approved):
LGTM

sebito91:
Any chance we can get a quick review + merge on this PR? Would be _great_ to have incorporated before year's end :D

xiam:
This is great @sebito91. Any chance this could be reviewed/merged soon?

Thanks!

chat.go
```diff
@@ -193,6 +247,15 @@ func MsgOptionPost() MsgOption {
 	}
 }
 
+// MsgOptionPostEphemeral posts an ephemeral message
+func MsgOptionPostEphemeral() MsgOption {
+	return func(config *sendConfig) error {
+		config.mode = chatPostEphemeral
+		config.values.Del("ts")
```
dvdplm, line 237: Why do we delete the `ts` here? Is it always set? Does it cause an error on the Slack side?

chat.go
```diff
@@ -272,6 +335,25 @@ func MsgOptionDisableMarkdown() MsgOption {
 	}
 }
 
+func MsgOptionPostEphemeralParameters(params PostEphemeralParameters) MsgOption {
```
dvdplm: Why do we need both this and `MsgOptionPostParameters`? Do ephemeral messages not support the options handled there?

chat.go
```diff
@@ -63,6 +74,17 @@ func NewPostMessageParameters() PostMessageParameters {
 	}
 }
 
+// NewPostEphemeralParameters provides an instance of PostEphemeralParameters with all the sane default values set
+func NewPostEphemeralParameters() PostEphemeralParameters {
+	return PostEphemeralParameters{
+		User:        DEFAULT_MESSAGE_USERNAME,
+		AsUser:      DEFAULT_MESSAGE_ASUSER,
+		Attachments: nil,
```
dvdplm: The API supports attachments in ephemeral messages right? Any chance we can get that in, and perhaps try to re-use the `postMessage` as much as possible (it also looks like the message text support the same attributes as text in a normal message)?

chat.go
sebito91: It doesn't cause any errors, but mimics the `MsgOptionPost`. The slack reply will give the timestamp on its own

chat.go
sebito91: Technically we don't, we could probably use the standard `MsgOptionsPostMessageParameters` function but in case this deviated in future it would be good to have its own space.

chat.go
sebito91: This is just the default, setting attachments to `nil` to start. You can (and we do) set attachments in these post types for sure, you just have to allocate that separately in your use of the struct

chat.go
dvdplm: I'd prefer we re-use code as much as possible (less code, less bugs etc) and the day the APIs deviate we'll refactor.

chat.go
dvdplm: I apologise, I wasn't clear. I was trying to say this: ephemeral messages seem to support much the same features as normal messages and so I think we should support them in the API as well. In other words: let's try to re-use as much code as possible from the normal message code paths. That way we'll get all the unfurling, attachment and text formatting options "for free". Do you think that is possible?

sebito91: Sure, can try this and get back to you.

sebito91:
@dvdplm mind taking a look now? Reusing the majority of existing `PostMessage` functionality now.

dvdplm:
@sebito91 looks lovely. LGTM.

PR #225, opened 2017-12-22
Changed Client.Debugf and Debugln to show from where they were called.

resc:
This change shows where the debug statements are called from, so that it's clearer which part of the code is logging what


dvdplm:
lgtm

PR #229, opened 2017-12-31
fix memory leak in managed connection

james-lawrence:
there is a go routine leak if disconnect is invoked on a managed connection that is not connected.
```
func printEvents(wg *sync.WaitGroup, id int, t *slack.RTM) {
	defer wg.Done()
	for msg := range t.IncomingEvents {
		log.Printf("id(%d): %+v", id, msg.Data)
		switch msg.Data.(type) {
		case *slack.DisconnectedEvent:
			return
		default:
		}
		time.Sleep(time.Second)
	}
}

func main() {
	wg := &sync.WaitGroup{}
	for i := 0; i < 10; i++ {
		wg.Add(1)
		c := slack.New("token")
		// c.SetDebug(true)
		rtm := c.NewRTMWithOptions(&slack.RTMOptions{})
		go rtm.ManageConnection()
		if err := rtm.Disconnect(); err != nil {
			log.Println("disconnect error", err)
		}
		log.Println("disconnect completed")
		go printEvents(wg, i, rtm)
	}
	wg.Wait()
}
```

### DIFF
--- a/rtm.go
+++ b/rtm.go
@@ -72,6 +72,7 @@ func (api *Client) NewRTMWithOptions(options *RTMOptions) *RTM {
 		isConnected:      false,
 		wasIntentional:   true,
 		killChannel:      make(chan bool),
+		disconnected:     make(chan struct{}),
 		forcePing:        make(chan bool),
 		rawEvents:        make(chan json.RawMessage),
 		idGen:            NewSafeID(1),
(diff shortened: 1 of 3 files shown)


james-lawrence:
@catherinetcai, @dvdplm mind take a look at this if you get a chance.

dvdplm:
LGTM and good catch. Out of curiosity, did you detect a leak in production and set out to find it or how did you get to this?

james-lawrence:
leak in production unfortunately. I noticed it a long time ago actually then forgot about it because we were deploying often enough to that subsystem it wasn't having any real impact.

dvdplm:
> …because we were deploying often enough…

Ha! Been there. :)

PR #233, opened 2017-12-31
nlopes/slack#232 - added missing Output method

mauleyzaola:
Fixes #232

PR #236, opened 2018-01-01
reduces required logger scope to only require an Output function.

james-lawrence:
as mentioned in in #233, reduces the required interface for logging to be satisfied just the Output method.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -1,0 +1,53 @@
+package slack
+
+import (
+	"fmt"
+	"sync"
+)
+
+// SetLogger let's library users supply a logger, so that api debugging
+// can be logged along with the application's debugging info.
+func SetLogger(l logProvider) {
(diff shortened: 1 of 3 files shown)


logger.go
```diff
@@ -0,0 +1,53 @@
...
+
+// logInternal represents the internal logging api we use.
+type logInternal interface {
+	Print(...interface{})
+	Printf(string, ...interface{})
+	Println(...interface{})
+	Output(int, string) error
+}
+
+// logger implements the additional methods used by our internal logging.
```
dvdplm: typo? Did you mean `ilogger`?

dvdplm:
The Travis failure looks spurious ("disk quota exceeded"). Looks good to me otherwise.

logger.go
james-lawrence: that I did.

PR #241, opened 2018-01-09
websocket_proxy: use http.ProxyFromEnvironment

zellyn:
http has a ProxyFromEnvironment that interprets HTTP_PROXY,
HTTPS_PROXY, and NO_PROXY. Seems less error-prone than doing it by
hand.


zellyn:
Hmmm. Just noticed #238, which would pre-empt this.

james-lawrence:
we're definitely going to switch to gorilla, hopefully soonish. going to close this out.

PR #243, opened 2018-01-12
Support slash commands

knqyf263:
I support `slash` commands.
https://api.slack.com/slash-commands

This PR mainly offer parse HTTP requests of `slash` commands
In addition, it also provides a convenient wrapper for use with http.HandlerFunc.

Please see an example.
Example 1: It's a very easy way to write a bot. You only need to write command handling.
Example 2: It's a flexible way. This example only use `parse` method.

I wrote the test code with reference to the following test code.
https://github.com/nlopes/slack/blob/master/stars_test.go

Please point out that if there are strange points!

### DIFF
--- a/examples/slash/slash.go
+++ b/examples/slash/slash.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"encoding/json"
-	"errors"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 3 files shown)


afjoseph:
It would be great to also have team ID validation.

james-lawrence:
looks fine to me, I'm generally against implementing the handler in the library though. specifically because of requests like @Obaied's, the http handler is going to be different based on the needs of each user of the library. all that being said, I'm good with this PR as is if you add documentation to the handler that it will not be extended beyond providing a base example for people to work with.

My prefered implementation would just have two functions, and the SlashCommand struct:
`SlashCommandParse(http.Request) (SlashCommand, error)`
`(t SlashCommand) ValidateToken(acceptableTokens ...string) (bool)`

knqyf263:
@james-lawrence Thank you for the review!
I understand. As you said, the needs of each user is different.
I fixed the function and example as you reviewed.
Please review again!

james-lawrence:
LGTM, thanks!

ghostsquad:
I just found this as I was doing some research on implementing slash command backend. I noticed that this function doesn't do the full `signing key` verification.

https://github.com/nlopes/slack/blob/master/security.go

Currently, I'm running into the problem of needing to run `ParseForm` and read the body (in order to do the verification, but it seems I can't do both?

PR #244, opened 2018-01-14
make NewRTM configurable

james-lawrence:
- adds dialer for managed RTM.
- deprecates the RTMWithOptions method.

### DIFF
--- a/rtm.go
+++ b/rtm.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/url"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


james-lawrence:
failure isn't understood, I suspect a timeout default causing it to hang.

zellyn:
Oh interesting. I missed this while creating #263. I think the switch to `RTMOption` is fine, but I'd argue that a `websocket.Dialer` is a better choice than a `net.Dialer` for changing how connections happen, because it has more websocket-specific stuff in it, in addition to a `net.Dialer` equivalent.

zellyn:
Of course, this leaks the gorilla `websocket.Dialer` type from the `nlopes/slack` package. I suppose once we move a few Go releases into the future, we could create an Alias.

james-lawrence:
@zellyn likely this PR won't be merged until I (or someone else!) gets a chance to look into the test failures.

I won't have time for the next few months probably due to other projects, work, and personal time. =/

zellyn:
@james-lawrence I'll incorporate these option-param changes into #263 and update it… any suggestions on other maintainers to ping?

james-lawrence:
I'll be able to review, just not investigate, but any one of the other maintainers can review as well.

zellyn:
@james-lawrence #269 is a combination of your PR and my old one, #263. The reason the tests were failing is that defaulting `UseStart` to `false` was hanging indefinitely. [Here's the fix](https://github.com/nlopes/slack/pull/269/files#diff-b10174736060c94b907283ae65ad8d80R24). I haven't figured out why yet, although I suspect it might be the test fake. ¯\\\_(ツ)\_/¯

Issue #240, opened 2018-01-09
Make bot respect Retry-After


pete0emerson:
This problem may be on me, but I'm not sure.

My RTM bot (based on your [example](https://github.com/nlopes/slack/blob/master/examples/websocket/websocket.go)) was offline this morning. After settign up `api.SetDebug(true)`, I saw that Slack has throttled me:

```slack-bot: 2018/01/09 09:09:26 slack.go:86: reconnection 5 failed: post: Slack server error: 429 Too Many Requests.
slack-bot: 2018/01/09 09:09:26 slack.go:92:  -> reconnecting in 406.399931ms
Event Received: Event Received: slack-bot: 2018/01/09 09:09:27 misc.go:164: HTTP/2.0 429 Too Many Requests
Content-Length: 34
Access-Control-Allow-Origin: *
Cache-Control: private, no-cache, no-store, must-revalidate
Content-Type: application/json; charset=utf-8
...
Retry-After: 43
...
{"ok":false,"error":"ratelimited"}
```

What appears to be happening, though, is that the bot continues to make requests, not respecting the Retry-After, and therefore gets continuously rate limited. In other words, I see the similar message above repeated over and over with no seeming attempt to throttle itself.

What is the right way to get the bot to respect the rate limiting, and is that best done in my code, or in your library?

Jmainguy:
check out https://github.com/nlopes/slack/issues/239, updating the code got my bot to start working again, even though respecting retry-after would be a cool feature to add.

james-lawrence:
#186 added basic infrastructure for this. updating where-ever it is missing it sohuld be pretty straight forward for anyone so inclined.

also let me know if updating to master doesn't resolve this.

james-lawrence:
try up

james-lawrence:
closing due to age and we have support in various areas now.

PR #246, opened 2018-01-14
Add an explicit event type for RTM rate-limit errors.

Helcaraxan:
Proposal for #240. The chosen implementation attempts to remain as much as possible in line with existing code-structure. It requires users to take action themselves on rate-limiting errors as the library remains transparent to them but it does provide them with an easier way to filter for such errors.

As the added comment in the code explains the identification of rate-limiting errors in the RTM API is brittle as there is no proper documentation of errors in that API. The values that are filtered against were retrieved experimentally by forcing a rate-limit on a private Slack workspace (screenshot below):

screen shot 2018-01-14 at 18 45 29

### DIFF
--- a/websocket_internals.go
+++ b/websocket_internals.go
@@ -63,6 +63,13 @@ func (m *MessageTooLongEvent) Error() string {
 	return fmt.Sprintf("Message too long (max %d characters)", m.MaxLength)
 }
 
+// RateLimitEvent is used when Slack warns that rate-limits are being hit.
+type RateLimitEvent struct{}
+
+func (e *RateLimitEvent) Error() string {
+	return "Messages are being sent too fast."
+}
+
 // OutgoingErrorEvent contains information in case there were errors sending messages
 type OutgoingErrorEvent struct {
 	Message  OutgoingMessage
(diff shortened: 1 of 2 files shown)


james-lawrence:
thanks for the PR.

may want to check out https://github.com/nlopes/slack/blob/master/misc.go#L119 for how to wire in ratelimit fixes.

Helcaraxan:
I had a look at that actually as you referenced it on #240. Given that the RTM API does not have a "retry-after" value in the error 😞 and was using other error type events in general (like the `MessageTooLongEvent`) I thought to define a new one.

It is definitely possible that I am misunderstanding what you are suggestion but from my perspective there is nothing really to wire-in. The rate-limit error will only be received with the `ack` of the messages that violates the limit and is not something immediately returned like the `http.StatusTooManyRequests` on Event API requests (and `rtm.SendMessage()` does not return an `error`). From the user-perspective the new `RateLimitEvent` will be observed and should be dealt with exactly like the similar `MessageTooLongEvent`.

james-lawrence:
yikes, what happens to the message when the RTM hits the rate limit? message is lost? nothing?

Helcaraxan:
This error, from what I can tell from my (obviously limited) experimentation, is a soft-limit. It tells you to slow down but still transmit the message. As far as my understanding goes you'll just get disconnected if you don't slow down. So no messages lost (again... as far as I understand).

james-lawrence:
LG, thanks Helcaraxan.

PR #249, opened 2018-01-18
Support some conversations API

suzuki-shunsuke:
* https://api.slack.com/methods/conversations.members
* https://api.slack.com/methods/conversations.archive
* https://api.slack.com/methods/conversations.unarchive
* https://api.slack.com/methods/conversations.setTopic
* https://api.slack.com/methods/conversations.setPurpose
* https://api.slack.com/methods/conversations.rename


dvdplm:
This looks great! :) One small thing: how about providing a default `Limit` value for `conversations.members`? The Slack docs suggest not fetching more than 200 at a time and I think we should respect that, perhaps even set our default lower to, say, 50?

james-lawrence:
looks good, had a I noticed it prior to merging in the http client changes I would have merged prior to that change. @suzuki-shunsuke mind making the minor adjustments necessary? if you don't I'll do it this weekend.

suzuki-shunsuke:
@dvdplm 

Thank you for reviewing.
According [the official documentation](https://api.slack.com/methods/conversations.members),
the parameter `limit` is optional and the default value is `100`,
so we don't need to set the default value.

suzuki-shunsuke:
@james-lawrence 
Thank you for reviewing.
I have merged master to resolve conflicts and added some modification.

james-lawrence:
@suzuki-shunsuke while I appreciate the removal, we can't quite remove those methods yet, as it would break compatability. there is a 2.0 issue mentioning its removal =(

suzuki-shunsuke:
@james-lawrence 
I'm sorry but I can't understand what you say.
The methods `getHTTPClient` and `SetHTTPClient` were removed at the previously merged commit https://github.com/nlopes/slack/commit/ba9c8fda035725042d054885cac8de247775a9ba#diff-f968539cea47dc5b6e9bd57eb21fba39.
The removal has already merged to master and has nothing to do with this PR.

james-lawrence:
look closer, the get might have been removed, but not the Set, it was just moved.

I'm now curious why they are showing up in this PR, then.

suzuki-shunsuke:
Oh, I see.
The method `SetHTTPClient` was not removed but moved to `slack.go`, wasn't it?

suzuki-shunsuke:
But please see https://github.com/nlopes/slack/pull/249/files .
I haven't removed anything.

james-lawrence:
haha oops, I looked at the merge commit =P sorry we're gtg.

PR #254, opened 2018-01-25
Support for reminders.add

phinze:
This is a straightforward implementation attempting to mirror patterns from
sibling handlers.

The only quirk is the fact that the endpoint actually supports "user" or
"channel" as optional parameters, even though "channel" is not in
Slack's API docs currently. I'm implementing this specifically for use
managing channel reminders so I added it in.

If I get the chance I'll circle back and round out the rest of the reminders API in follow up PRs.

### DIFF
--- a/reminders.go
+++ b/reminders.go
@@ -1,0 +1,45 @@
+package slack
+
+import (
+	"context"
+	"errors"
+	"net/url"
+)
+
+// AddReminder adds a reminder for a channel or a user.
+//
+// Only one of ChannelID or UserID should be provided. Both can be omitted, in
+// which case the reminder is set for the authenticated user.
+//
+// See https://api.slack.com/methods/reminders.add (NOTE: the ability to set
+// reminders on a channel is currently undocumented but has been tested to
+// work)
+func (api *Client) AddReminder(channelID, userID, text, time string) error {
+	return api.AddReminderContext(context.Background(), channelID, userID, text, time)
+}
+
+// AddReminderContext adds a reminder for a channel or a user with a custom context.
+//
+// See AddReminder for full details.
+func (api *Client) AddReminderContext(ctx context.Context, channelID, userID, text, time string) error {
+	values := url.Values{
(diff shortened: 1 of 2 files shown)


dvdplm:
lgtm

reminders.go
```diff
@@ -0,0 +1,45 @@
...
+
+// AddReminderContext adds a reminder for a channel or a user with a custom context.
+//
+// See AddReminder for full details.
+func (api *Client) AddReminderContext(ctx context.Context, channelID, userID, text, time string) error {
+	values := url.Values{
+		"token":   {api.token},
+		"text":    {text},
+		"time":    {time},
+		"channel": {channelID},
```
james-lawrence, line 29: I'd remove this from the map if you're setting it below.

reminders.go
```diff
@@ -0,0 +1,45 @@
...
+// reminders on a channel is currently undocumented but has been tested to
+// work)
+func (api *Client) AddReminder(channelID, userID, text, time string) error {
+	return api.AddReminderContext(context.Background(), channelID, userID, text, time)
+}
+
+// AddReminderContext adds a reminder for a channel or a user with a custom context.
+//
+// See AddReminder for full details.
+func (api *Client) AddReminderContext(ctx context.Context, channelID, userID, text, time string) error {
```
james-lawrence, line 24: can you have a channel and a user simultaneously? otherwise thoughts on just having a methods for each case?

dcu:
what's missing to merge this one?

james-lawrence:
I would like an answer to my question about the method API at a minimum.

dcu:
@phinze can you answer? this would be amazing

james-lawrence:
@dcu if you're that interested you can do the research and get back to us. at the very least it'll let us decide what route to go with the exposed API.

phinze:
Hi folks - @dcu my need for this went away which is why it's languished - I'd be happy to have you take over this PR if you have the spare cycles!

PR #257, opened 2018-02-03
add missing interactive/slash-command msg fields

lrstanley:
Only ones that I noticed that were missing so far.

-----

https://api.slack.com/interactive-messages
 * Responding with an error message: "replace_original"

https://api.slack.com/slash-commands
 * "In Channel" vs "Ephemeral" responses: "response_type"


dvdplm:
lgtm

PR #269, opened 2018-02-21
rtm: add RTM options, and gorilla websocket Dialer

zellyn:
This is a combination of #244 and #263, with a test fix of #244 to `UseRTMStart(true)` in the test.

This allows NewRTM to accept options: RTMOptionUseStart mimics the
existing RTMOptions.UseRTMStart, and RTMOptionDialer allows passing in
a gorilla websocket Dialer. The various Dialer fields are generally
useful; as an example, this will allow callers to specify
per-connection proxy settings without mutating environment variables.

### DIFF
--- a/rtm.go
+++ b/rtm.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net/url"
 	"time"
+
+	"github.com/gorilla/websocket"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


james-lawrence:
LGTM. thanks for digging into the test failure.

PR #272, opened 2018-02-22
Fixed "invalid_limit" error occurs.

### LINKED ISSUE #271: An "invalid_limit" error occurs.
If "Limit" is set to anything other than 0 for the "GetConversationRepliesParameters" used in the "GetConversationReplies" method, an "invalid_limit" error will occur and will not work.

uchimanajet7:
The part that was not correctly converted from a character string to a numerical value with the following code. We corrected this to be converted correctly.

```
if params.Limit != 0 {
	values.Add("limit", string(params.Limit))
}
```

fixes #271

### DIFF
--- a/conversation.go
+++ b/conversation.go
@@ -393,7 +393,7 @@ func (api *Client) GetConversationRepliesContext(ctx context.Context, params *Ge
 		values.Add("latest", params.Latest)
 	}
 	if params.Limit != 0 {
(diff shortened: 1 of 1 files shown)


james-lawrence:
wow, surprised that even compiled... TIL something new about go that never should have been allowed. thanks for the PR.

PR #275, opened 2018-03-06
Set DEFAULT_SEARCH_COUNT to match current API value

awh:


dvdplm:
While I agree that we should try to stay close to the API defaults I'm not sure I think it's worth breaking users' code that happen to rely on our current default of 100. Other than consistency, what is the reason for this change?

awh:
> what is the reason for this change?

The method that builds the query (`_search` in `search.go`) elides any search parameters that match the default:

```
 99     if params.Count != DEFAULT_SEARCH_COUNT {
100         values.Add("count", strconv.Itoa(params.Count))
101     }
```

As a consequence if you specify search parameters `slack.SearchParameters{Count:100}` it doesn't pass a `count=` value to the Slack API at all, and the documented default of 20 is assumed by them instead. Presumably this has been broken for anyone specifying a count of 100 since Slack changed the default!

Perhaps a better fix would be to just remove the parameter elision from `_search` entirely, and then you don't have to worry about keeping up with changes to the API defaults in future.

dvdplm:
I see what you mean now. It looks like the whole of `_search` is broken in the same way: if you pass in a value that is the same as the default then the parameter is skipped and not sent along to Slack at all. Anywhere our defaults differ from Slacks, we have a bug.
If we fix `_search` by adding all `params` to `values` then we send out a bigger payload than we need to, but on the other hand it's annoying to have to keep our defaults perfectly in sync with Slacks. I think I prefer sending more data. Something like:

```golang
	values := url.Values{
		"token": {api.token},
		"query": {query},
		"sort": {params.Sort},
		"sortDirection": {params.SortDirection},
		…
		"page": {strconv.Itoa(params.Page)}
	}

	response = &searchResponseFull{}
	err := post(ctx, api.httpclient, path, values, response, api.debug)
	…
```

What do you think?

james-lawrence:
I'm pretty indifferent on the path we take here, both are fine to me as they both will result in consistent behaviour. willing to accept this PR or a PR that always sends the values.

james-lawrence:
@awh I'll leave the decision to you for now, if you want this PR merged, or are willing to do the change to always send the parameters, just let us know.

PR #276, opened 2018-03-06
Parse attachments returned during message search

awh:


dvdplm:
Looks like this adds the `attachments` field, correct? Do you mind linking to the Slack docs? I looked over https://api.slack.com/methods/search.messages but could not see it there. Thanks!

awh:
It's not documented on that page, but then I don't think the given example is supposed to be exhaustive (it's captioned 'Typical success response'). The API is certainly returning it though, and the thing I need is inside that field, hence the PR :smile: 

The only thing I can point to is that https://api.slack.com/methods/search.messages says "The matching items are returned as hashes containing contextual messages." and https://api.slack.com/docs/message-attachments#attachment_structure says "Messages can have zero or more attachments, defined as an array."

dvdplm:
Fair enough, you've convinced me. The Slack docs can be frustrating at times, thank you for the links/additional details! :)

PR #264, opened 2018-02-17
Handle gorilla/websocket EOF close error.

jdanlewis:
Gorilla does not seem to throw an io.EOF. This PR should handle this error case and resolve #260.

Issue #260, opened 2018-02-07
Panic with gorilla websocket


parsley42:
I have a 'bot plugin that sends ~4k replies, which is the limit of what slack allows. With the latest nlopes/slack, the bot panics with the dump shown below. With my local fork reverting the gorilla websocket commit, I get my loooong messages as expected. If I get time this weekend I may try to dig deeper, but figured folks might want a heads-up. My code just uses the RTM.SendMessage(s.conn.NewTypingMessage(chanID)) for sending messages.

Here's the full panic / stack trace.
panic: repeated read on failed websocket connection

goroutine 36 [running]:
github.com/uva-its/gopherbot/vendor/github.com/gorilla/websocket.(*Conn).NextReader(0xc042044f00, 0xc042430358, 0x0, 0x0, 0xc04234fec0, 0xc04234fec0)
        C:/Users/dlp7y/go/src/github.com/uva-its/gopherbot/vendor/github.com/gorilla/websocket/conn.go:958 +0x3e0
github.com/uva-its/gopherbot/vendor/github.com/gorilla/websocket.(*Conn).ReadJSON(0xc042044f00, 0x8b65c0, 0xc0422b15a0, 0xba51c0, 0xbd6d30)
        C:/Users/dlp7y/go/src/github.com/uva-its/gopherbot/vendor/github.com/gorilla/websocket/json.go:45 +0x36
github.com/uva-its/gopherbot/vendor/github.com/nlopes/slack.(*RTM).receiveIncomingEvent(0xc042414640)
        C:/Users/dlp7y/go/src/github.com/uva-its/gopherbot/vendor/github.com/nlopes/slack/websocket_managed_conn.go:288 +0xb1
github.com/uva-its/gopherbot/vendor/github.com/nlopes/slack.(*RTM).handleIncomingEvents(0xc042414640, 0xc04236d260)
        C:/Users/dlp7y/go/src/github.com/uva-its/gopherbot/vendor/github.com/nlopes/slack/websocket_managed_conn.go:223 +0x34
created by github.com/uva-its/gopherbot/vendor/github.com/nlopes/slack.(*RTM).ManageConnection
        C:/Users/dlp7y/go/src/github.com/uva-its/gopherbot/vendor/github.com/nlopes/slack/websocket_managed_conn.go:54 +0x26c

dvdplm:
This seems very worrying, thank you for reporting it.

[...]

PR #278, opened 2018-03-14
Update error handling for gorilla/websocket ReadJSON()

parsley42:
Even after the previous fixes for #260 and #264, I had problems with a robot panic'ing in gorilla/websocket after a few days. I did some digging around in the gorilla/websocket code. The 'ReadJSON' function called from receiveIncomingEvent() returns one of three possible error values:
 1. io.ErrUnexpectedEOF (but never io.EOF) from json ... Decode()
 2. The error value from NextReader(...)
 3. Other error values from json ... Decode()

From the comments for NextReader():
```go
// Applications must break out of the application's read loop when this method
// returns a non-nil error value. Errors returned from this method are
// permanent. Once this method returns a non-nil error, all subsequent calls to
// this method return the same error.
```

For the first case, I'm forcing a ping. For the second and third case, it wasn't clear to me that the call to websocket.IsCloseError(...) catches all possible error returns from NextReader(), and I don't know a straightforward means of differentiating all json Decode() error values from all NextReader() error values. I would guess that non-`io.ErrUnexpectedEOF` error values from json Decode() are unlikely, and more likely I was getting error values returned from NextReader() that weren't being handled. Given the comments on NextReader(), I felt the safest way to handle other errors was to kill the connection to force reconnect.

I'm currently running with this patch for [Gopherbot](https://github.com/lnxjedi/gopherbot), but would appreciate others having a look and testing. @jdanlewis - what are your thoughts?

### DIFF
--- a/websocket_managed_conn.go
+++ b/websocket_managed_conn.go
@@ -291,12 +291,16 @@ func (rtm *RTM) receiveIncomingEvent() {
 	event := json.RawMessage{}
 	err := rtm.conn.ReadJSON(&event)
 	switch {
-	// ReadJSON calls NextReader which states:
-	// Applications must break out of the application's read loop when this method
-	// returns a non-nil error value. Errors returned from this method are
-	// permanent. Once this method returns a non-nil error, all subsequent calls to
-	// this method return the same error.
+	case err == io.ErrUnexpectedEOF:
+		// EOF's don't seem to signify a failed connection so instead we ignore
+		// them here and detect a failed connection upon attempting to send a
+		// 'PING' message
+
+		// trigger a 'PING' to detect potential websocket disconnect
+		rtm.forcePing <- true
 	case err != nil:
+		// All other errors from ReadJSON come from NextReader, and should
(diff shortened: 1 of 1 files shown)


jdanlewis:
@parsley42 I ran a couple of short tests and this seems to be fine, but I'm not able to force an error besides the websocket.CloseError. 

If you want to distinguish between errors from JSON Decode and NextReader, we could decode the JSON ourselves. Then, we could force reconnect anytime NextReader returns an error.

james-lawrence:
LGTM. sorry for the delay.

ArcticSnowman:
@james-lawrence  When can we expect a release with this in?

james-lawrence:
0.4.0 should have it?

ArcticSnowman:
Ah that's not been made a release...  so the latest is still 0.2.0.... afaik
Managed to get govendor to update the package..

james-lawrence:
ah, I consider tags as the release. sorry will make a note to ensure github has the releases as well.

Issue #94, opened 2016-08-16
Support Events API


dougnukem:
Slack announced a new Events API which provides similiar functionality to the Real-time message API (RTM) but with a simpler webhook interface.
- https://api.slack.com/events-api
- https://api.slack.com/faq#events_api

james-lawrence:
will review/accept PRs for this.

high level requirements:
- don't implement the handler, that is application dependent.
- do implement the data structures and utility functions to reduce the http handler to a few lines.

james-lawrence:
closing as we have basic support.

PR #283, opened 2018-03-21
Add Events API helper middleware

JackWink:
Hey there! We're currently using the slack library in our project, but had a need to use the Events API as well. I'm trying to contribute back some middleware we developed to get the Event API up and running.

I saw this was requested in #94, and unfortunately I don't have time to write parsers for all the event structures right now, I'm hoping that this middleware is general enough to provide as a starting point for future work.

Commit Message:
In order to use slack's Events API, your webhook must respond to a challenge request from Slack. This commit adds middleware that will look for Slack's challenge request and responds. Any other valid JSON request is forwarded to the user's handler for parsing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,11 @@ func main() {
 
 See https://github.com/nlopes/slack/blob/master/examples/websocket/websocket.go
 
+## Events API
+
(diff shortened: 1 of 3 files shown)


JackWink:
doesn't seem like there is intent to ever pull this in

PR #291, opened 2018-04-02
improvements to ephemeral messaging

james-lawrence:
puts it inline with the SendMessage API.

also starting to lay the ground work for the send message api to support response URLs from things like slash commands.


dvdplm:
lgtm

PR #293, opened 2018-04-05
Events API support

ACollectionOfAtoms:
Here's another attempt at https://github.com/nlopes/slack/issues/94 (I started on this work before I realized https://github.com/nlopes/slack/pull/283 was open /shrug)

I'll be honest I'm still learning Go! But I needed the Events API capability for a project and so thought I would share my work. 

Would GREATLY appreciate feedback (because you know, still learning Go). 

I tried to keep the API as similar as possible to parsing RTM events. 

@james-lawrence when you said in #94 :
> - don't implement the handler, that is application dependent.
> - do implement the data structures and utility functions to reduce the http handler to a few lines.

I'm pretty sure I got the "don't implement the handler" part but not so sure about the second bullet point.

Anyway there is an [EXAMPLE](https://github.com/nlopes/slack/compare/master...ACollectionOfAtoms:events-api?expand=1#diff-70197fda1fff6e1e38fedb4fb8f269faR15)

I'm very open to refactoring as needed (as it's a Go-learning experience for me) and of course if this PR isn't up to snuff that's cool with me too; still would appreciate feedback.


james-lawrence:
will review next weekend. sorry for the delay.

pa3ng:
+1

kneal:
+1

james-lawrence:
sorry guys, been busy havent had time to think about this one. =(

events.go
```diff
@@ -0,0 +1,182 @@
...
+		e.Type,
+		e,
+		EventsAPIInnerEvent{iE.Type, recvEvent},
+	}, nil
+}
+
+// ParseEventsAPIEvent parses the outter and inner events (if applicable) of an events
+// api event returning a EventsAPIEvent type. If the event is a url_verification event,
+// the inner event is empty.
+func (api *Client) ParseEventsAPIEvent(rawEvent json.RawMessage) (EventsAPIEvent, error) {
```
james-lawrence: this doesn't need to be attached to the client. i'd just remove the debug as it'll return an error in that case.

I'd also make a custom error type for the events API to have the metadata in a structured format and for the `Error() string` method just return whatever slack claims the error is unaltered.

events_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+	"testing"
+)
+
+var c = New("my-token")
+
+func TestParserOuterCallBackEvent(t *testing.T) {
+	eventsAPIRawCallbackEvent := `
+			{
+				"token": "XXYYZZ",
```
james-lawrence: this is the verification token correct? how do we validate these events against the token?

james-lawrence (commented):
Overall looks good.

because this is the first pass at an events API.
I'm thinking of punting into its own package if possible with a weaker promise of API stability. thoughts on this?

would give us more time over here to see how people react to the API without having to be as concerned about maintaining compatibility. but at the same time let people use whats available.

thinking to put it in a `slackevents` package will also have the side benefit of reading better.

events_test.go
JackWink: I believe these are just shared secrets. Curious if you'd be open to allowing http middleware to do the validation of these and the webhook URL validation. The approach I took in #283 could be complimentary to this PR and do the validation? If you'd rather leave it entirely to the user then I think it's just a string comparison** to verify.

** [Time Independent/Secure comparison](https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare)

ACollectionOfAtoms: Hm, might be missing that part @james-lawrence and also thanks @JackWink for the pointers. 

Will re-read some of the Slack docs to ensure correct usage!

ACollectionOfAtoms:
@james-lawrence I think a separate package makes a lot of sense. Also I'm all for mentioning the Events API is a continued WIP in the readme or something like that. 

Will address PR comments along with creating a `slackevents` package ASAP. 

Oh and I wouldn't mind addressing any future Events API related issues if/when this gets merged.

events_test.go
james-lawrence: doing middleware means parsing the json mutliple times, I was mainly pointing out that a function should be included to do a basic comparison so people don't have to write the 4-6 lines themselves =)

examples/events/events.go
```diff
@@ -0,0 +1,45 @@
...
+			}
+			w.Header().Set("Content-Type", "text")
+			w.Write([]byte(r.Challenge))
+		}
+		if eventsAPIEvent.Type == slack.CallbackEvent {
+			postParams := slack.PostMessageParameters{}
+			innerEvent := eventsAPIEvent.InnerEvent
+			switch ev := innerEvent.Data.(type) {
+			case *slack.AppMentionEvent:
+				api.PostMessage(ev.Channel, "Yes, hello.", postParams)
```
JackWink: Two notes on this example:

1. You need to write back an OK, otherwise slack will redeliver the event
2. Sending a reply should probably be done after writing OK in case slack's API times out. Slack will redeliver if you don't respond within 3 seconds. https://api.slack.com/events-api#graceful_retries

events_test.go
JackWink: fair enough, I'd prefer that `ParseEventsAPIEvent` take in the verification token, or allow you to explicitly pass an option not to verify it.  Something like:

```go
type EventApiConfig struct {
    VerificationToken string
    DoNotVerify bool
}

type EventApiOption func(cfg *EventApiConfig)

func UnsafeVerification() EventApiOption {
  return func(cfg *EventApiConfig) {
    cfg.DoNotVerify = true
  }
}

func WithVerificationToken(token string) EventApiOption {
  return func(cfg *EventApiConfig) {
    cfg.VerificationToken = token
  }
}
func ParseEventsApiEvent(msg json.RawMessage, opts ...EventApiOption) {
  cfg := EventApiConfig{}
  for _, opt := opts {
    opt(&cfg)
  }
  ...
  if cfg.DoNotVerify {
    return data, nil
  }

  if cfg.VerificationToken == "" {
    return nil, fmt.Errorf("Missing event verification token")
  }
  
  if subtle.ConstantTimeCompare([]byte(cfg.VerificationToken), []byte(data.Token)) != 1 {
    return nil, fmt.Errorf("Event verification token doesn't match")
  }
  return data, nil
}

Comment (james-lawrence) [reply events_test.go]: lets pass on that part atm. I like the idea, except there are cases where we might need to use multiple verification tokens (if you're running multiple bots for example)

Comment (JackWink) [reply events_test.go]: I think that more advance use-case would fall under the DoNotVerify option, you could still provide the validation function separate from the parser. I feel like if you separate the validation and parsing you're more likely to have users skip validation entirely. IMO secure-by-default with the option to shoot yourself in the foot is better.

Comment (james-lawrence) [reply events_test.go]: I mean I'd just make the option take a function that accepts the request's token and returns true/false.

```golang
type verifier interface {
  Verify(t string) bool
}

func OptionVerifyToken(v verifier) Option {
  // ...
}
```

mainly want to avoid bogging down this PR with these details atm.

JackWink: That interface would work for me

sjqnn:
+1

james-lawrence:
@ACollectionOfAtoms sounds good. i'll fast track it once its in the slackevents package.

slackevents/parsers.go
```diff
@@ -0,0 +1,194 @@
...
+	return EventsAPIEvent{
+		e.Token,
+		e.Type,
+		e,
+		EventsAPIInnerEvent{iE.Type, recvEvent},
+	}, nil
+}
+
+type Config struct {
+	VerificationToken string
```
james-lawrence, line 121: I think the config should just contain the verifier type atm. but can be discussed later.

slackevents/parsers.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func (c TokenComparator) Verify(t string) bool {
+	return subtle.ConstantTimeCompare([]byte(c.VerificationToken), []byte(t)) == 1
+}
+
+// ParseEventsAPIEvent parses the outter and inner events (if applicable) of an events
+// api event returning a EventsAPIEvent type. If the event is a url_verification event,
+// the inner event is empty.
+func ParseEventsAPIEvent(rawEvent json.RawMessage, opts ...Option) (EventsAPIEvent, error) {
```
james-lawrence: this can be renamed to ParseEvent.

james-lawrence (approved):

ACollectionOfAtoms:
The API ends up looking like:
```golang
eventsAPIEvent, e := slackevents.ParseEvent(json.RawMessage(body), slackevents.OptionVerifyToken(&slackevents.TokenComparator{"TOKEN"}))
```

This last batch of commits adds tons more test and events (I think I've added all of them except for the ones in "Developer preview" mode.

james-lawrence:
once the conflicts are fixed i'll merge.

ACollectionOfAtoms:
@james-lawrence Will you update the README after you merge and mention the whole "still in early development" thing? I've added a lot more test but I still haven't had the chance to update my own Slack bot to do some manual QA

ACollectionOfAtoms:
I can also update the readme of course! (thought you might want that opportunity though)

james-lawrence:
Feel free to update the readme. also if you could rebase off the latest master that'd be great.

ACollectionOfAtoms:
@james-lawrence done and done

ACollectionOfAtoms:
BTW @pa3ng @nac2791 @sjqnn and anyone else who was waiting for this PR to be merged, if you run into any issues with the Events API please feel free to open an issue and ping me in it!

PR #298, opened 2018-04-17
Support users.profile.get API

suzuki-shunsuke:
https://api.slack.com/methods/users.profile.get

### DIFF
--- a/users.go
+++ b/users.go
@@ -393,3 +393,30 @@ func (api *Client) UnsetUserCustomStatus() error {
 func (api *Client) UnsetUserCustomStatusContext(ctx context.Context) error {
 	return api.SetUserCustomStatusContext(ctx, "", "")
 }
+
+// GetUserProfile retrieves a user's profile information.
+func (api *Client) GetUserProfile(userID string, includeLabels bool) (*UserProfile, error) {
+	return api.GetUserProfileContext(context.Background(), userID, includeLabels)
+}
+
+type getUserProfileResponse struct {
+	SlackResponse
+	Profile *UserProfile `json:"profile"`
+}
+
+// GetUserProfileContext retrieves a user's profile information with a context.
+func (api *Client) GetUserProfileContext(ctx context.Context, userID string, includeLabels bool) (*UserProfile, error) {
+	values := url.Values{
+		"token": {api.token}, "user": {userID},
+		"include_labels": {strconv.FormatBool(includeLabels)}}
+	resp := &getUserProfileResponse{}
+
+	err := post(ctx, api.httpclient, "users.profile.get", values, &resp, api.debug)
+	if err != nil {
+		return nil, err
+	}
+	if !resp.Ok {
+		return nil, errors.New(resp.Error)
+	}
+	return resp.Profile, nil
+}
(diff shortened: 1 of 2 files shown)


james-lawrence:
as always, thanks @suzuki-shunsuke

suzuki-shunsuke:
@james-lawrence 
You're welcome.
And thank you for merging.
I'm glad to contribute the awesome OSS project.

users.go
```diff
@@ -16,27 +16,35 @@ const (
...
+	Image72               string                            `json:"image_72"`
+	Image192              string                            `json:"image_192"`
+	ImageOriginal         string                            `json:"image_original"`
+	Title                 string                            `json:"title"`
+	BotID                 string                            `json:"bot_id,omitempty"`
+	ApiAppID              string                            `json:"api_app_id,omitempty"`
+	StatusText            string                            `json:"status_text,omitempty"`
+	StatusEmoji           string                            `json:"status_emoji,omitempty"`
+	Team                  string                            `json:"team"`
+	Fields                map[string]UserProfileCustomField `json:"fields"`
```
nolanlum, line 40: There's a small issue with this `Fields` field: Slack is inconsistent with the type of this field in the profile object.

If there are custom fields present, then `fields` will be a JSON object, and all is well. If there are NO custom fields present, then `fields` is an empty array (on one particularly old Slack team, `fields` was actually `null`?!). This causes an `unmarshalling_error` for an otherwise valid `profile` object.

suzuki-shunsuke: Thank you for your comment and I'm sorry for this issue.
I agree.
I send the Pull Request to revert the change.

https://github.com/nlopes/slack/pull/304

We have to think how to support custom fields.

nolanlum: The cleanest way would probably(?) be to have a custom unmarshaller for this struct that calls through to `json.Unmarshal` for everything but `Fields` (we'll have to set the struct field tag to `json:"-"`), and then check the datatype of the `fields` in the JSON response before unmarshalling that as well. The downside of this approach is we would have to write a custom marshaller as well.

An alternative approach is to compose some `UserProfileWithCustomFields` struct with an anonymous `UserProfile` field, but that would get messy, since you would have to deal with two types for essentially the same JSON blob, depending on whether or not you know there are custom fields in your Slack workspace (and `GetUserProfile` functions would have to return values accordingly).

A third approach (though I'm unsure if this would actually work) would be to make the type of `Fields` `interface{}` so `json.Unmarshal` could handle it, and then have client library code deal with making sense of it.

As I actually don't have much experience developing Golang libraries that wrap JSON APIs, I'd like to get the opinion of a project maintainer (@james-lawrence ?) as well.

It would also be nice to inform Slack of this small bug in their API, I suppose :)

suzuki-shunsuke: This is one of solutions.

https://play.golang.org/p/sKslV43mXEG

james-lawrence: @suzuki-shunsuke your example code looks good, minus the pointer for the field. you don't need it to be a pointer the json library handles it correctly.

suzuki-shunsuke: @james-lawrence 
Thank you for your review.
I will send a pull request within tomorrow.

suzuki-shunsuke: I send a Pull Request.

https://github.com/nlopes/slack/pull/305

Issue #302, opened 2018-04-28
race condition when disconnecting from RTM


milosgajdos:
There is a race condition in the `rtm` part of the `slack` package.

This is a simple program to reproduce it (**please see below how to build it to reproduce the race**):

```Go
package main

import (
	"fmt"
	"log"
	"os"
	"os/signal"
	"syscall"
	"time"

	"github.com/nlopes/slack"
)

func main() {
	// get the Slack API key
	apiKey := os.Getenv("SLACK_API_KEY")
	// create Slack API client
	api := slack.New(apiKey)

	// create RTM API connection
	rtm := api.NewRTM()
	go rtm.ManageConnection()

	// watch slack messages
	go watchMessages(rtm)

	// Signal handler to stop the bot when termination signal is received
	sigChan := make(chan os.Signal, 1)
	// register signal handler
	signal.Notify(sigChan, os.Interrupt, os.Kill, syscall.SIGTERM)

	timeout := 30 * time.Second
	ticker := time.NewTicker(timeout)

	select {
	case <-ticker.C:
		fmt.Printf("Time out")
	case sig := <-sigChan:
		fmt.Printf("Got signal: %s: Shutting down\n", sig)
	}

	if err := rtm.Disconnect(); err != nil {
		fmt.Printf("Error: %s\n", err)
		os.Exit(1)
	}
}

// watchMessages watches Slack messages and prints them with stdout
func watchMessages(rtm *slack.RTM) {
	// monitor all slack messages
	for msg := range rtm.IncomingEvents {
		switch ev := msg.Data.(type) {
		case *slack.MessageEvent:
			fmt.Printf("Message: %s\n", ev.Text)

		case *slack.LatencyReport:
			log.Printf("Current Slack RTM latency: %v", ev.Value)

		case *slack.RTMError:
			fmt.Printf("Error: %s\n", ev.Error())

		case *slack.InvalidAuthEvent:
			fmt.Printf("Invalid credentials")
			return

		default:
		}
	}
}
```

You must build the above program with `-race` flag:

```console
$ go build -race
```

The result:

```console
$ ./slackrace
^CGot signal: interrupt: Shutting down
==================
WARNING: DATA RACE
Read at 0x00c4200e6551 by main goroutine:
  main.main()
      /Users/milosgajdos/go/src/slackrace/vendor/github.com/nlopes/slack/websocket.go:71 +0x3be

Previous write at 0x00c4200e6551 by goroutine 7:
[...]

PR #303, opened 2018-04-28
Fixes race condition encountered when diconnecting from RTM API

milosgajdos:
The race condition is described in [#302](https://github.com/nlopes/slack/issues/302) including a small program that allows to reproduce it.

Closes nlopes/slack/#302


james-lawrence:
LG thanks.

PR #301, opened 2018-04-22
Fixed PostMessage error bug.  fixed #294

asiyani:
Added condition to check `SlackResponse` for any Error.
[info](https://api.slack.com/web)
```
{
    "ok": true,
    "stuff": "This is good"
}
```
```
{
    "ok": false,
    "error": "something_bad"
}
```
if `SlackResponse.Ok` is false `SendMessageContext` will return an Error.

If any issue please let me know.

Thank you.

PR #307, opened 2018-05-06
cleanup some error handling

james-lawrence:
SlackResponse and WebResponse were basically identical, this standardizes on WebResponse and adds a utility error method to WebResponse to clean up error handling throughout the library.

this is the base work I requested for #301, and includes the fix from that PR.

### DIFF
--- a/bots.go
+++ b/bots.go
@@ -16,7 +16,7 @@ type Bot struct {
 
 type botResponseFull struct {
 	Bot `json:"bot,omitempty"` // GetBotInfo
-	SlackResponse
+	WebResponse
 }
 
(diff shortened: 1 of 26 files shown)


dvdplm:
Not loving the `WebResponse` name tbh: "web" is such a generic term and implies web pages are involved. Both are public types but I wonder if that is really necessary. I do like the utility error method and cleanup. Are there were methods returning `SlackResponse` or `WebResponse` directly? I am pretty sure this will break some of Chatlios code (/cc @jeberly ) for instance.

james-lawrence:
no methods return them directly unfortunately both were exposed on bunch of types though. I didn't like either name either. =) happy to switch it though.

PR #308, opened 2018-05-07
Improve usage of interface{} in the HTTP helper functions

theckman:
The HTTP helper functions (post and postForm) take an `intf interface{}`
parameter, with the goal of using it to unmarshal the JSON response. This
requires that those calling the helper functions pass in a pointer to the type
they want to unmarshal the JSON in to.

The `parseResponseBody` function was a bit different in that it took a
`*interface{}` instead. Looking at its behaviors, this looks to have been
accidental when the unmarshaling behavior was implemented early on (1b96f659ad).
`json.Unmarshal` needs the `interface{}` value itself to contain a pointer, it
doesn't need to (and probably shouldn't) be a pointer.

To go in to deeper detail, the `interface{}` itself encapsulates the original
value and its type. When the JSON Unmarshaler tries to use the value within the
interface, it needs it to be a pointer so that it can manipulate the data and
set the values. The interface being a pointer only allows you to change what
data it encapsulates, and that's not generally how you'd use `interface{}` with
JSON.

This change updates the `*interface{}` typed argument of `parseResponseBody` to
be `interface{}` instead, and updates all of the call sites to use the new API.

Signed-off-by: Tim Heckman <t@heckman.io>

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -66,7 +66,7 @@ func fileUploadReq(ctx context.Context, path, fieldname, filename string, values
 	return req, nil
 }
 
-func parseResponseBody(body io.ReadCloser, intf *interface{}, debug bool) error {
+func parseResponseBody(body io.ReadCloser, intf interface{}, debug bool) error {
 	response, err := ioutil.ReadAll(body)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


dvdplm:
lgtm

james-lawrence:
@theckman you're absolutely correct, I noticed this as well but never bothered to deal with it =)

PR #309, opened 2018-05-08
Read user response metadata from correct JSON key

tbroadley:
I noticed that `GetUsers` does not retrieve all results when there are more than 200 users in an organization. I believe this is happening because the API calls in `users.go` are looking for response metadata under the "metadata" key of the response, whereas the metadata are actually found under the "response_metadata" key.

### DIFF
--- a/users.go
+++ b/users.go
@@ -173,7 +173,7 @@ type userResponseFull struct {
 	User         `json:"user,omitempty"` // GetUserInfo
 	UserPresence                         // GetUserPresence
 	WebResponse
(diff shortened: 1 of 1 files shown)


james-lawrence:
huh, what about the user list pagination? is that also broken? I would suspect it is.

james-lawrence:
ops misread the code, yes pagination is broken for this reason. going to fix conflict and merge. :disappointed: my bad.

tbroadley:
Nice, much appreciated!

PR #312, opened 2018-05-09
Misc bugfixes

james-lawrence:
- reverts to using SlackRepsonse (sorry anyone who was impacted by this)
- fixes a mutex being left locked. (uncovered by errors being correctly returned, may be other issues to resolve in this area now)
- fixes invalid package names in slackevents.


james-lawrence:
@ACollectionOfAtoms we both missed the packages issue, but thought I'd ping you just as an FYI since IIRC you said you were new to go.

PR #314, opened 2018-05-10
add custom UnmarshalJSON to JSONTime

nl5887:
the UnmarshalJSON function will unmarshal both json string and ints, this
prevents issues while importing slack archives.


info.go
```diff
@@ -127,6 +128,23 @@ func (t JSONTime) Time() time.Time {
 	return time.Unix(int64(t), 0)
 }
 
+// UnmarshalJSON will unmarshal both string and int JSON values
+func (t *JSONTime) UnmarshalJSON(buf []byte) error {
+	s := buf
+
+	if s[0] == '"' {
```
james-lawrence: may result in a panic.

mind switching to using trim?

```
bytes.Trim(buf, `"`)
```

nl5887:
Good one, thx & fixed.

PR #320, opened 2018-05-22
Capture response metadata properly

freeformz:
This also removes an extraneous log statement.


freeformz:
I am using this patch applied to vendor here: https://github.com/flexd/slackinviter/pull/25 which powers https://invite.slack.golangbridge.org/

theckman (approved):

PR #321, opened 2018-05-23
Added support for interactive message buttons & Extended functionality of SendMessage

alexstojda:


slackevents/parsers.go
```diff
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"reflect"
 
-	"github.com/nlopes/slack"
+	"github.com/alexstojda/slack"
```
james-lawrence: these need to be fixed.

chat.go
```diff
@@ -128,13 +136,20 @@ func (api *Client) PostEphemeralContext(ctx context.Context, channelID, userID s
 }
 
 // UpdateMessage updates a message in a channel
-func (api *Client) UpdateMessage(channelID, timestamp, text string) (string, string, string, error) {
-	return api.UpdateMessageContext(context.Background(), channelID, timestamp, text)
+func (api *Client) UpdateMessage(channelID, timestamp, text string, params UpdateMessageParameters) (string, string, string, error) {
```
james-lawrence: changes to the arguments of this function won't be accepted. imo its a dead code path.

SendMessage is easier to extend.

chat.go
```diff
@@ -385,3 +400,19 @@ func MsgOptionPostMessageParameters(params PostMessageParameters) MsgOption {
 		return nil
 	}
 }
+
+func MsgOptionUpdateMessageParmeters(params UpdateMessageParameters) MsgOption {
```
james-lawrence: not sure what these add that isn't already available via SendMessage.

slackevents/action_events.go
```diff
@@ -0,0 +1,42 @@
...
+	ActionTimestamp  json.Number              `json:"action_ts"`
+	MessageTimestamp json.Number              `json:"message_ts"`
+	AttachmentId     json.Number              `json:"attachment_id"`
+	Token            string                   `json:"token"`
+	OriginalMessage  slack.Message            `json:"original_message"`
+	ResponseUrl      string                   `json:"response_url"`
+	TriggerId        string                   `json:"trigger_id"`
+}
+
+func UnmarshallMessageAction(payloadString string, p *MessageAction) {
```
james-lawrence: Unnecessary code. and swallows an error. also where does the print function come from.... that should be removed as well.

james-lawrence (changes requested):

james-lawrence:
changes to UpdateMessage seem unnecessary for this PR.

chat.go
alexstojda: Changes to the function are necessary to fully support the [chat.update](https://api.slack.com/methods/chat.update) API endpoint. 

Presently, the UpdateMessage function here only accepts updates to the text of a message itself. However, if a user wants to, for example, modify the attachments of an existing message, the present implementation does not allow this. My modifications allow for the full use of all the available options for `chat.update` defined in the slack API docs.

Please let me know if that clarifies things a bit for you on this, I'd be more than happy to further explain if needed.

chat.go
alexstojda: As mentioned in [this comment](https://github.com/nlopes/slack/pull/321/files#r194226603), it adds the missing functionality that the chat.update API endpoint provides, which is not yet used by the original definition of the UpdateMessage function.

slackevents/action_events.go
alexstojda: The function was wrongly named. It was, in fact, the parser for the MessageAction. I have moved it to the proper location of parsers.go and added the token verification option function just like the regular event parser.

chat.go
james-lawrence: Let me rephrase my original comment a little bit.

I know what you're trying to accomplish, however its already been accomplished via the SendMessage API. UpdateMessage is effectively deprecated functionality. No changes to it will be accepted that break compatibility.

If you feel some option is missing from the SendMessage API to accomplish your task then by all means add the necessary Option functions.

Then write the ~5 line function you want for your application's UpdateMessage Option function in your application.

slackevents/parsers.go
```diff
@@ -192,3 +192,24 @@ func ParseEvent(rawEvent json.RawMessage, opts ...Option) (EventsAPIEvent, error
...
+	}
+
+	cfg := &Config{}
+	cfg.VerificationToken = action.Token
+	for _, opt := range opts {
+		opt(cfg)
+	}
+
+	if !cfg.TokenVerified {
+		return MessageAction{}, errors.New("no")
```
james-lawrence: see https://github.com/nlopes/slack/pull/311#discussion_r193254321

chat.go
alexstojda: Ok, I see what you mean with SendMessage. 

That being said, I still feel that if you are going to implement the functionality of an endpoint, you should implement its full functionality as defined by the API provider. 

Also, is the change really breaking? If an argument isn't provided it will simply not add any extra information to the API request call, no? Thus any code that already uses this functionality should not experience any changes?

chat.go
james-lawrence: Here is an example: (not current possible without some minor additions):
In your application impement the following:
```
func MsgOptionUpdate(ts string) slack.MessageOption {
  return slack.MsgOptionCompose(slack.MsgOptionUpdate(ts), slack.MsgOptionAsUser(true), slack.MsgOptionLinkNames(true), slack.MsgOptionParse(true))
}
```
What is needed by this library:
```
func MsgOptionCompose(options ...MsgOption) MsgOption { ... }
func MsgOptionParse(b bool) MsgOption { ... }
func MsgOptionLinkNames(b bool) MsgOption { ... }
```

slackevents/parsers.go
alexstojda: I was wondering why the original error from ParseEvent was just "no" lol, I changed it to the same error text.

chat.go
alexstojda: @james-lawrence Would you mind explaining your example a bit more? I don't really follow. 

Also please read my reply to your other comment above :)

chat.go
james-lawrence: >is the change really breaking

yup definitely breaking. if code will not compile its a breaking change. We even consider some subtler changes where the API is identical but behaviour changes to be breaking.

We try to limit these events where possible. see issues related to API breaking changes that havent been done yet. This is the very reason UpdateMessage, PostMessage, DeleteMessage still exist. because removing them would break API compatibility. If we were willing to break compatibility I'd just remove them.

>I still feel that if you are going to implement the functionality of an endpoint, you should implement its full functionality as defined by the API provider.

Agreed, we differ on the definition of implemented. =) I'm claiming everything you are trying to accomplish can be (with possibly some minor extensions) using the existing SendMessage API.

alexstojda: Alright, I'll remove the changes for the sake of the rest of this PR.

chat.go
james-lawrence: MsgOptionUpdate -> your codes setup for performing a update message. this is specific to your code. everyone has a different implementation. some people update as user, some don't. others do not parse messages, some do. some add attachments, some don't. There is no one size fits all.

this is why it doesn't belong in the library.

The remaining 3 functions are what need to be implemented to replace your current changes to UpdateMessage. These methods are what should be in your PR to extend the functionality.
```
// this function combines multiple options into a single option.
func MsgOptionCompose(options ...MsgOption) MsgOption { 
  return func(c *sendConfig) {
     for _, opt := range options { opt(c) }
  }
 }
```

```
func MsgOptionParse(b bool) MsgOption {
  return func(c *sendConfig) {
    var v string
    if b { v = "1" } else { b = "0" }
    c.values.Set("parse", v)
  }
}
```

```
func MsgOptionUpdate(ts string) slack.MsgOption {
  return slack.MsgOptionCompose(slack.MsgOptionUpdate(ts), slack.MsgOptionAsUser(true), slack.MsgOptionLinkNames(true), slack.MsgOptionParse(true))
}

// in your application
client.SendMessage(channelId, msg, MsgOptionUpdate(ts), MsgOptionAttachments(a))
```

chat.go
alexstojda: However, please elaborate on your example so I can make the implementation in my project

james-lawrence: see https://github.com/nlopes/slack/pull/321#discussion_r194229460 =)

chat.go
alexstojda: Alright added the functions in the latest commit

chat.go
```diff
@@ -331,6 +331,23 @@ func MsgOptionDisableMarkdown() MsgOption {
 	}
 }
 
+// this function combines multiple options into a single option.
+func MsgOptionCompose(options ...MsgOption) MsgOption {
+	return func(c *sendConfig) error {
+		for _, opt := range options { opt(c) }
```
james-lawrence: you'll want to return the first error from applying the options.

forgot the signature included an error or I would have put that in the example.

james-lawrence (approved):

PR #323, opened 2018-05-24
Add call to users.conversations

asvoboda:
Implements https://api.slack.com/methods/users.conversations

### DIFF
--- a/conversation.go
+++ b/conversation.go
@@ -64,6 +64,13 @@ type GetUsersInConversationParameters struct {
 	Limit     int
 }
 
+type GetConversationsForUserParameters struct {
+	UserID string
+	Cursor string
+	Types  []string
+	Limit  int
+}
+
 type responseMetaData struct {
 	NextCursor string `json:"next_cursor"`
 }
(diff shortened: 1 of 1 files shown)


asvoboda:
Is there anything else that you'd like me to add to this PR?

conversation.go
```diff
@@ -100,6 +107,41 @@ func (api *Client) GetUsersInConversationContext(ctx context.Context, params *Ge
 	return response.Members, response.ResponseMetaData.NextCursor, nil
 }
 
+// GetConversationsForUser returns the list conversations for a given user
+func (api *Client) GetConversationsForUser(params *GetConversationsForUserParameters) (channels []Channel, nextCursor string, err error) {
+	return api.GetConversationsForUserContext(context.Background(), params)
+}
+
+// GetConversationsForUserContext returns the list conversations for a given user with a custom context
+func (api *Client) GetConversationsForUserContext(ctx context.Context, params *GetConversationsForUserParameters) (channels []Channel, nextCursor string, err error) {
```
james-lawrence, line 116: it'd be better if you returned the entire GetConversationsForUserParameters data for use. slightly easier to use as an API.

james-lawrence (commented):
overall looks fine, 1 minor comment. sorry about the delay in reviewing.

conversation.go
asvoboda: It seems like the rest of the conversation.go method have the same api that lets delegates that work to the caller. I didn't want to introduce a new method that functioned differently. I can definitely add another higher level method for the caller to return all data if you'd like.

james-lawrence: fair point, can always change it later.

conversation.go
```diff
@@ -100,6 +107,41 @@ func (api *Client) GetUsersInConversationContext(ctx context.Context, params *Ge
...
+	}
+	if params.Types != nil {
+		values.Add("types", strings.Join(params.Types, ","))
+	}
+	response := struct {
+		Channels         []Channel        `json:"channels"`
+		ResponseMetaData responseMetaData `json:"response_metadata"`
+		SlackResponse
+	}{}
+	err = post(ctx, api.httpclient, "users.conversations", values, &response, api.debug)
```
richmondwang, line 135: this should be `postSlackMethod`? 
builds are failing now.

suzuki-shunsuke: I send PR to fix it.

https://github.com/nlopes/slack/pull/401

PR #311, opened 2018-05-09
Add thread timestamp to message event

pa3ng:
Status: **POST REVIEW CHANGES DONE**

- Added the `thread_ts` field to the`MessageEvent` struct for being able to reply to threads (docs: https://api.slack.com/docs/message-threading)

- Updated the token verification error message to be more informative.

PR #328, opened 2018-06-05
Add ThreadTimeStamp to MessageEvent

ruudk:


pa3ng:
@ruudk I think this is a duplicate of PR #311

ruudk:
You're right, sorry!

PR #329, opened 2018-06-05
Add Hidden to MessageEvent

ruudk:


slackevents/inner_events.go
```diff
@@ -67,6 +67,7 @@ type MessageEvent struct {
 	Channel        string      `json:"channel"`
 	ChannelType    string      `json:"channel_type"`
 	EventTimeStamp json.Number `json:"event_ts"`
+	Hidden         bool        `json:"hidden"`
```
whithajess, line 70: Its worth adding this in line with what is over here: https://github.com/nlopes/slack/blob/master/messages.go#L39

whithajess (changes requested):

slackevents/inner_events.go
ruudk: Actually, it would make more sense if we used the same Msg struct.

PR #330, opened 2018-06-08
Add support for godep

kei-yamazaki:
Thank you for providing a nice library!

I believe that adding support for godep will make development easier.


kei-yamazaki:
[slack-test imports](https://github.com/lusis/slack-test/blob/3c758769bfa68068b76a8a65a26baee9c5a7391f/types.go#L9) `nlopes/slack's` WebResponse, but this has been reverted with [this commit](https://github.com/nlopes/slack/commit/8e0186d9e989fd22319f9c5dcec949092305a6c8).
`lusis/slack-test` and `nlopes/slack` are a bit difficult as they are circular references.

kei-yamazaki:
I reverted https://github.com/nlopes/slack/commit/8e0186d9e989fd22319f9c5dcec949092305a6c8.
It may be necessary to discuss whether this commit is really necessary.

james-lawrence:
Its not necessary in the strictness sense. but this will not be accepted with it reverted.

Both SlackResponse/WebResponse were serving the same purpose in this package and both were exported (unfortunately) so different packages are relying on different names. We arbitrarily settled on SlackResponse as WebResponse was too general a name.

I suggest a PR to slacktest to update its usage.

james-lawrence:
in the mean time, you can just ignore slacktest from deps as its a testing only dependency.

kei-yamazaki:
Thank you for your response. understood.
I ignored the slack-test according to your idea.

PR #346, opened 2018-06-26
Feat: add some helper to Info, like GetFooByName

flw-cn:
Feat: add some useful helper to Info, like GetFooByName

+ add GetChannelByName(channelName string) *Channel
+ add GetGroupByName(groupName string) *Group
+ add GetUserByName(userName string) []*User

re-implement these function for speedup:

* GetUserByID(userID string) *User
* GetGroupByID(groupID string) *Group
* GetIMByID(imID string) *IM
* GetBotByID(botID string) *Bot
* GetChannelByID(channelID string) *Channel

### DIFF
--- a/info.go
+++ b/info.go
@@ -144,14 +144,23 @@ type Icons struct {
 // Info contains various details about Users, Channels, Bots and the authenticated user.
 // It is returned by StartRTM or included in the "ConnectedEvent" RTM event.
 type Info struct {
-	URL      string       `json:"url,omitempty"`
-	User     *UserDetails `json:"self,omitempty"`
-	Team     *Team        `json:"team,omitempty"`
-	Users    []User       `json:"users,omitempty"`
-	Channels []Channel    `json:"channels,omitempty"`
-	Groups   []Group      `json:"groups,omitempty"`
-	Bots     []Bot        `json:"bots,omitempty"`
-	IMs      []IM         `json:"ims,omitempty"`
+	URL        string              `json:"url,omitempty"`
+	User       *UserDetails        `json:"self,omitempty"`
+	Team       *Team               `json:"team,omitempty"`
+	Users      []User              `json:"users,omitempty"`
+	Channels   []Channel           `json:"channels,omitempty"`
+	Groups     []Group             `json:"groups,omitempty"`
+	Bots       []Bot               `json:"bots,omitempty"`
+	IMs        []IM                `json:"ims,omitempty"`
+	id2user    map[string]*User    // used for locate user by id
+	name2user  map[string][]*User  // used for locate user by name
+	id2bot     map[string]*Bot     // used for locate bot by id
+	name2bot   map[string][]*Bot   // used for locate bot by name
+	id2chan    map[string]*Channel // used for locate channel by id
+	name2chan  map[string]*Channel // used for locate channel by name
+	id2group   map[string]*Group   // used for locate group by id
+	name2group map[string]*Group   // used for locate group by name
+	id2im      map[string]*IM      // used for locate IM by id
 }
 
 type infoResponseFull struct {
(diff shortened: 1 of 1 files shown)


james-lawrence:
very unlikely to merge this.

these can all by implemented using a Reduce function for the different structures and is far less complicated and more flexible. ~ 10 lines of code per type.

I also believe slack has methods for some of these to retrieve values by a name.

flw-cn:
@james-lawrence Thank you for your reply. But I still have some questions.

> these can all by implemented using a Reduce function for the different structures and is far less complicated and more flexible. ~ 10 lines of code per type.

I don't quite understand what you mean, can you tell me what is better?

> I also believe slack has methods for some of these to retrieve values by a name.

Which method can do this?

james-lawrence:
@flw-cn I stand corrected, I was thinking of the https://api.slack.com/methods/users.lookupByEmail which is not lookup by username.


regarding the reduce functionality here is an example of a before (your code) and what I am suggesting:
```golang
// GetUserByName retrieves user(maybe more than one) information by user name
func (info *Info) GetUserByName(userName string) []*User {
	if info.name2user == nil {
		info.name2user = make(map[string][]*User, 32)
	} else if list := info.name2user[userName]; list != nil {
		return list
	}

	for i, user := range info.Users {
		info.name2user[user.Name] = append(info.name2user[user.Name], &info.Users[i])
	}

	return info.name2user[userName]
}
```

```golang
SelectUsersBy(users []User, pred func(User) bool) (results []User) {
  for _, u := range users {
    if pred(u) {
      results = append(results, u)
    }
  }
  return results
}

filtered := SelectUserBy(info.Users, UserByName("foobar"))
```

the map you're creating is pretty memory intensive for larger teams and Info is being deprecated by slack because its costly to build. creating this cache should be left up to the users of the library to decide if they want that sort of behavior.

flw-cn:
@james-lawrence Thank you for your patience!

I am beginning to agree with you a bit, that is, if the library is considered to be a simple Slack API, then indeed possible to put that extra works on the user side.

As we discussed, the Slack API does not provide `users.lookupByName`, `channels.lookupByName` and `groups.lookupByName`, so the library user has no choice but to send a message via `slack.Client.SendMessage("CABCDEFGH", ...)` instead of `slack.Client.SendMessage("#topic", ...)`. `user` and `group` are similar to that.

So let's stand on the user's side, what is the easiest way to send a channel message? 

Worse, we can't get a channel ID from Slack UI simply. So we must store that maps by myself. This means that we must wrap another layer base on the Slack API, which goes against the principle of out of the box.

I admit that my code doesn't look so neat, and the naming is a bit ugly, but it solves the problem and makes it easy for users to send and logging messages.

As for the memory waste you mentioned, because I use pointers instead of replicas, the added maps should not consume too much memory.

On the other hand, because of the existence of the maps, the time wasted by `for range` is reduced, so there is a saving in time. This is also in line with the principle of used space to exchange time.

Is this any propose to improve it and leading it possible to merge into the repository? I hope to hear from you.

james-lawrence:
I hear you, but you're trying to solve a user experience issue in the library. which is basically a no-go.

At the end of the day you're still going to be required to fetch the list of channels in order to display which options are available to the user, or rely on the user to provide the channel name in which case you're going to need to fetch the channels to lookup the ID.

Either way you end up fetching the list of entities. Now you can start caching these results to speed up the time to completion of the request, which is wonderful. but caching implementations are almost always application specific.

All that being said, I'm not against including a basic implementation for cache implementations for various common tasks for applications that may be using this library. just not this implementation. =)

checkout the slackutilsx package. I'd recommend putting said functionality there.

james-lawrence:
closing due to age, and lack of activity.

PR #348, opened 2018-06-26
Parse timestamp as json.Number

t-vorobyova:
- type json.Number is an alias for string
- also see http://stackoverflow.com/questions/24480835/go-decoding-json-int-into-string

### DIFF
--- a/files.go
+++ b/files.go
@@ -2,6 +2,7 @@ package slack
 
 import (
 	"context"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


james-lawrence:
requires a test prior to merging. as this has the chance to be broken in the future due to other edge cases. slack's api has a ton of these inconsistencies.

james-lawrence:
closing due to age.

PR #349, opened 2018-06-27
Fix ephemeral message timestamp

edwardIshaq:
a possible fix for https://github.com/nlopes/slack/issues/347
Added a new variable called `MessageTimeStamp` to `chatResponseFull` struct with an function to fetch the timestamp so we can return in an ephemeral post message

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -23,10 +23,18 @@ const (
 )
 
 type chatResponseFull struct {
-	Channel   string `json:"channel"`
-	Timestamp string `json:"ts"`
-	Text      string `json:"text"`
+	Channel          string `json:"channel"`
+	Timestamp        string `json:"ts"`
+	MessageTimeStamp string `json:"message_ts"`
+	Text             string `json:"text"`
 	SlackResponse
+}
+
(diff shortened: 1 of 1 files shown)


chat.go
```diff
@@ -23,12 +23,20 @@ const (
...
-	Timestamp string `json:"ts"`
-	Text      string `json:"text"`
+	Channel          string `json:"channel"`
+	Timestamp        string `json:"ts"`
+	MessageTimeStamp string `json:"message_ts"`
+	Text             string `json:"text"`
 	SlackResponse
 }
 
+func (c chatResponseFull) getMessageTimestamp() string {
```
james-lawrence: mind adding the notes from your issue to a comment inside this method? otherwise LGTM.
useful information for people in the future looking at this code and being like WTF.

james-lawrence (commented):
LGTM, minor request.

chat.go
edwardIshaq: Sounds like a good idea @james-lawrence 
Thank you 🙇

Issue #280, opened 2018-03-16
Allow NewOutgoingMessage to reply in thread


Gardnem6:
In message.go, there's a property for thread_ts in the OutgoingMessage struct that isn't used on the NewOutgoingMessage function.  It would be nice to allow this property to be set as part of the simple message, since the thread_ts is necessary to reply in thread to a user.  I can write and pr the change myself if preferred, but the change looks pretty simple.

james-lawrence:
feel free.

tylarb:
Submitted PR https://github.com/nlopes/slack/pull/351 on which should resolve this

ghostsquad:
This should probably be closed now that #351 is merged?

PR #351, opened 2018-07-05
Added a timestamp string argument to the rtm.NewOutgoingMessage function

tylarb:
Enables function to be used to reply to or start a new thread

Per slack guidelines, an empty string starts a new thread/posts directly to channel, but any TS string that matches a previous post adds a reply. Test fix just passes an empty string.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -331,6 +331,22 @@ func MsgOptionDisableMarkdown() MsgOption {
 	}
 }
 
+// MsgOptionTS sets the thread TS of the message to enable creating or replying to a thread
+func MsgOptionTS(ts string) MsgOption {
+	return func(config *sendConfig) error {
+		config.values.Set("thread_ts", ts)
+		return nil
+	}
+}
+
+// MsgOptionBroadcast sets reply_broadcast to true
+func MsgOptionBroadcast() MsgOption {
+	return func(config *sendConfig) error {
+		config.values.Set("reply_broadcast", "true")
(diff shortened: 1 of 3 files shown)


james-lawrence:
break API compat, i'd recommend some option style varg instead.

tylarb:
Updated the commit to use a structure similar to that in chat.go. 

I also added a "ThreadBroadcast" bool to the outgoing message struct - according to [SlackAPI](https://api.slack.com/docs/message-threading), "ThreadBroadcast" the title of the feature, but it actually looks like "reply_broadcast" is the json bool which is handled correctly when posting - the docs don't give a relevant example, but see the [python API for reference](https://github.com/slackapi/python-slackclient/blob/8a6e36e55e56a3e41154b824a883fd16a272dc8f/slackclient/client.py#L138)

I tested both of these features, and the syntax should be sufficient to extend in the future, although I don't know how much it will be required to as the RTM API will seem to be somewhat feature limited compared to the Events API, etc.

tylarb:
I also think this thread + thread broadcast could be included as an option in the chat API but this PR currently resolves my immediate need as I'm only using RTM for now. 

Do you want any change/addition to the chat as well, or is that OK to be in a separate PR?

james-lawrence:
Would love the addition to the chat api. will take it however you wish to provide it =)

ghostsquad:
Resolves #280

PR #356, opened 2018-07-16
Add convience method for webhooks

### LINKED ISSUE #355: Webhook Message Posting
This api doesn't seem to support posting messages to a webhook url. At least I can't find it.

I'd like to contribute if that's the case.

ghostsquad:
Resolves #355

### DIFF
--- a/dialog_test.go
+++ b/dialog_test.go
@@ -299,7 +299,7 @@ func TestOpenDialog(t *testing.T) {
 	}
 	err = api.OpenDialog("", *dialog)
 	if err == nil {
-		t.Errorf("Did not error with empty trigger: %s", err)
+		t.Errorf("Did not error with empty trigger, %s", err)
 		return
 	}
 }
--- a/webhooks.go
+++ b/webhooks.go
@@ -1,0 +1,33 @@
+package slack
+
+import (
+	"github.com/pkg/errors"
+	"net/http"
+	"bytes"
+	"encoding/json"
+)
+
+type WebhookMessage struct {
+	Text    		string 				`json:"text,omitempty"`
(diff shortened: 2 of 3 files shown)


dialog_test.go
```diff
@@ -298,7 +298,7 @@ func TestOpenDialog(t *testing.T) {
 	}
 	err = api.OpenDialog("", *dialog)
 	if err == nil {
-		t.Errorf("Did not error with empty trigger", err)
```
ghostsquad: I found this problem when I ran tests locally.

webhooks.go
```diff
@@ -0,0 +1,33 @@
+package slack
+
+import (
+	"github.com/pkg/errors"
+	"net/http"
+	"bytes"
+	"encoding/json"
+)
+
+type WebhookMessage struct {
```
james-lawrence, line 10: do we need another type here or will the standard type Message (or one of its variants) work?

ghostsquad: The slack documentation indicates that only these fields are used. It would be confusing/bad design to use a struct with fields that are unused by the request.

ghostsquad: The solution to this would be to use interfaces for inputs. That would require custom serialization.

dialog_test.go
RaMin0: Fixed in 7a2193d189e649d76c4a90e274e49b24b725fcc5.

ghostsquad:
@james-lawrence I've resolved merge conflicts. Please re-evaluate if this PR is ok to merge. Thanks!

james-lawrence:
LGTM.

Issue #378, opened 2018-08-07
Bug: incorrect struct tag for MessageAction


billglover:
The struct tag here is `original_message`:

https://github.com/nlopes/slack/blob/21749ab136a8739656c58a53e95836f9ce280a13/slackevents/action_events.go#L30

The impact of this is that the Original message is not parsed when unmarshaling MessageActions.

Based on the [Slack documentation](https://api.slack.com/actions) this should be `message`.

billglover:
Let me know if you'd like me to submit a Pull Request and I'll be happy to provide a fix.

jameslafa:
+1, I just noticed the same error and what about to report this issue.

PR #379, opened 2018-08-07
Rename message json field inside MessageAction

jameslafa:
Here is a simple pull request to fix the issue #378.

### DIFF
--- a/slackevents/action_events.go
+++ b/slackevents/action_events.go
@@ -27,7 +27,7 @@ type MessageAction struct {
 	MessageTimestamp json.Number              `json:"message_ts"`
 	AttachmentId     json.Number              `json:"attachment_id"`
 	Token            string                   `json:"token"`
-	OriginalMessage  slack.Message            `json:"original_message"`
+	OriginalMessage  slack.Message            `json:"message"`
 	ResponseUrl      string                   `json:"response_url"`
 	TriggerId        string                   `json:"trigger_id"`
 }


james-lawrence:
thanks btw.

cjsaylor:
@james-lawrence I think there may be a problem with this. While this may fix in one area, it looks like interactive messages flow through this `MessageAction` as well, and it has "original_message" as the key in order to allow replacing the message with out the actions in it.

See https://api.slack.com/interactive-messages

So this change breaks interactive messaging (my use case).

james-lawrence:
well shit... I'll push a change tonight. we'll just put both fields.

jameslafa:
Yes I was not very confident pushing this without tests actually, I was lacking a bit of time. 
I think we should add that as soon as we can to make sure it does not happen again.

james-lawrence:
problem is I'm not terribly familiar with the events API and slack occasionally makes changes to the structure of their api.

so minor changes like this which seem reasonable given relevant documentation can get overlooked.

berfarah:
Question (from a completely uninformed standpoint here): why are there two types for this?

In `slackevents`:
https://github.com/nlopes/slack/blob/752f784a75e8b640f048a8a453466cca64c83a47/slackevents/action_events.go#L20-L35

and in `slack`:
https://github.com/nlopes/slack/blob/752f784a75e8b640f048a8a453466cca64c83a47/attachments.go#L44-L63

Which of these is the maintained and/or preferred way to interact with actions?

james-lawrence:
MessageAction is for the events API.
slack/attachments is for the interactions API. and is being refactored into an interactions structure.

your use should depend on which slack API you're using. they exist because no one has ensured they truly are identical and be used interchangeably and that it will remain that way.

berfarah:
@james-lawrence Thanks for taking the time to respond. Sounds good! I'm only using the interactive API, so I think for now I'll use the former. One thing to note: there's a very minor difference between the two structs where `ID` and `URL` are capitalized in `slack/attachments.go` and not in `slackevents`.

james-lawrence:
feel free to submit a fix for slackevents.

berfarah:
@james-lawrence cheers, done! Thanks for maintaining!

PR #386, opened 2018-08-13
Expose team ID from the originating event when parsed.

cjsaylor:
This makes it possible to know the source team for event subscriptions for slack apps that operate across multiple work spaces. Without the team ID, it is not possible to determine which oauth token to use for interactions with that event subscription.


james-lawrence:
LGTM.

PR #392, opened 2018-09-01
Enhance GetUserGroups call with params

gouline:
- Create `GetUserGroupsParams` struct with field descriptions (from the [docs](https://api.slack.com/methods/usergroups.list#arguments))
- Propagate new parameters to `GetUserGroups` and `GetUserGroupsContext`
- Add list of users to `UserGroup` struct (when `IncludeUsers` is enabled)
- Update unit test to include a list of users

Heavily inspired by #218, which was never merged, but needed the users to be included as well.

### DIFF
--- a/usergroups.go
+++ b/usergroups.go
@@ -122,32 +122,50 @@ func (api *Client) EnableUserGroupContext(ctx context.Context, userGroup string)
 	return response.UserGroup, nil
 }
 
-// GetUserGroupsParams contains arguments for GetUserGroups call
-type GetUserGroupsParams struct {
-	// IncludeCount	- Include the number of users in each User Group. Default: false
-	IncludeCount bool
-	// IncludeDisabled - Include disabled User Groups. Default: false
-	IncludeDisabled bool
-	// IncludeUsers - Include the list of users for each User Group. Default: false
-	IncludeUsers bool
-}
+// GetUserGroupsOption options for the GetUserGroups method call.
+type GetUserGroupsOption func(*GetUserGroupsParams)
 
-// NewGetUserGroupsParams creates a new instance GetUserGroupsParams with defaults
-func NewGetUserGroupsParams() GetUserGroupsParams {
-	return GetUserGroupsParams{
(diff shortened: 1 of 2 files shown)


usergroups.go
```diff
@@ -121,16 +122,44 @@ func (api *Client) EnableUserGroupContext(ctx context.Context, userGroup string)
...
+	// IncludeCount	- Include the number of users in each User Group. Default: false
+	IncludeCount bool
+	// IncludeDisabled - Include disabled User Groups. Default: false
+	IncludeDisabled bool
+	// IncludeUsers - Include the list of users for each User Group. Default: false
+	IncludeUsers bool
+}
+
+// NewGetUserGroupsParams creates a new instance GetUserGroupsParams with defaults
+func NewGetUserGroupsParams() GetUserGroupsParams {
```
james-lawrence: this is kind of pointless, zero value is already false. this entire function can be dropped and replaced with `GetUserGroupsParams{}`

usergroups.go
```diff
@@ -121,16 +122,44 @@ func (api *Client) EnableUserGroupContext(ctx context.Context, userGroup string)
...
+		IncludeCount:    false,
+		IncludeDisabled: false,
+		IncludeUsers:    false,
+	}
+}
+
 // GetUserGroups returns a list of user groups for the team
-func (api *Client) GetUserGroups() ([]UserGroup, error) {
-	return api.GetUserGroupsContext(context.Background())
+func (api *Client) GetUserGroups(params GetUserGroupsParams) ([]UserGroup, error) {
```
james-lawrence: as this breaks the API, I would recommend switching to an (options ...option) pattern. plenty of examples in the library.

james-lawrence (changes requested):

gouline:
@james-lawrence thanks for your review, I addressed both comments.

james-lawrence (approved):

james-lawrence:
thanks mgouline!

PR #393, opened 2018-09-04
Dialog improvements

james-lawrence:
attempted to take the best parts of both sets of dialog work.

- marking dialogs as experiemental until api is consistent within the library.
- fix missing dependency added by webhooks.


edwardIshaq:
Thanks for adding this @james-lawrence  🙌 
Sorry I didnt get around to finishing it up on my last cycle

PR #398, opened 2018-09-12
Support the thread_ts parameter of files.upload API

suzuki-shunsuke:
https://api.slack.com/methods/files.upload


james-lawrence:
thanks as always.

PR #399, opened 2018-09-13
Add ability to verify via signing secrets for improved security

alyosha:
Verification tokens have been deprecated by Slack and, while still usable, are far less secure than the new signing secrets method of verifying requests. This PR adds `security.go`, which will allow users to easily judge the validity of their incoming requests. 

Reference: https://api.slack.com/docs/verifying-requests-from-slack#about

### DIFF
--- a/security.go
+++ b/security.go
@@ -46,5 +46,5 @@ func (v SecretsVerifier) Ensure(signingSecret string) error {
 		return nil
 	}
 
-	return errors.New("invalid token")
+	return errors.New("invalid request")
 }
--- a/security.go
+++ b/security.go
@@ -20,7 +20,7 @@ type SecretsVerifier struct {
 // NewSecretsVerifier returns a SecretsVerifier object in exchange for an http.Header object and signing secret
 func NewSecretsVerifier(header http.Header, signingSecret string) (SecretsVerifier, error) {
 	if header["X-Slack-Signature"][0] == "" || header["X-Slack-Request-Timestamp"][0] == "" {
-		return SecretsVerifier{}, errors.New("headers are empty, cannot create SecretsVerifier")
+		return SecretsVerifier{}, errors.New("Headers are empty, cannot create SecretsVerifier")
 	}
 
 	hash := hmac.New(sha256.New, []byte(signingSecret))
@@ -43,5 +43,5 @@ func (v SecretsVerifier) Ensure() error {
 		return nil
 	}
 
-	return fmt.Errorf("invalid request verification token %s, expected %s", v.slackSig, computed)
+	return fmt.Errorf("Expected signing signature: %s, but computed: %s", v.slackSig, computed)
 }
--- a/security.go
+++ b/security.go
@@ -37,7 +37,7 @@ func (v *SecretsVerifier) Write(body []byte) (n int, err error) {
 }
 
 // Ensure compares the signature sent from Slack with the actual computed hash to judge validity
-func (v SecretsVerifier) Ensure(signingSecret string) error {
+func (v SecretsVerifier) Ensure() error {
 	computed := "v0=" + string(hex.EncodeToString(v.hmac.Sum(nil)))
 	if computed == v.slackSig {
 		return nil
--- a/security.go
+++ b/security.go
@@ -6,45 +6,42 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"hash"
 	"net/http"
 )
 
 // SecretsVerifier contains the information needed to verify that the request comes from Slack
 type SecretsVerifier struct {
-	slackSig    string
-	timeStamp   string
-	requestBody string
+	slackSig  string
+	timeStamp string
+	hmac      hash.Hash
 }
 
-// NewSecretsVerifierFromHeader returns a new SecretsVerifier object in exchange for an http.Header object
-func NewSecretsVerifierFromHeader(header http.Header) (SecretsVerifier, error) {
+// NewSecretsVerifier returns a SecretsVerifier object in exchange for an http.Header object and signing secret
+func NewSecretsVerifier(header http.Header, signingSecret string) (SecretsVerifier, error) {
 	if header["X-Slack-Signature"][0] == "" || header["X-Slack-Request-Timestamp"][0] == "" {
 		return SecretsVerifier{}, errors.New("headers are empty, cannot create SecretsVerifier")
 	}
 
+	hash := hmac.New(sha256.New, []byte(signingSecret))
+	hash.Write([]byte(fmt.Sprintf("v0:%s:", header["X-Slack-Request-Timestamp"][0])))
 	return SecretsVerifier{
 		slackSig:  header["X-Slack-Signature"][0],
 		timeStamp: header["X-Slack-Request-Timestamp"][0],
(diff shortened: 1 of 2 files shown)


security.go
```diff
@@ -0,0 +1,50 @@
...
+	"errors"
+	"fmt"
+	"net/http"
+)
+
+// SecretsVerifier contains the information needed to verify that the request comes from Slack
+type SecretsVerifier struct {
+	slackSig    string
+	timeStamp   string
+	requestBody string
```
james-lawrence: remove this, and replace it with a hmac field.

security.go
```diff
@@ -0,0 +1,50 @@
...
+
+// NewSecretsVerifierFromHeader returns a new SecretsVerifier object in exchange for an http.Header object
+func NewSecretsVerifierFromHeader(header http.Header) (SecretsVerifier, error) {
+	if header["X-Slack-Signature"][0] == "" || header["X-Slack-Request-Timestamp"][0] == "" {
+		return SecretsVerifier{}, errors.New("headers are empty, cannot create SecretsVerifier")
+	}
+
+	return SecretsVerifier{
+		slackSig:  header["X-Slack-Signature"][0],
+		timeStamp: header["X-Slack-Request-Timestamp"][0],
```
james-lawrence: pass the secret through the intializer. 

should be
```golang
return SecretsVerifier{
	slackSig:  header["X-Slack-Signature"][0],
	timeStamp: header["X-Slack-Request-Timestamp"][0],
        hmac: hmac.New(sha256.New, signingSecret),
}
```

security.go
```diff
@@ -0,0 +1,50 @@
...
+	}, nil
+}
+
+func (v *SecretsVerifier) Write(body []byte) (n int, err error) {
+	v.requestBody = string(body)
+	return len(body), nil
```
james-lawrence: afraid this method is a bug, you may not get the entire content in a single invocation of this method. it should be:

```golang
func (v *SecretsVerifier) Write(body []byte) (n int, err error) {
     return v.hmac.Write(body)
}
```

james-lawrence (changes requested):

security.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Ensure compares the signature sent from Slack with the actual computed hash to judge validity
+func (v SecretsVerifier) Ensure(signingSecret string) error {
+	message := fmt.Sprintf("v0:%v:%v", v.timeStamp, v.requestBody)
+
+	mac := hmac.New(sha256.New, []byte(signingSecret))
+	mac.Write([]byte(message))
+
+	actualSignature := "v0=" + string(hex.EncodeToString(mac.Sum(nil)))
```
james-lawrence:
```golang
func (v SecretsVerifier) Ensure() error {
      computed := "v0=" + string(hex.EncodeToString(v.hmac.Sum(nil)))
      if computed == v.slackSig {
         return nil
      }

     return errors.Errorf("invalid request verification token %s, expected %s", v.slackSig, computed)
}
```

security.go
alyosha: the signing secret is an environment variable obtained from the admin panel--it isn't contained in the header so are you suggesting passing it in as an argument? also it is a string so will need to be converted to bytes to pass in to `hmac.New`

security.go
alyosha: you need to concatenate "v0" and timestamp before hashing the resulting string, per Slack's instructions. will test to verify but I don't think this method will provide a valid signing secret
https://api.slack.com/docs/verifying-requests-from-slack

security.go
alyosha: You need to add "v0" back to the computed hash at the end to compare to the signature passed from Slack in the Header, but you also need to compute the hash will the full message string --> v0+timestamp+requestBody

alyosha93:
i implemented all the changes you describe above and tested, but the resulting hashes are different. unless I'm misunderstanding what you are suggesting, it doesn't seem like you are concatenating the necessary pieces together before hashing the string, and unless this is done precisely it results in differing hashes

security.go
james-lawrence: ah I missed the prefix, just add it to the hash during initialization.

security.go
james-lawrence: i'm psuedo coding for you, not ensuring everything is exactly correct.

security.go
james-lawrence:
```golang
hash := hmac.New(sha256.New, signingSecret)
hash.Write([]byte(fmt.Sprintf("v0:%s:", v.timeStamp)))
return SecretsVerifier{
	slackSig:  header["X-Slack-Signature"][0],
	timeStamp: header["X-Slack-Request-Timestamp"][0],
        hmac: hash,
}
```

alyosha: this still returns different hashes because you need to hash the prefix and the body together--don't think it's possible to do this all in the initializer

security.go
alyosha: understood-- thanks and sorry. just wanted to confirm we were on the same page about the meaning of the `signingSecret`

security.go
james-lawrence: simply not true. see [this example](https://play.golang.org/p/cBvHsjWhb8h)

james-lawrence: and just to make it [abundantly clear](https://play.golang.org/p/uiPH95wnZtK)

james-lawrence: just an FYI: most (all?) hashing algorithms are based on streams of data. so you can absolutely do partial writes for prefixes etc. and still end up with the same result. its the order of the content that gets hashed that matters.

alyosha: yeah sorry if my point was unclear, the wording was definitely bad. I get that you can write in a different order and still get the same result, my point was just that you aren't including it here so it will not return the right hash. also my point about not being able to do it all in the initializer was just that we would need to pass body in as an argument and that would complicate things so it doesn't seem like the best approach to me (the other option being calling the write function again on the request body back in the calling method) -- I'll make the changes and commit again to make it clearer, but thanks for being so thorough with your responses

alyosha:
@james-lawrence committed changes, no rush but please take a look when you have a minute m(_ _)m

security.go
```diff
@@ -0,0 +1,47 @@
...
+		hmac:      hash,
+	}, nil
+}
+
+func (v *SecretsVerifier) Write(body []byte) (n int, err error) {
+	return v.hmac.Write(body)
+}
+
+// Ensure compares the signature sent from Slack with the actual computed hash to judge validity
+func (v SecretsVerifier) Ensure(signingSecret string) error {
```
james-lawrence: arg not used anymore

alyosha: sorry, nice catch

alyosha: updated

james-lawrence (approved):

james-lawrence:
LGTM.

james-lawrence:
any chance of writing a simple test for the code?

alyosha:
sure thing, will hit you up when I've added the test

alyosha:
@james-lawrence just committed the test. probably didn't need to get so realistic with the dummy values but @Garethgr went through the trouble to make a working example in the issue so I thought why not just use his variables

james-lawrence:
beautiful. thanks so much.

PR #402, opened 2018-09-14
conversation: Fix call to post function

mtyurt:
Current version fails with error: `./conversation.go:137:8: undefined: post`

This error is introduced by the latest commit, I assume it's supposed to be `postSlackMethod`


suzuki-shunsuke:
This is duplicate to https://github.com/nlopes/slack/pull/401

PR #405, opened 2018-09-18
Count characters rather than bytes

veloek:
The limit suggested by Slack is referring to the number of characters, not the number of bytes which is returned from len() on a string. Converting to a rune slice is a quick way to get the correct number of characters in the message text.


james-lawrence:
LGTM, mind just rebasing against master? I made a mistake this morning when moving some personal branches around and had to fix master.

veloek:
Sure, no problem :slightly_smiling_face:

Issue #331, opened 2018-06-09
Update Example Directory


james-lawrence:
would be wonderful if any new users would like to update the examples directory with more up to date documentation.

PR #421, opened 2018-10-10
Update examples messages

ykrsm:
#331

The example was not working before because of updates on PostMessage
method.

### DIFF
--- a/examples/messages/messages.go
+++ b/examples/messages/messages.go
@@ -27,7 +27,7 @@ func main() {
 		},
 	}
 
-	channelID, timestamp, err := api.PostMessage("CHNNEL_ID", "Some text", params)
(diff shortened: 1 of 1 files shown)


james-lawrence:
looks like you were using an older version of the library? hence why tests are failing to compile for your change.

ykrsm:
Hello @james-lawrence,

Thank you for your comment.

Sorry for asking such a noob question but I am not sure how I got an older version of the library.

All I did was fork this repo on github and cloned it to my local machine.
I edited examples/messages/messages.go, commit, push, and made PR.

My guess is that before I fork this repo, I used `nlopes/slack` in my other project and used dep to resolve dependencies.
Do you think that might have some effect?

james-lawrence:
No idea, I don't make assumptions about local environments, or how people manage to get into bad states. I can only tell you what the code in its current state does.

PR #432, opened 2018-10-17
Added api.test endpoint

CodeMonk:
Added the api.test endpoint, which I use to test connection to slack.

### DIFF
--- a/api.go
+++ b/api.go
@@ -1,0 +1,45 @@
+package slack
+
+import (
+	"context"
+	"errors"
+	"net/url"
+)
+
+// APIResponseFull contains our API response from the api.test endpoint
+type APIResponseFull struct {
+	SlackResponse
+	Args map[string]string `json:"args,omitempty"`
+}
+
+func apiRequest(ctx context.Context, client httpClient, path string, values url.Values, debug debug) (*APIResponseFull, error) {
+	response := &APIResponseFull{}
+	err := postSlackMethod(ctx, client, path, values, response, debug)
+	if err != nil {
+		return nil, err
+	}
+	if !response.Ok {
+		return nil, errors.New(response.Error)
+	}
+	return response, nil
+}
+
+// GetApiTest will test the API by hitting the api.test endpoint, and
+// returning that response. It is a good way to make sure you're authenticated,
+// and talking to slack.
+func (api *Client) GetApiTest() (*APIResponseFull, error) {
+	return api.GetApiTestContext(context.Background())
+}
+
+// GetApiTestContext will retrieve the satus from api.test
+func (api *Client) GetApiTestContext(ctx context.Context) (*APIResponseFull, error) {
+	values := url.Values{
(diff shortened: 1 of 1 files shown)


api.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+	"errors"
+	"net/url"
+)
+
+// APIResponseFull contains our API response from the api.test endpoint
+type APIResponseFull struct {
+	SlackResponse
+	Args map[string]string `json:"args,omitempty"`
```
james-lawrence, line 12: yuck. lets avoid this, we are not in a dynamic language. also according to the documentation this is effectively unused because the client needs to provide the parameters to be returned (if I am reading it properly). which you've provided no way for users to supply these parameters.

api.go
```diff
@@ -0,0 +1,45 @@
...
+	if !response.Ok {
+		return nil, errors.New(response.Error)
+	}
+	return response, nil
+}
+
+// GetApiTest will test the API by hitting the api.test endpoint, and
+// returning that response. It is a good way to make sure you're authenticated,
+// and talking to slack.
+func (api *Client) GetApiTest() (*APIResponseFull, error) {
```
james-lawrence, line 30: lets be a little more idiomatic.

```go
func (api *Client) Test() (err error) {
   response := SlackResponse{}
   if err = postSlackMethod(ctx, client, path, values, &response, api.debug); err != nil {
      return err
   }

    return response.Err()
}
```

if we were to expose the full api becomes a little more complicated.
```go
func (api *Client) Test(payload json.RawMessage) (resp json.RawMessage, err error) {
   // ... again more complicated as you'll need to make the API call, get the raw response and then parse it to check for errors.
}
```

api.go
```diff
@@ -0,0 +1,45 @@
...
+
+// APIResponseFull contains our API response from the api.test endpoint
+type APIResponseFull struct {
+	SlackResponse
+	Args map[string]string `json:"args,omitempty"`
+}
+
+func apiRequest(ctx context.Context, client httpClient, path string, values url.Values, debug debug) (*APIResponseFull, error) {
+	response := &APIResponseFull{}
+	err := postSlackMethod(ctx, client, path, values, response, debug)
```
james-lawrence, line 17: ~~postSlackMethod's signature has changed. I suggest you rebase off the latest master.~~

misread sorry, change the parameter name to `d` please. =)

james-lawrence (changes requested):

api.go
```diff
@@ -0,0 +1,45 @@
+package slack
```
james-lawrence, line 1: i dislike this file name. too generic. move it into the file with the client definition if you're looking for somewhere to put it.

CodeMonk:
Yeah, between me posting this, and now, things have changed a LOT. Since this is live where I work, I'm going to close it until the next version comes out, then make sure it is congruent with the other routes.

PR #433, opened 2018-10-17
Added auth.revoke endpoint

CodeMonk:
Added auth.revoke endpoint so that an application can revoke it's oauth tokens.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -1,0 +1,43 @@
+package slack
+
+import (
+	"context"
+	"errors"
+	"net/url"
+)
+
+// AuthRevokeResponse contains our Auth response from the auth.revoke endpoint
+type AuthRevokeResponse struct {
+	SlackResponse      // Contains the "ok", and "Error", if any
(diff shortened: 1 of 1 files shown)


auth.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// authRequest sends the actual request, and unmarshals the response
+func authRequest(ctx context.Context, client httpClient, path string, values url.Values, d debug) (*AuthRevokeResponse, error) {
+	response := &AuthRevokeResponse{}
+	err := postSlackMethod(ctx, client, path, values, response, d)
+	if err != nil {
+		return nil, err
+	}
+	if !response.Ok {
```
james-lawrence: response has an error method for you to use.

james-lawrence (changes requested):
fix the error handling and its GTG.

CodeMonk:
Fixed the error handling per review comments.

james-lawrence (approved):

PR #437, opened 2018-10-30
Retrieve list of conversations for a given user

tauffredou:
Leaving the userId empty will get the list of conversations for the current user (or bot)


james-lawrence:
curious what lead to you making this change, but LGTM.

tauffredou:
@james-lawrence The documentation describes the user argument as optional. This lets few meanings for the empty value :

1.  all
2. none
3. implicit user

The first two values have very little sens, remains the third which means the "current user" in our context

As the argument is optional, it's safe to allow forwarding the empty user to the API without knowing what it exactly refers to.

james-lawrence:
all that makes sense, I was more commenting on if it was sending an empty string did slack do something weird instead of defaulting to the implicit user?

PR #438, opened 2018-10-31
Add support for unfurling links

seanhoughton:
Hello, I have added support for the `chat.unfurl` method.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -275,7 +275,7 @@ func MsgOptionUnfurl(timestamp string, unfurls map[string]Attachment) MsgOption 
 		if err == nil {
 			config.values.Add("unfurls", string(unfurlsStr))
 		}
-		return nil
+		return err
 	}
 }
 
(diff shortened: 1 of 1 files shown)


chat.go
```diff
@@ -260,6 +266,19 @@ func MsgOptionDelete(timestamp string) MsgOption {
...
+// MsgOptionUnfurl unfurls a message based on the timestamp.
+func MsgOptionUnfurl(timestamp string, unfurls map[string]Attachment) MsgOption {
+	return func(config *sendConfig) error {
+		config.endpoint = APIURL + string(chatUnfurl)
+		config.values.Add("ts", timestamp)
+		unfurlsStr, err := json.Marshal(unfurls)
+		if err == nil {
+			config.values.Add("unfurls", string(unfurlsStr))
+		}
+		return nil
```
james-lawrence: error should be returned in if occurs.

james-lawrence (changes requested):
error handling needs fixing.

seanhoughton:
Good catch, I've updated the code to return any error that might happen if the marshal fails.

james-lawrence (approved):

james-lawrence:
thanks!

PR #440, opened 2018-11-03
fixes #439

### LINKED ISSUE #439: Security issues in SecretsVerifier
Hi! I've noticed a couple of security issues in the implementation of `SecretsVerifier`. Specifically, there are two important ways in which it diverges from the recommendations set forth in the documentation here: https://api.slack.com/docs/verifying-requests-from-slack.

First, and most important: `SecretsVerifier` uses `==` to compare the request signature and the expected signature. This leaks timing information, as `==` short-circuits to `false` as soon as it encounters a non-matching character. It should instead use [`hmac.Equal`](https://golang.org/pkg/crypto/hmac/#Equal), which is specifically implemented to perform such a comparison without leaking timing information. The docs call this out specifically:

>Note: for best practice, use an hmac compare function instead of directly comparing the signatures for equality.

Additionally, `SecretsVerifier` does not check that the timestamp is recent. From the docs:

>While you're extracting the timestamp, check to make sure that the request occurred recently.

There's no officially recommended timeframe, but the example in the docs uses 5 minutes which seems reasonable.

I'd be happy to submit a PR with these improvements, if so desired.

P.S.: Not a security issue, but I think semantically it makes more sense to rename this struct to `SignatureVerifier`. We could drop in a type alias (e.g. `type SecretsVerifier = SignatureVerifier`) if preserving backwards compatibility is a concern.

james-lawrence:
- use hmac.Equal
- check timestamp

### DIFF
--- a/security.go
+++ b/security.go
@@ -8,28 +8,61 @@ import (
 	"fmt"
 	"hash"
 	"net/http"
+	"strconv"
+	"time"
+)
+
+// Signature headers
+const (
+	hSignature = "X-Slack-Signature"
+	hTimestamp = "X-Slack-Request-Timestamp"
 )
 
 // SecretsVerifier contains the information needed to verify that the request comes from Slack
 type SecretsVerifier struct {
-	slackSig  string
-	timeStamp string
+	signature string
 	hmac      hash.Hash
 }
 
-// NewSecretsVerifier returns a SecretsVerifier object in exchange for an http.Header object and signing secret
-func NewSecretsVerifier(header http.Header, signingSecret string) (SecretsVerifier, error) {
-	if header["X-Slack-Signature"][0] == "" || header["X-Slack-Request-Timestamp"][0] == "" {
-		return SecretsVerifier{}, errors.New("Headers are empty, cannot create SecretsVerifier")
+func unsafeSignatureVerifier(header http.Header, secret string) (_ SecretsVerifier, err error) {
+	signature := header.Get(hSignature)
+	stimestamp := header.Get(hTimestamp)
+
+	if signature == "" || stimestamp == "" {
+		return SecretsVerifier{}, errors.New("missing headers")
 	}
 
-	hash := hmac.New(sha256.New, []byte(signingSecret))
-	hash.Write([]byte(fmt.Sprintf("v0:%s:", header["X-Slack-Request-Timestamp"][0])))
+	hash := hmac.New(sha256.New, []byte(secret))
+	hash.Write([]byte(fmt.Sprintf("v0:%s:", stimestamp)))
+
 	return SecretsVerifier{
-		slackSig:  header["X-Slack-Signature"][0],
-		timeStamp: header["X-Slack-Request-Timestamp"][0],
+		signature: signature,
 		hmac:      hash,
 	}, nil
+}
+
+// NewSecretsVerifier returns a SecretsVerifier object in exchange for an http.Header object and signing secret
+func NewSecretsVerifier(header http.Header, secret string) (sv SecretsVerifier, err error) {
+	var (
+		timestamp int64
+	)
+
+	stimestamp := header.Get(hTimestamp)
+
+	if sv, err = unsafeSignatureVerifier(header, secret); err != nil {
+		return sv, err
+	}
+
+	if timestamp, err = strconv.ParseInt(stimestamp, 10, 64); err != nil {
+		return SecretsVerifier{}, err
+	}
+
+	diff := absDuration(time.Now().Sub(time.Unix(timestamp, 0)))
+	if diff > 5*time.Minute {
+		return sv, fmt.Errorf("timestamp is too old")
+	}
+
+	return sv, err
 }
 
 func (v *SecretsVerifier) Write(body []byte) (n int, err error) {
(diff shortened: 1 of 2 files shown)


security.go
```diff
@@ -1,47 +1,91 @@
...
+		return sv, err
+	}
+
+	if timestamp, err = strconv.ParseInt(stimestamp, 10, 64); err != nil {
+		return SecretsVerifier{}, err
+	}
+
+	diff := absDuration(time.Now().Sub(time.Unix(timestamp, 0)))
+	if diff > 5*time.Minute {
+		return sv, fmt.Errorf("timestamp is too old")
```
adamrothman: I'm concerned that this makes it too easy to do the wrong thing. The `SecretsVerifier` returned here will still work, so if the caller doesn't check the error, they may accept a request whose timestamp is too old.

I think the best solution is to move the timestamp parsing/check to `Ensure()`. This simplifies things by letting you get rid of `unsafeSignatureVerifier()`. Alternatively you could change this to return an empty `SecretsVerifier`, but that doesn't feel as nice.

security.go
```diff
@@ -1,47 +1,91 @@
...
+	if hmac.Equal([]byte(computed), []byte(v.signature)) {
+		return nil
+	}
+
+	return fmt.Errorf("Expected signing signature: %s, but computed: %s", v.signature, computed)
+}
+
+func abs64(n int64) int64 {
+	y := n >> 63
+	return (n ^ y) - y
```
adamrothman, line 94: Neat!

security.go
```diff
@@ -1,47 +1,91 @@
...
+func (v *SecretsVerifier) Write(body []byte) (n int, err error) {
+	return v.hmac.Write(body)
+}
+
+// Ensure compares the signature sent from Slack with the actual computed hash to judge validity
+func (v SecretsVerifier) Ensure() error {
+	computed := "v0=" + string(hex.EncodeToString(v.hmac.Sum(nil)))
+
+	// use hmac.Equal prevent leaking timing information.
+	if hmac.Equal([]byte(computed), []byte(v.signature)) {
```
adamrothman: This seems overly-complicated. It computes the HMAC (`[]bytes`), hex encodes that to a `string`, then converts that back to `[]bytes` (along with the header signature) for comparison.

IMO it's more elegant to decode the header signature (removing the `"v0="` prefix) with [`hex.DecodeString()`](https://golang.org/pkg/encoding/hex/#DecodeString) and pass that and the computed HMAC directly to `hmac.Equal()`.

adamrothman (changes requested):
Just a few minor things – looks good overall.

security.go
james-lawrence: haha good catch wasn't paying attention to the conversions.

security.go
james-lawrence: I had the same thought, not going to worry about it for now. returning blank SecretVerifier is fine for now.

I was thinking about getting rid of all the error checks in the new method and handling them all in ensure. but decided to pass on that for now. this PR is for the vulnerabilities not the cleaniness =)

jimsmart: This code also accepts timestamps with _future_ times in the next 5 mins (due to use of abs).

— Is this intended behaviour?

adamrothman: @jimsmart Yes, it's reasonable to allow for clock skew in either direction. The documentation uses abs as well: https://api.slack.com/docs/verifying-requests-from-slack

james-lawrence: @jimsmart as adam said, either direction is fine as you don't which clock is out of sync in which direction. =)

security.go
yordis: @adamrothman this is too clever for me to understand, why we need this `absDuration` and `abs64`?

Reading Slack docs, they use

```
if absolute_value(time.time() - timestamp) > 60 * 5:
  ...
```

Where I have no clue what they mean by `absolute_value` function call since I googled it in Python land but didn't get any results back, probably there is some computer science term that I am missing.

So I will appreciate if you could explain me these two functions in this context 🙏

adamrothman: The absolute value of a number, formally denoted as `|x|`, is defined as its magnitude or distance from zero. That is to say:
* `|x| = x when x > 0`
* `|x| = -x when x < 0`

In this context, the absolute value of the difference between the local system time and the request timestamp is used to allow for clock skew in either direction.

This particular implementation uses [two's complement](https://en.wikipedia.org/wiki/Two%27s_complement) arithmetic. It comes from [chapter 2 of _Hacker's Delight_ by Henry S. Warren](https://books.google.com.au/books?id=VicPJYM0I5QC&lpg=PA18&ots=2o-SROAuXq&dq=hackers%20delight%20absolute&pg=PA18#v=onepage&q=hackers%20delight%20absolute&f=false), by way of Ryan Armstrong's blog: [Optimized abs() for int64 in Go](http://cavaliercoder.com/blog/optimized-abs-for-int64-in-go.html).

The `absDuration` function just handles the type conversions to make everything a little more ergonomic.

yordis: > The absolute value of a number, formally denoted as |x|, 

Ahhhhh that absolute value 🤦‍♂ gosh I am stupid, I thought was something else 😄

Is that `absolute_value` a real function in Python thou? It will be really really helpful if the docs actually use valid Python functions `abs(...)`.

Thanks a lot for the explanation, appreciated!

PR #442, opened 2018-11-07
add support for pin events

chilicheech:
## Objective
We want to be able to handle `pin_added` and `pin_removed` events.
## Context
We want to use this library to handle pin events. When an item is pinned or unpinned in slack we want our app to handle those events.
Most of the new structs in this file are copied pasted from the slack package. The only new is `ItemMessage`. We had to use that name because `Message` is already declared as a constant.

It would be really awesome if we can put structs that are shared between the `slack` and `slackevents` package in a package they both import so we reduce code duplication. But I believe that can be accomplished in a separate PR.

### DIFF
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -98,6 +98,33 @@ type MemberJoinedChannelEvent struct {
 	Inviter     string `json:"inviter"`
 }
 
+type pinEvent struct {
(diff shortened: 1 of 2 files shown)


james-lawrence:
thanks, looks great.

james-lawrence:
as for sharing the structs, maybe someday. not opposed to it.

Issue #403, opened 2018-09-16
Client should support batched presence updates


theckman:
Slack made a change in the API last year that would cause `presence_change` events to no longer be sent over the RTM connection if you're using the original format. Instead, clients needed to opt-in to the new batched `presence_change` events (as it's a breaking API change) if they wanted to subscribe to presence change notifications of users:

- https://api.slack.com/docs/presence-and-status#batching

To accomplish this we'll need to update the `PresenceChangeEvent`'s `User` field to be an `[]string`. We can then pass the new option when making the `rtm.connect` API call.

suekto-andreas:
hi @theckman, @james-lawrence  - created a PR for this enhancement, kindly browse in: https://github.com/nlopes/slack/pull/444

PR #444, opened 2018-11-10
 #403 Handle batch_presence_aware

suekto-andreas:
This is in relation to issue #403 

In order to support this:
1. batch_presence_aware is treated as a connection parameter
2. introduce a connParams map to host parameter to embed into the generated connection URL
3. example of usage is given in example/connparams/connparams.go

Kindly help to review, please feel free to feedback. Thanks !

### DIFF
--- a/examples/connparams/connparams.go
+++ b/examples/connparams/connparams.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/nlopes/slack"
+)
+
+func main() {
+	api := slack.New(
+		"YOUR TOKEN HERE",
+		slack.OptionDebug(true),
+		slack.OptionLog(log.New(os.Stdout, "slack-bot: ", log.Lshortfile|log.LstdFlags)),
+	)
+
+	// turn on the batch_presence_aware option
+	rtm := api.NewRTM(slack.RTMOptionConnParams(map[string]string{
+		"batch_presence_aware": "1",
+	}))
+	go rtm.ManageConnection()
+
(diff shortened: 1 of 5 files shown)


examples/connparams/connparams.go
```diff
@@ -0,0 +1,62 @@
...
+
+func main() {
+	api := slack.New(
+		"YOUR TOKEN HERE",
+		slack.OptionDebug(true),
+		slack.OptionLog(log.New(os.Stdout, "slack-bot: ", log.Lshortfile|log.LstdFlags)),
+	)
+
+	// turn on the batch_presence_aware option
+	rtm := api.NewRTM(slack.RTMOptionConnParams(map[string]string{
```
james-lawrence: I believe you want [url.Values](https://golang.org/pkg/net/url/#Values) here.

suekto-andreas: Thanks! yep that express it better - updated.

james-lawrence:
LGTM, tyvm.

ramikalai:
Does this actually work? The docs seem to say that this param needs to be passed to rtm.connect (this is the function which returns the websocket URL)... I don't believe this is the case currently and Im not receiving any presence events

suekto-andreas:
@ramikalai  I have been not using this for a while, unless there is Slack API changes, it was working. Could you help to share how to replicate your finding ? (i.e. snippet of code etc.)

Probably a bug ticket is a better approach to communicate, complete with replication step.

ramikalai:
Hey @suekto-andreas, I was doing something silly! Got it to work using the current implementation. Thank you!

PR #445, opened 2018-11-14
Fix #429 Removed attachment field from PostMessageParameters

### LINKED ISSUE #429: MsgOptionPostMessageParameters doesn't add attachments
`MsgOptionPostMessageParamters` looks like it was added to support backwards compatibility with users of `PostMessageParameters`. However it appears that attachments are not handled by the options function.

Here's a test case that fails:

```golang
import (
	"net/url"
	"strings"
	"testing"

	"github.com/go-test/deep"
	"github.com/nlopes/slack"
)

func TestConvertMessageParams(t *testing.T) {

	params := &PostMessageParameters{
		Mrkdwn:   true,
		Username: "Test User",
		Attachments: []*Attachment{
			{
				Title:     "Google",
				TitleLink: "http://google.com",
			},
		},
	}

	opts := ConvertMsgParams(params)
	actualURL, actualParams, err := slack.UnsafeApplyMsgOptions("token", "channel", opts)

	if err != nil {
		t.Errorf("Unexpected error: %+v", err)
	}

	expectedURL := "https://slack.com/api/chat.postMessage"
	if actualURL != expectedURL {
		t.Errorf("Expected URL '%s', was '%s'", expectedURL, actualURL)
	}

	expectedParams, _ := url.ParseQuery("attachments=%5B%7B%22fallback%22%3A%22%22%2C%22title%22%3A%22Google%22%2C%22title_link%22%3A%22http%3A%2F%2Fgoogle.com%22%2C%22text%22%3A%22%22%7D%5D&channel=channel&token=token&unfurl_media=false&username=Test+User")
	if diff := deep.Equal(expectedParams, actualParams); diff != nil {
		t.Errorf("unexpected params:\n%+v", strings.Join(diff, "\n"))
	}

}
```

A work around is to add copy the code in `MsgOptionAttachments` into `MsgOptionPostMessageParamters`.

is2ei:
Fixed https://github.com/nlopes/slack/issues/429

I removed attachment field from PostMessageParameters.
Could you review it, please?

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -43,19 +43,18 @@ func (c chatResponseFull) getMessageTimestamp() string {
 
 // PostMessageParameters contains all the parameters necessary (including the optional ones) for a PostMessage() request
 type PostMessageParameters struct {
-	Username        string       `json:"username"`
-	AsUser          bool         `json:"as_user"`
-	Parse           string       `json:"parse"`
-	ThreadTimestamp string       `json:"thread_ts"`
-	ReplyBroadcast  bool         `json:"reply_broadcast"`
-	LinkNames       int          `json:"link_names"`
-	Attachments     []Attachment `json:"attachments"`
-	UnfurlLinks     bool         `json:"unfurl_links"`
-	UnfurlMedia     bool         `json:"unfurl_media"`
-	IconURL         string       `json:"icon_url"`
-	IconEmoji       string       `json:"icon_emoji"`
-	Markdown        bool         `json:"mrkdwn,omitempty"`
-	EscapeText      bool         `json:"escape_text"`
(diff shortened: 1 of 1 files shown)


james-lawrence:
Looks good. mind confirming that only send message uses this structure? if there are other areas then might want to hold off until we look at those.

is2ei:
@nlopes  Thanks. You mean you worry if `PostMessageParameters` is used by other than posting messages? I checked it by searching keyword `PostMessageParameters` and it seems `PostMessage` uses`PostMessageParameters`. Please Let me know if you think I didn't get correctly what you pointed out .

james-lawrence:
nope sounds like you did. then we're good.

PR #455, opened 2018-12-26
Added include_locale in Web APIs

tomodian:
Hi, thank you for building such a great package!
Not sure this is handled by other PRs, but I added `include_locale` in several APIs so that Slack returns locale by default.

According to [Slack doc](https://api.slack.com/changelog/2017-09-locale-locale-locale), the locale is opt-in and doesn't seem to have any side-effect, therefore using it as default is reasonable.

Looking forward for your comments!


james-lawrence:
would prefer this being a provided option but fine as is. Thanks.

tomodian:
@james-lawrence thanks for the merge!

PR #457, opened 2019-01-02
Error when user forgets to set the Filename when using a Reader for file uploads

nubunto:
I just got bit by this issue and took me a while of investigating to get to the bottom. See [this gist](https://gist.github.com/nubunto/030bead224f1829c60149b0b04d2c19e) for more information on how to reproduce.

This PR adds a small piece of documentation to the FileUploadParameters type, specifying that you'll get errors when using the Reader and not providing the Filename.

It also enforces that semantic in the `UploadFileContext` method, by checking when there is a provided Reader, the Filename option is also set. It also adds a test for this behaviour.

### DIFF
--- a/files.go
+++ b/files.go
@@ -93,6 +93,8 @@ type File struct {
 //
 // There are three ways to upload a file. You can either set Content if file is small, set Reader if file is large,
 // or provide a local file path in File to upload it from your filesystem.
+//
+// Note that when using the Reader option, you *must* specify the Filename, otherwise the Slack API isn't happy.
 type FileUploadParameters struct {
 	File            string
 	Content         string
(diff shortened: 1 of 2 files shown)


files.go
```diff
@@ -248,6 +251,9 @@ func (api *Client) UploadFileContext(ctx context.Context, params FileUploadParam
 	} else if params.File != "" {
 		err = postLocalWithMultipartResponse(ctx, api.httpclient, "files.upload", params.File, "file", values, response, api)
 	} else if params.Reader != nil {
+		if params.Filename == "" {
```
james-lawrence: I'd just move this to the top and return it immediately. otherwise PR looks good.

nubunto: OK! I'll commit the changes.

james-lawrence:
thanks, sorry for the delay.

PR #462, opened 2019-01-08
Add Reminders API

jamiehannaford:
This PR adds support for the Reminders API, as documented [here](https://api.slack.com/methods/reminders.add).

It builds on @phinze's work in https://github.com/nlopes/slack/pull/254 but updates it to work with recent changes. LMK if anything is missing or changes are needed!


reminders.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return api.doReminder(context.Background(), "reminders.add", values)
+}
+
+// AddUserReminder adds a reminder for a user.
+//
+// See https://api.slack.com/methods/reminders.add (NOTE: the ability to set
+// reminders on a channel is currently undocumented but has been tested to
+// work)
+func (api *Client) AddUserReminder(userID, text, time string) (*Reminder, error) {
```
james-lawrence, line 52: time as a string is a little unclear what is the expected format?

reminders.go
```diff
@@ -0,0 +1,82 @@
...
+	SlackResponse
+	Reminder Reminder `json:"reminder"`
+}
+
+func (api *Client) doReminder(ctx context.Context, path string, values url.Values) (*Reminder, error) {
+	response := &reminderResp{}
+	if err := postSlackMethod(ctx, api.httpclient, path, values, response, api); err != nil {
+		return nil, err
+	}
+	if !response.Ok {
```
james-lawrence: response.Err() see SlackResponse.

Gopkg.lock
```diff
@@ -2,38 +2,52 @@
...
   packages = ["assert"]
+  pruneopts = "UT"
   revision = "f35b8ab0b5a2cef36673838d662e249dd9c94686"
   version = "v1.2.2"
 
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "596fa546322c2a1e9708a10c9f39aca2e04792b477fab86fb2899fbaab776070"
+  input-imports = [
```
james-lawrence: why did this file change?

james-lawrence (changes requested):
fix up the error handling, and don't commit the lock file and i'll accept the changes.

though I do find the use of a string for time extremely odd. but not familiar enough with the api to worry about it for now.

reminders.go
jamiehannaford: It can be more than just a timestamp. According to [the docs](https://api.slack.com/methods/reminders.add) it is:

> When this reminder should happen: the Unix timestamp (up to five years from now), the number of seconds until the reminder (if within 24 hours), or a natural language description (Ex. "in 15 minutes," or "every Thursday")

So a string seems most fitting IMO.

Gopkg.lock
jamiehannaford: This was accidental, I'll revert.

jamiehannaford:
@james-lawrence Thanks for the review, I've addressed all feedback.

reminders.go
```diff
@@ -0,0 +1,79 @@
...
+func (api *Client) DeleteReminder(id string) error {
+	values := url.Values{
+		"token":    {api.token},
+		"reminder": {id},
+	}
+	response := &SlackResponse{}
+	if err := postSlackMethod(context.Background(), api.httpclient, "reminders.delete", values, response, api); err != nil {
+		return err
+	}
+	if !response.Ok {
```
james-lawrence: sorry wasn't explicit about this one as well.
`response.Err()`

james-lawrence (approved):

james-lawrence:
fix the build and I'll merge. (you didn't remove an unused import)

PR #463, opened 2019-01-09
Update file interface - add shares attribute

wyam:
We use your library in Go to create a Slack's app and play with files but we find out that the File struct is not up to date from Slack API (https://api.slack.com/methods/files.info). 
So we just add the shares struct.

### DIFF
--- a/files.go
+++ b/files.go
@@ -94,7 +94,7 @@ type share struct {
 }
 
 type shareFileInfo struct {
-	ReplyUsers      string `json:"reply_users"`
+	ReplyUsers      []string `json:"reply_users"`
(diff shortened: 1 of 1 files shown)


files.go
```diff
@@ -86,6 +86,22 @@ type File struct {
 	CommentsCount   int      `json:"comments_count"`
 	NumStars        int      `json:"num_stars"`
 	IsStarred       bool     `json:"is_starred"`
+	Shares          share    `json:"shares"`
+}
+
+type share struct {
```
james-lawrence: this should be public. same with ShareFileInfo.

james-lawrence (changes requested):

james-lawrence (approved):

PR #465, opened 2019-01-14
Add previous message to slack message struct on edited messages

alexbrazier:
If a message has been edited then Slack will send the previous message back when using RTM.
This PR just adds the previous message to the `Message` struct if it exists.

I've tested this using my own fork and it works as expected.


james-lawrence:
thanks, sorry for the delay.

PR #466, opened 2019-01-15
Add dialog_cancellation interaction type.

Weyland:
While using the library I noticed that there was an InteractionType missing called `dialog_cancellation `.


james-lawrence:
thanks LGTM.

PR #468, opened 2019-01-23
Add MsgOptions for icon_url and icon_emoji

ereyes01:
The icon URL and emoji were the only 2 fields that can be set in `MsgOptionPostMessageParameters()` for backward compatibility, but not exposed as individual `MsgOption`'s


james-lawrence:
thanks.

PR #469, opened 2019-01-23
Added apps.uninstall method

CodeMonk:
It's useful for removing your app from an instance. When it is called,
    you should receive an AppUninstalledEvent, if you're checking for
    such things.

### DIFF
--- a/apps.go
+++ b/apps.go
@@ -1,0 +1,44 @@
+package slack
+
+import (
+	"context"
+	"net/url"
+)
+
+// AppsUninstallResponse contains our Auth response from the auth.revoke endpoint
+type AppsUninstallResponse struct {
+	SlackResponse // Contains the "ok", and "Error", if any
+}
+
+// appsRequest sends the actual request, and unmarshals the response
+func appsRequest(ctx context.Context, client httpClient, path string, values url.Values, d debug) (*AppsUninstallResponse, error) {
+	response := &AppsUninstallResponse{}
+	err := postSlackMethod(ctx, client, path, values, response, d)
+	if err != nil {
+		return nil, err
+	}
+
+	return response, response.Err()
+}
+
+// SendAppsUninstall will send a revocation for our token. We are re-supplying the token here, but, if
+// "" is sent instead, it will use the token stored in New()
+func (api *Client) SendAppsUninstall(clientID, clientSecret, token string) (*AppsUninstallResponse, error) {
+	return api.SendAppsUninstallContext(context.Background(), clientID, clientSecret, token)
+}
+
+// SendAppsUninstallContext will retrieve the satus from api.test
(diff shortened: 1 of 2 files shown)


apps.go
```diff
@@ -0,0 +1,44 @@
+package slack
+
+import (
+	"context"
+	"net/url"
+)
+
+// AppsUninstallResponse contains our Auth response from the auth.revoke endpoint
+type AppsUninstallResponse struct {
```
james-lawrence, line 9: seems unnecessary..... only thing of value returned by these methods is an error?

CodeMonk: Or success . . . Not sure if you're proposing a change . . .

james-lawrence: I'm proposing you remove this type altogether and only return an error. this type doesn't provide any additional value to the public api.

CodeMonk: But, it's holding the ACTUAL response from slack. I modeled it off of other functions, but, obviously picked the wrong one.

Could you please point me to another slack endpoint that returns like you're suggesting? I'm happy to make any changes, but, I'm not sure exactly what you're suggesting . . . . particularly since almost ALL api endpoints return SlackResponse + any custom data, and this particular endpoint has no custom data to return.

james-lawrence: > But, it's holding the ACTUAL response from slack.

so what? the actual response provides no new information. other methods have actual data to return this one doesn't. (also no it doesn't hold the actual response, it holds a representation of the response.)

nothing in the SlackResponse contains any new or useful information that the error (or lack thereof) will contain.

>Could you please point me to another slack endpoint that returns like you're suggesting?

Yes actually I can, but I'm not terribly inclined to do so. why? because said functions existence is completely immaterial. following a pattern because its a pattern only makes sense when the conditions for using the pattern are the same.

your argument that every function does this is not convincing for why this particular function should do it. nor is your claim that every function does it correct.

ZeeStorm:
Just a bystander here that needed this functionality-- but I noticed the API documentation (https://api.slack.com/methods/apps.uninstall) recommends the preferred HTTP method is GET and I believe this PR is using `postSlackMethod` which fires a POST instead.

james-lawrence:
that is odd.... given it changes state... but good catch.

CodeMonk:
Weird - I'm actually using this code in production, so I know that the POST works. I'll test if a GET also works -- documentation could be wrong.

james-lawrence:
i suspect the documentation is wrong tbh.

ZeeStorm:
I reported an issue with the usage of it to support, and apparently the documentation is "very out of date". The feature has completely removed support for user tokens (`xoxp`) and will only support workspace tokens (`xoxa`), which makes the use case a lot more difficult for bot-based usage. They have asked the API team to update the documentation for the method. Anyways, just figured I'd report my findings.

PR #471, opened 2019-02-10
Fix logging for expected and computed signatures

alyosha:
# Summary
While working on a personal project I noticed that some logging in the `security.go` file wasn't outputting properly. I suppose alternatively we could edit this statement altogether, but logging output looks strange in it's current state. 

## Example current output
`Expected signing signature: 1C_V^K^t-lƕ*", but computed: ^OԬNf3/9Knn1`

### DIFF
--- a/security.go
+++ b/security.go
@@ -88,7 +88,7 @@ func (v SecretsVerifier) Ensure() error {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


james-lawrence:
guess it depends on your definition of strange..... =)

PR #292, opened 2018-04-02
Download slack files using API

jmswenski:
Adding download functionality to the client with a new method GetFile
which will write the contents to a provided io.Writer

Resolves: #141

PR #472, opened 2019-02-10
Add get file method

alyosha:
# Summary
Decided to make a new PR since #292 was stale, needed a rebase, and because I made a few different implementation choices. Would love to see this method available for a number of personal projects I'm working on. 

Went ahead and addressed your feedback from the previous PR @james-lawrence and added a test, addressed the segfault (plus one additional case), and also split out the status code error handling into its own helper method. 

I am not a hundred percent confident with the test, so would appreciate any feedback/thought you might have.

Happy to make any changes required 🙇

### DIFF
--- a/files.go
+++ b/files.go
@@ -187,6 +187,11 @@ func (api *Client) GetFileInfoContext(ctx context.Context, fileID string, count,
 	return &response.File, response.Comments, &response.Paging, nil
 }
 
+// GetFile retreives a given file from its private download URL
+func (api *Client) GetFile(file *File, writer io.Writer) error {
+	return downloadFile(api.httpclient, api.token, file.URLPrivateDownload, writer, api)
+}
+
 // GetFiles retrieves all files according to the parameters given
 func (api *Client) GetFiles(params GetFilesParameters) ([]File, *Paging, error) {
 	return api.GetFilesContext(context.Background(), params)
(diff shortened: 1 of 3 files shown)


alyosha:
@james-lawrence any chance at getting a review? would love to use this in an upcoming proj

files.go
```diff
@@ -187,6 +187,11 @@ func (api *Client) GetFileInfoContext(ctx context.Context, fileID string, count,
 	return &response.File, response.Comments, &response.Paging, nil
 }
 
+// GetFile retreives a given file from its private download URL
+func (api *Client) GetFile(file *File, writer io.Writer) error {
```
james-lawrence: instead of passing in the file object, just pass in the string url.

james-lawrence (changes requested):
minor change for the api.

as for the test.... i mean it doesn't really test anything except that initial if statement. and that the request received a response. but its fine.

alyosha:
yeah, I didn't think there was much else to test since it's mostly dependent on the response from the http req. happy to add any cases you can think but in the meantime made the API fix you suggested 🙏 thanks for the quick response

james-lawrence (approved):
LGTM.

james-lawrence:
just need to fix your test =)

alyosha:
haha I realized it right as I stepped in the shower... thanks so much for the review!

PR #473, opened 2019-02-15
Add enterprise node to user

jjjjeeffff:
With Enterprise Grid, users can have not only a workspace ID but also a global ID. The global ID is needed sometimes when doing things like inviting users to channels via the API. This just adds the missing node that is present for Enterprise Grid users.

[Docs](https://api.slack.com/types/user#enterprise_grid_user_objects)


james-lawrence:
LGTM.

PR #474, opened 2019-02-15
Get Team Custom Fields, Set User Custom Fields

MattDavisRV:
Added functionality to retrieve defined custom fields from the the Team.  

Added ability to update User custom fields from the information available from the new team request.

### DIFF
--- a/users.go
+++ b/users.go
@@ -507,7 +507,7 @@ func (api *Client) SetUserCustomFieldsContext(ctx context.Context, userID string
 	}
 
 	if !response.Ok {
-		return errors.New(response.Error)
+		return response.Err()
 	}
 
 	return nil
--- a/users.go
+++ b/users.go
@@ -472,7 +472,7 @@ func (api *Client) SetUserCustomFields(userID string, customFields map[string]Us
 	return api.SetUserCustomFieldsContext(context.Background(), userID, customFields)
 }
 
-// SetUserCustomFieldsContext will set a custom status and emoji for the currently authenticated user with a custom context
+// SetUserCustomFieldsContext will set a users custom profile field with context.
 //
 // For more information see SetUserCustomFields
 func (api *Client) SetUserCustomFieldsContext(ctx context.Context, userID string, customFields map[string]UserProfileCustomField) error {
--- a/users.go
+++ b/users.go
@@ -464,6 +464,56 @@ func (api *Client) DeleteUserPhotoContext(ctx context.Context) error {
 	return response.Err()
 }
 
+// SetUserCustomFields sets Custom Profile fields on the provided users account.  Due to the non-repeating elements
+// within the request, a map fields is required.  The key in the map signifies the field that will be updated.
+//
+// See GetTeamProfile for information to retrieve possible fields for your account.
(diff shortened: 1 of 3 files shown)


MattDavisRV:
Slack separates Profile and Custom Field update requests.  This was needed to update additional fields that were not apart of the main profile request.  To issue the update request, slack doesn't use a conventional request structure, requiring the need for the map[string]UserCustomProfile. 

Requesting the team custom variables was added to know which IDs to send in the above request. 

Open to suggestions and/or updates to this PR.

users.go
```diff
@@ -464,6 +464,56 @@ func (api *Client) DeleteUserPhotoContext(ctx context.Context) error {
...
 
+// SetUserCustomFields sets Custom Profile fields on the provided users account.  Due to the non-repeating elements
+// within the request, a map fields is required.  The key in the map signifies the field that will be updated.
+//
+// See GetTeamProfile for information to retrieve possible fields for your account.
+func (api *Client) SetUserCustomFields(userID string, customFields map[string]UserProfileCustomField) error {
+	return api.SetUserCustomFieldsContext(context.Background(), userID, customFields)
+}
+
+// SetUserCustomFieldsContext will set a custom status and emoji for the currently authenticated user with a custom context
```
olivierlemasle: Comment is wrong (copy-paste error) ;-)

olivierlemasle (approved):
:+1:

users.go
MattDavisRV: Good catch!  Should be fine now

james-lawrence:
TeamProfile.... not familiar with this part of slacks api. does it represent customization values for the team itself? not the users?

MattDavisRV:
TeamProfile defines custom fields that belong to the team, but attributed to the user.  They're represented as a ID/Value pair, which can only be retrieved from the added code. 

As an example (in my use case), I've added a custom field called employeeId to the team, then added this attribute to each user to properly identify them during interactive messages (dialogs, etc).

users.go
```diff
@@ -464,6 +464,56 @@ func (api *Client) DeleteUserPhotoContext(ctx context.Context) error {
...
+		"user":    {userID},
+		"profile": {string(profile)},
+	}
+
+	response := &userResponseFull{}
+	if err := postForm(ctx, api.httpclient, APIURL+"users.profile.set", values, response, api); err != nil {
+		return err
+	}
+
+	if !response.Ok {
```
james-lawrence, line 509: return response.Err()

james-lawrence (changes requested):
one minor comment.

users.go
james-lawrence, line 509: sorry if i wasnt clear. response.Err() does `if !response.Ok {}`

you can literally just say `reutnr response.Err()` here

olivierlemasle:
@MattDavisRV You removed the branch `SetUserCustomFields` from your fork repo. Was it on purpose?

MattDavisRV:
@olivierlemasle no, it was not on purpose.  Not really sure what happened there.  I've been busy with work.  I'll try to reimplement since it was a small change along with the change requested here.

MattDavisRV:
I'm waiting for #496 to be completed before I swing back to this PR.  I will complete the requested changes then.

PR #478, opened 2019-02-20
Add options to GetUsers

rssathe:


rssathe:
@james-lawrence added this because options aren't trickled through in this call to the pagination. For large workspaces with many users this can trigger rate limits if capped at 200 per request. We typically work with very large workspaces and thus need to set this to > 200

conversation.go
```diff
@@ -109,6 +109,29 @@ func (api *Client) GetUsersInConversationContext(ctx context.Context, params *Ge
 	return response.Members, response.ResponseMetaData.NextCursor, nil
 }
 
+// GetAllConversationsForUser gets all conversations for a given user
+func (api *Client) GetAllConversationsForUser(
```
james-lawrence, line 115: not too keen on including these functions in the slack package. Only reason the user version exists is because it existed prior to implementing pagination for users.

I'd be happy to include them in the slackutilsx package though.

james-lawrence (changes requested):
happy to include the GetUsers changes, though technically that method is deprecated in favor of GetUsersPaginated. mainly because no API breakage.

as for the other two methods, I'd prefer a cleanup of the existing method (even if it breaks API compat) to make pagination easier rather than supporting two new methods.

james-lawrence:
as an aside the real fix is the implement handling the rate limit within the GetUsers method, increasing the result size doesn't solve the underlying issue your running into.

james-lawrence:
and to be explicit:

lets split this into two PRs one for the get user changes, because those are easy to accept.

and one for the two new functions to discuss how to handle convenience utility functions for paginated endpoints.

rssathe:
@james-lawrence apologies I had only intended to submit the GetUsers change here but I commited the extra stuff to master because we use our fork in production. In a sense I'm glad you got a chance to review anyways. I'll submit two seperate prs.

I'm not sure what the conclusion is for the other two methods as you left many comments. Are you saying I should keep the logic as is and move to slackutilsx or modify the logic in the existing methods

rssathe:
> as an aside the real fix is the implement handling the rate limit within the GetUsers method, increasing the result size doesn't solve the underlying issue your running into.

This is true but more than anything at least the optionality to increase the result size is nice rather than letting that be preset. For the most part large workspaces don't trigger rate limits if you can set the limit to the max amounts, while the default of 200 definitely will. For other endpoints, I have built in a timer based throttler using time.Ticker. Happy to include some of that in here as well if you think its appropriate. My hunch was that that was better left to the end user of this client to implement

james-lawrence:
>I'm not sure what the conclusion is for the other two methods as you left many comments. Are you saying I should keep the logic as is and move to slackutilsx or modify the logic in the existing methods

I'm actually saying I'd prefer both. ideally the core slack package should only return pagination objects like GetUsersPaginated does.

then for methods like GetUsers which loads everything into memory before returning should be in slackutilsx, they are 3 line helper functions for convenience. but really not supply any new capabilities to the core library.


>or other endpoints, I have built in a timer based throttler using time.Ticker.

I wouldn't worry about it for this PR, slack returns how long you should wait so really you just need to sleep and retry when encountered.

PR #481, opened 2019-02-26
Add option to GetConversationsForUserParameters

PeterFeicht:
This adds an option to GetConversationsForUserParameters to exclude archived channels when getting the channels a user is part of.

### DIFF
--- a/conversation.go
+++ b/conversation.go
@@ -67,10 +67,11 @@ type GetUsersInConversationParameters struct {
 }
 
 type GetConversationsForUserParameters struct {
-	UserID string
-	Cursor string
-	Types  []string
-	Limit  int
+	UserID          string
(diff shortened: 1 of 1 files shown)


PeterFeicht:
Is there something I should do to address those failed checks? I can't make sense of the metalinter output...

james-lawrence:
short answer is yes, there is something you should do about it. Not sure what though havent seen that error before.

james-lawrence:
if someone wants to pick this up and figure out the build error. feel free.

PeterFeicht:
Oh, I completely forgot about this. Turns out it was wrong alignment of the struct field types caused by adding the new field.

james-lawrence:
LGTM.

PR #483, opened 2019-02-26
Add webhook runtime customization attributes

uthng:
Add some runtime customization attributes to override some informations of legacy Incoming Webhook such as: username, icon emoji, icon url and channel (cf. https://api.slack.com/custom-integrations/incoming-webhooks)


james-lawrence:
thanks.

PR #484, opened 2019-02-27
Add webhooks example

jnst:
I added the example code of [Incoming Webhooks](https://api.slack.com/incoming-webhooks) that we can most easily try the Slack API.


james-lawrence:
thanks.

PR #486, opened 2019-02-27
Support `thread_ts` parameter for Incoming Webhook

akshaylb:
Support `thread_ts` parameter for replying on existing threads via Incoming Webhooks.
Verification for API support can be checked [here](https://api.slack.com/incoming-webhooks).


james-lawrence:
thanks, sorry for the delay.

PR #487, opened 2019-03-03
Add Status expiration field in SetUserCustomStatus

### LINKED ISSUE #482: No ability to set a "User Custom Status" expiration
From what I saw, there's no way we can set an expiration time when creating a custom status set  but we can directly through the Slack API (https://api.slack.com/docs/presence-and-status).

I made a tiny change on the SetUserCustomStatus function on a fork, I'll open a PR later when I'll write/update some test.

juanwolf:
This PR closes #482.  You can set an expiration timestamp to a status. I've been using it [here](https://github.com/juanwolf/gomodoro/blob/master/pkg/outputs/slack.go#L37) for few weeks now and it works pretty well


james-lawrence:
LGTM, though this function might need some reworks in the future if it keeps growing.

PR #480, opened 2019-02-23
Slack Blocks

MattDavisRV:
Initial Implementation of Blocks provided by slack. Please refer to the provided examples and README within the example directory for additional information and usage.  

@james-lawrence open to discussion and changes based on this implementation.

PR #489, opened 2019-03-16
Fix misconfigured package name used in examples/blocks/blocks.go

0gajun:
`examples/blocks/blocks.go` was introduced in #480, but it doesn't use `nlopes/slack` package, but the PR's author's package.

I think that it is better to use the upstream repository, `nlpopes/slack`, to avoid confusion when running examples.


james-lawrence:
oh god. haha. thanks

PR #490, opened 2019-03-17
Block related interface changes/utility additions

alyosha:
# Summary 
Started messing around with blocks on some pet projects and there were a few changes I thought might be helpful. Have tested locally, but just started using blocks very recently so please feel free to let me know if I'm missing anything. 

### Changes
I split each of the three changes in this PR into separate commits:
1. **Exporting the block/blockElement interfaces**
The main reason I did this was so that I could use an array of elements/blocks with variadic functions such as `NewActionBlock` and `MsgOptionBlocks` (new method I added in this PR).
2. **Adding a `MsgOption` helper method for blocks**
This is pretty self-explanatory. I personally found the attachment version helpful and would like to have a block counterpart.
3. **Adding exported consts for both of the text object types**
Just didn't like having to input the text value every time here.

### DIFF
--- a/block.go
+++ b/block.go
@@ -31,6 +31,12 @@ const (
 
 // block defines an interface all block types should implement
 // to ensure consistency between blocks.
+type block interface {
+	blockType() MessageBlockType
+}
+
+// Block is the exported version of the block interface and is used to pass
+// multiple blocks in exported variadic functions, etc.
 type Block interface {
 	blockType() MessageBlockType
 }
@@ -45,7 +51,7 @@ func NewBlockMessage(blocks ...Block) Message {
 }
 
 // AddBlockMessage appends a block to the end of the existing list of blocks
-func AddBlockMessage(message Message, newBlk Block) Message {
+func AddBlockMessage(message Message, newBlk block) Message {
 	message.Msg.Blocks = append(message.Msg.Blocks, newBlk)
 	return message
 }
--- a/block.go
+++ b/block.go
@@ -29,14 +29,8 @@ const (
 	motOptionGroup  MessageObjectType = "option_group"
 )
 
-// block defines an interface all block types should implement
+// Block defines an interface all block types should implement
 // to ensure consistency between blocks.
-type block interface {
-	blockType() MessageBlockType
-}
-
-// Block is the exported version of the block interface and is used to pass
-// multiple blocks in exported variadic functions, etc.
 type Block interface {
 	blockType() MessageBlockType
 }
@@ -51,7 +45,7 @@ func NewBlockMessage(blocks ...Block) Message {
 }
 
 // AddBlockMessage appends a block to the end of the existing list of blocks
-func AddBlockMessage(message Message, newBlk block) Message {
+func AddBlockMessage(message Message, newBlk Block) Message {
 	message.Msg.Blocks = append(message.Msg.Blocks, newBlk)
 	return message
 }
--- a/block.go
+++ b/block.go
@@ -31,12 +31,12 @@ const (
 
 // block defines an interface all block types should implement
 // to ensure consistency between blocks.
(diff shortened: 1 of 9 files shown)


alyosha:
@james-lawrence I noticed after pushing that you requested the interfaces be made private in @MattDavisRV's previous PR (#480)

As I mention above, the main reason I wanted to make them public was so that I could use an array of blocks/elements with variadic functions like [`NewActionBlock`](https://github.com/nlopes/slack/blob/master/block_action.go#L18) and `MsgOptionBlocks`

Good chance I am missing something here, so if there is a way to accomplish the same thing while keeping the interface private, please let me know.

james-lawrence:
@alyosha just define an identical interface if you want to do that.

alyosha:
@james-lawrence hope I've understood you correctly here. I defined an identical exported interface for both `block` and `blockElement` and am only using them in the locations I require them. In all other cases, I have left the private interface as is. 

The one exception was with `blockObject`, which only was being used in one place previously, but which was problematic for the same issues as mentioned above. I opted to simply replace the existing interface in this case since it would be unused were I to define an identical public interface, but please let me know your thoughts. 

thanks in advance as always 🙏

sargun:
Could you potentially break out the `MsgOptionBlocks` and `PlainTextType` bits out of this PR? Without it, having to do the JSON encoding by-hand is not ideal for blocks, and kind of a blocker.

james-lawrence:
@alyosha guess I was unclear. I simply meant that 3rd party applications can define identical interfaces in their code instead of us exporting if they wanted to write their own vararg functions. not that we should define another one in the library (which is just silly as you noted).

the functions you added don't require exposing the interfaces directly by the library.


is there an actual usecase for people that exposing the interfaces unlocks that is worth the API increase and isn't just missing functionality in the library itself?

alyosha:
@james-lawrence for both the `MsgOptionBlocks` method I added, as well as for `NewBlockMessage`, `NewActionBlock` and `NewContextBlock`, it is not possible to use these methods when you have a variadic number of blocks/blockElements and you don’t know how many you will end up with.

If you try to do the following, the program will not compile:
```
elements := []slack.ButtonBlockElement{...}
blck := slack.NewActionBlock("sample_id", elements...)
```
`cannot use elements (type []slack.ButtonBlockElement) as type []slack.blockElement in argument to slack.NewActionBlock`

and obviously you cannot make an array of the interface `blockElement` type since it is unexported.

What I want to do is the following:
```
elements := []slack.BlockElement{...}
blck := slack.NewActionBlock("sample_id", elements...)
```

in my mind this is the only meaningful way to use this method, since in most of my use cases I have no idea how many elements/blocks I will need.

defining an interface locally will not solve this problem as far as I can see, which is why I misunderstood your comment. even if it did, I consider the functionality it provides pretty fundamental to the block logic so I'd like to see it in the library itself. I  have updated my PR again to reflect the changes I'd like to see.

alyosha:
@sargun while I definitely agree and am blocked in a similar way on my projects, I think we need to figure out this interface question first otherwise it will be impossible to pass a variadic number of blocks to the `MsgOptionBlocks` method (unless someone has an alternative approach)

alyosha:
@james-lawrence I only just noticed this issue now after the fact, but it seems like a number of other people have encountered this same problem and proposed exporting the interfaces as a solution:
https://github.com/nlopes/slack/issues/476#issuecomment-471189536

I understand your position but if you are against exporting, it would be great if you could provide an alternative means to accomplish the above functionality :bow:

james-lawrence:
yeah not against it just thinking it through =)

james-lawrence (approved):
LGTM.

alyosha:
Thanks as always @james-lawrence :100:

Issue #493, opened 2019-03-24
GetChannelInfo fails if channel history has Block messages


khayyamsaleem:
Because `GetChannelInfo` includes recent messages, and blocks were just recently supported here, `GetChannelInfo` is failing, with the error:
`json: cannot unmarshal object into Go struct field Message.blocks of type slack.Block`

khayyamsaleem:
@alyosha #476 #490

khayyamsaleem:
my workaround for getting a list of members for a channel without `GetChannelInfo`:
```go
slackAPI := slack.New(os.Getenv("SLACK_TOKEN"))
allUsers, _, err := slackAPI.GetUsersInConversation(&slack.GetUsersInConversationParameters{
	ChannelID: os.Getenv("SLACK_CHANNEL_ID"),
	Cursor:    "",
	Limit:     100,
})
//allUsers is an array of userIDs (strings)
```

alyosha:
This same error will occur for interaction callbacks, and any other method that tries to unmarshal the block JSON into the Block interface. Should be addressed in the same PR I mentioned in https://github.com/nlopes/slack/issues/476 thanks for the mention :+1:

james-lawrence:
can we confirm this has been fixed?

jwatroba:
@khayyamsaleem are you able to confirm if this is fixed? Thanks.

khayyamsaleem:
@jwatroba i've since been using the workaround i mentioned, but I can try reverting and get back to you.

alyosha:
@khayyamsaleem any update on this? think it should probably be resolved

khayyamsaleem:
@alyosha just confirmed, master's GetChannelInfo function works

james-lawrence:
great, closing.

PR #494, opened 2019-03-26
Refactor blocks implementation

alyosha:
# Summary
The current use of `Block`/`BlockElement` interfaces as initially introduced in #480 is proving problematic as Go interfaces have no underlying data structures into which we can unmarshal their corresponding JSON values. At the moment, attempts to unmarshal `Block`/`BlockElement` result in an error message similar to the one seen in issue #493. 

In addition to the above issue, Slack is now returning dynamic JSON values for each of the following keys, so some sort of type assertion and custom unmarshalling seems necessary regardless.
- `actions` (i.e. attachment or block)
- `blocks`
- `block_elements`

(not an exhaustive list)

The interfaces themselves still have a valid use case in simplifying the process of creating/sending Block messages, so I have left them more or less as is (still exported).

I have added new structs for use in unmarshalling/accessing the JSON response and implemented custom unmarshalling/marshalling logic for each where necessary. 

I wanted to get a working PR up ASAP and I realize this is still rather rough. I will be adding more tests and cleaning up in the next few days, but in the meantime please feel free to leave feedback. Also, if there is an easier way to accomplish the same thing, I am all ears.


block_conv.go
```diff
@@ -0,0 +1,359 @@
...
+// UnmarshalJSON implements the Unmarshaller interface for Blocks, so that any JSON
+// unmarshalling is delegated and proper type determination can be made before unmarshal
+func (b *Blocks) UnmarshalJSON(data []byte) error {
+	var raw []json.RawMessage
+	err := json.Unmarshal(data, &raw)
+	if err != nil {
+		return err
+	}
+
+	for _, r := range raw {
```
james-lawrence, line 39: this loses ordering of the blocks doesn't it?

the blocks type seems unncessary we want this function but have it return an array of the Block interface.
```golang
func foo(data []byte) (blocks []block, err error) {
        type union struct {
           Type string `json:"type"`
        }
	err := json.Unmarshal(data, &raw)
	if err != nil {
		return err
	}
        for _, r := range raw {
            u := union{}
            json.Unmarshal(r, &u)

            switch(u.Type) {
            case "actions":
                var tmp ActionBlock
                json.Unmarshal(r, &tmp)
                blocks = append(blocks, tmp)
            }
        }
        return blocks, nil
}
```

block_conv.go
```diff
@@ -0,0 +1,359 @@
...
+		return nil, err
+	}
+	return block, nil
+}
+
+func (b *Blocks) appendToBlocks(appendBlocks []Block) {
+	for _, block := range appendBlocks {
+		switch block.(type) {
+		case *ActionBlock:
+			b.ActionBlocks = append(b.ActionBlocks, block.(*ActionBlock))
```
james-lawrence: assertions are unnecessary, though I believe this method goes away anyways
```golang
 switch x := block.(type) {
 case *ActionBlock: b.ActionBlocks = append(b.ActionBlocks, x);
```

block_conv.go
alyosha: @james-lawrence if we return an array of the block interface, how do you actually access the fields in that block? I want to be able to unmarshal the callback response, and then access the data in it for subsequent tasks, hence the `Blocks` type

khayyamsaleem:
can some examples of handling the block interaction callbacks be added to the examples library?

alyosha:
@khayyamsaleem sure I will add some examples later, but the basic idea is that any blocks present will be unmarshalled into their type's corresponding slice in the `Blocks` struct 👍

block_conv.go
james-lawrence: type switch. we're effectively talking about sum types here. which generally use type switches to determine the current type of an element.

I think its more important to maintain order of the messages than it is for each to be explicitly typed in the results.

khayyamsaleem:
@alyosha is it possible to receive the Block Interactions as RTM events?

block_conv.go
alyosha: @james-lawrence that's a fair point about ordering that I hadn't really considered. I suppose if you wanted to replace a message with the same block order + some small modifications, it would be pretty inconvenient to not know the original order of the blocks.

just to confirm, you are suggesting that library users implement the type switch themselves directly, yes?

alyosha:
@khayyamsaleem not sure I fully understand your question, but blocks are interactive components so the callbacks from interactions with them should all be forwarded to the callback URL you specify in your app dashboard

block_conv.go
alyosha: will update the PR after confirming this point and ask for re-review 👍

james-lawrence: Yes, I'm suggesting that library users implement the type switch themselves as they know exactly how they want to treat the blocks.

james-lawrence:
@alyosha as an aside I did a lot of internal refactoring last week of the internal api's in the library. I doubt much of it impacts you but you may want to rebase off master.

block_conv.go
alyosha: All right I will adapt this and leave some examples on how to unmarshal, thanks for the response

alyosha:
sorry for the wait all, confirmed how to move forward with the dynamic unmarshalling so I will try and get this fixed/ready for re-review by @james-lawrence this week
🙏

examples/blocks/blocks.go
```diff
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"github.com/nlopes/slack"
+	"github.com/alyosha/slack"
```
alyosha: oops 👀 🤔

alyosha:
@james-lawrence not sure if something is up with the CI, but keeps failing with the following message:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated.
```

Can run tests locally without issue if I switch all the package names to point to my fork and not the upstream

james-lawrence:
interesting, will take a look this weekend.

alyosha:
closing and reopening PR to try and address CI issue
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/27185#issuecomment-404503572

alyosha:
I am now able to reproduce the hanging test behavior locally and am continuing to debug.

alyosha:
CI issue resolved just making a few more convenience changes and adding examples, will re-request review tomorrow

block_element.go
```diff
@@ -3,26 +3,26 @@ package slack
...
 const (
-	metImage      MessageElementType = "image"
-	metButton     MessageElementType = "button"
-	metOverflow   MessageElementType = "overflow"
-	metDatepicker MessageElementType = "datepicker"
-	metSelect     MessageElementType = "static_select"
-
-	mixedElementImage MixedElementType = "mixed_image"
-	mixedElementText  MixedElementType = "mixed_text"
+	MetImage      MessageElementType = "image"
```
james-lawrence: lets keep acronyms in caps.

block_context.go
```diff
@@ -7,18 +7,29 @@ package slack
...
-func (s ContextBlock) blockType() MessageBlockType {
+// BlockType returns the type of the block
+func (s ContextBlock) BlockType() MessageBlockType {
 	return s.Type
 }
 
-// NewContextBlock returns a newinstance of a context block
-func NewContextBlock(blockID string, elements ...BlockObject) *ContextBlock {
+type ContextElements struct {
+	ContextElementSet []MixedElement
```
james-lawrence: s/ContextElementSet/Elements/

block_context.go
```diff
@@ -7,18 +7,29 @@ package slack
...
+}
+
+// NewContextElements is a convenience method for generating ContextElements
+func NewContextElements(contextElements []MixedElement) ContextElements {
```
james-lawrence: s/[]MixedElement/...MixedElement/

block_element.go
alyosha: will update this for all occurrences but probably tomorrow since I'm about to sleep

block_context.go
james-lawrence: actually does this method need to exist? or can we just use `...MixedElement` for NewContextBlock

block_context.go
alyosha: agree `Elements` is better here. I went with `BlockSet` in the case of `Blocks` because I couldn't really come up with anything better, but totally open to suggestions

block_conv.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	err := json.Unmarshal(data, &raw)
+	if err != nil {
+		return err
+	}
+
+	var blocks Blocks
+	for _, r := range raw {
+		var obj map[string]interface{}
```
james-lawrence: define a type inside this method for this unmarshalling. generally avoid maps were possible.
```
type sumtype struct {
  blockType string `json:"type"`
}
s := sumtype{}
json.Unmarshal(r, &s)
```

block_object.go
```diff
@@ -1,44 +1,132 @@
...
-		AltText:  altText,
+func unmarshalBlockObject(r json.RawMessage, object blockObject) (blockObject, error) {
+	err := json.Unmarshal(r, object)
+	if err != nil {
+		return nil, err
+	}
+	return object, nil
+}
+
+func (o *BlockObjects) appendToBlockObjects(appendObjects []blockObject) {
```
james-lawrence: this function seems entirely unnecessary, also seems unused now.

block_context.go
alyosha: will look into this tomorrow ^ :eyes:

block_conv.go
james-lawrence: you could also define this type and embedded it in the block types. ./shrug depends how much it'll be used.

block_object.go
alyosha: might indeed be unnecessary now, will confirm tomorrow 👍

alyosha: removed

block_conv.go
alyosha: changed from map to struct

block_context.go
alyosha: indeed did not need to exist; removed and updated as suggested

block_context.go
alyosha: Maybe `LayoutBlocks` would be a better naming choice than `BlockSet`?

block_element.go
alyosha: done ^

alyosha:
@james-lawrence updated based on your feedback and added a type switch example to `examples/block.go`. left the conversations unresolved in case that was easier but feel free to resolve them if you're good 👍

harrisonturton:
This is a big blocker for me - glad to see it being resolved!

alyosha:
@harrisonturton yeah this is a blocker for all of my personal projects as well at the moment (no pun intended). sorry for the delay on getting this done 🙏

alyosha:
@james-lawrence know how busy you are but please give this another look if you have a second 🙏

alyosha:
@james-lawrence friendly ping 🙃

james-lawrence:
sorry, been swamped w/ non-software related issues. I'll get to this tonight.

dolfelt:
I noticed when using this branch that on marshaling back to JSON there is an unnecessary blocks key that causes errors when responding to interaction requests.

```json
{
    "replace_original": false,
    "delete_original": true,
    "blocks": {
        "blocks": null
    }
}
```

When adding `omitempty` to the `BlockSet` it fixes the issue.

```go
type Blocks struct {
	BlockSet []Block `json:"blocks,omitempty"`
}
```

block.go
```diff
@@ -7,45 +7,55 @@ package slack
...
 // to ensure consistency between blocks.
 type Block interface {
-	blockType() MessageBlockType
+	BlockType() MessageBlockType
+}
+
+// Blocks is a convenience struct defined to allow dynamic unmarshalling of
+// the "blocks" value in Slack's JSON response, which varies depending on block type
+type Blocks struct {
+	BlockSet []Block `json:"blocks"`
```
james-lawrence, line 28: omit empty to prevent valdiation errors as noted in the other comment.

james-lawrence (approved):
Okay, so I'm going to merge this as is. even w/ the follow up work required just to unblock it.

and we'll go from there. plenty of stuff I can see to improve but should block hte bulk of the work.

please follow up with a PR fixing the validation erorr.

alyosha:
will follow up with the `omitempty` change tonight–thanks as always @james-lawrence and thanks also to @dolfelt

PR #495, opened 2019-03-28
Adds support for hint text on select inputs in Dialogs.

ewollesen:
While it's not explicitly called out in their documentation, it is
supported (tested empirically), and it's shown in screenshots on
https://api.slack.com/best-practices/creating-useful-dialogs


james-lawrence:
LGTM, thanks.

PR #500, opened 2019-04-04
refactor to ease testing

james-lawrence:
lots of churn on this PR to pull some helper methods from package scoped to client scoped.

BREAKING:
- APIURL is now const, can no longer set it. use OptionAPIURL to
adjust the endpoint for a client.

WARNINGS:
- Take special care if you rely on File upload/download and
SetUserPhotoContext

- refactor internal api to support per client APIURL configuration
- adds a slacktest package (copied and modified from
lusis/slack-test)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,38 +2,52 @@
 
 
 [[projects]]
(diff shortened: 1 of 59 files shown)


james-lawrence:
races just in test package will cleanup this weekend.

victorcoder:
Thanks @james-lawrence

PR #503, opened 2019-04-04
Update usergroups.go to support default channel updates

atatkin:
Allows updates to usergroups to update the channels field of the usergroup.


atatkin:
Linking the API docs real quick to show how channels are updated in their API.

https://api.slack.com/methods/usergroups.update

james-lawrence:
LGTM thanks.

PR #505, opened 2019-04-07
more expressive incoming webhook errors

rgarcia:
- Export StatusCodeError so that library consumers can check for
  errors of this type and condition on status codes
- Return RateLimitedError from PostWebhook() so that consumers can
  implement retry logic.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -42,16 +42,16 @@ func (t SlackResponse) Err() error {
 
 // StatusCodeError represents an http response error.
 // type httpStatusCode interface { HTTPStatusCode() int } to handle it.
-type statusCodeError struct {
+type StatusCodeError struct {
 	Code   int
 	Status string
 }
 
-func (t statusCodeError) Error() string {
+func (t StatusCodeError) Error() string {
 	return fmt.Sprintf("slack server error: %s", t.Status)
 }
 
-func (t statusCodeError) HTTPStatusCode() int {
+func (t StatusCodeError) HTTPStatusCode() int {
 	return t.Code
 }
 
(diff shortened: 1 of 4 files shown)


misc.go
```diff
@@ -42,16 +42,16 @@ func (t SlackResponse) Err() error {
 
 // StatusCodeError represents an http response error.
 // type httpStatusCode interface { HTTPStatusCode() int } to handle it.
-type statusCodeError struct {
```
james-lawrence, line 45: this is not necessary.

library consumers can use:
```
type statuscode interface {
   HTTPStatusCode() int
}
if x, ok := err.(statuscode); ok {
}
```

james-lawrence (changes requested):

misc.go
rgarcia: This seems pretty cumbersome and not idiomatic to me

james-lawrence: I tend to agree w/ [dave cheney on errors](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully).

and so does the golang team if you look at their proposed changes to error handling coming up. 

short answer: won't be merged long as this change is included. doesn't provide any additional utility and hamstrings changes to the statuscode error type in the future.

james-lawrence: as an aside, I'm happy to alleviate some of the usability issues of the statuscode error in other ways. such as having the interface type exposed in slackutils package (with proper tests tying it back to statuscode so when I invariably break it a test warns us)

rgarcia: Thanks for the link, TIL I've been doing error types incorrectly. The specific usability issue we're having is exactly the one in Dave Cheney's example: deciding when to retry an error from Slack. I think a Retryable interface would solve that. Will give that a go.

The slackutils package does give me some pause... https://blog.golang.org/package-names

james-lawrence: the slackutils package is suppose to give you pause. but there is precedence see httputils etc.

slackutils is for stuff that isn't well thought out, or well structured enough to put into the core library.

I'm also less strict about breaking changes in slackutils. its meant to be a grab bag that may break of utility stuff.

rgarcia:
closing in favor of https://github.com/nlopes/slack/pull/513/files

PR #508, opened 2019-04-11
added enterprise id to auth test

sdalezman:
enterprise id is an optional argument from auth test to determine if
you're dealing with an enterprise org


james-lawrence:
LGTM, thanks.

PR #509, opened 2019-04-13
Improve NewGroupedSelectDialog

### LINKED ISSUE #419: NewGroupedSelectDialogInput() output appears in random order
`func NewGroupedSelectDialogInput(name, label string, groups map[string]map[string]string) *DialogInputSelect`

Since the input argument of this function is a map, and the function itself does nothing to enforce an order, the output is random.

tylarb:
- Add convinience function `NewDialogOptionGroup` and test for the same
- move `optionsFromArray` and `optionsFromMap` into `selectOptions....` and into the testing, as they aren't used in the exported library whatsoever.
- Test refractor supporting the above
- Should resolve https://github.com/nlopes/slack/issues/419

Authored-by: Tyler Ramer <tyaramer@gmail.com>

### DIFF
--- a/dialog_select.go
+++ b/dialog_select.go
@@ -55,14 +55,7 @@ func NewStaticSelectDialogInput(name, label string, options []DialogSelectOption
 }
 
 // NewGroupedSelectDialogInput creates grouped options select input for Dialogs.
-func NewGroupedSelectDialogInput(name, label string, groups map[string]map[string]string) *DialogInputSelect {
-	optionGroups := []DialogOptionGroup{}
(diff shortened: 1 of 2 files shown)


tylarb:
Just realized that it looks like this is now deprecated and as well the option "name" isn't `label` but `text`. Let me test this, or we can pass it by in favor of blocks anyway.

https://api.slack.com/docs/message-menus

james-lawrence:
LGTM, I'm accepting this regardless of deprecation status. as it improves the situation and deprecated doesn't mean unused.

PR #512, opened 2019-04-15
Add test for if parse for user field is enable

shiimaxx:
I think https://github.com/nlopes/slack/issues/365 was already fixed by being exported `GroupConversation` contained in `Channel` struct.

I added a test case that checks if parse for user field is enabled for confirming above.


james-lawrence:
thanks, I love receiving test cases.

PR #513, opened 2019-04-15
add retryable behavior to errors

rgarcia:
Makes it easier to decide when errors are retryable. Continuation / replacement PR for feedback in https://github.com/nlopes/slack/pull/505/files

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -55,6 +55,13 @@ func (t statusCodeError) HTTPStatusCode() int {
 	return t.Code
 }
 
+func (t statusCodeError) Retryable() bool {
+	if t.Code >= 500 || t.Code == http.StatusTooManyRequests {
+		return true
+	}
+	return false
+}
+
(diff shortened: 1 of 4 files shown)


misc.go
```diff
@@ -55,6 +55,13 @@ func (t statusCodeError) HTTPStatusCode() int {
 	return t.Code
 }
 
+func (t statusCodeError) Retryable() bool {
+	if t.Code >= 500 || t.Code == http.StatusTooManyRequests {
```
james-lawrence, line 59: when will this method ever return false.... =)

rgarcia: oops e09ebc0

james-lawrence: TBH I wouldn't expect a 500 to be something that can be retried. but i'm going to let it slide for now. I reserve the right to remove it later.

james-lawrence (approved):

PR #515, opened 2019-04-17
files.list: support for cursor based pagination

jasonhancock:
I have a slack workspace with ~6500 files uploaded to it. When trying to page through all 6500 with count/page based pagination, I noticed that after page 100 I would only get the first page of results back. After filing a support ticket with slack, they steered me to cursor based pagination for this endpoint. I didn't want to break the existing GetFiles/GetFilesContext api, so added ListFiles/ListFilesContext instead.

With these changes I'm able to page through all of my files without issue.

### DIFF
--- a/files.go
+++ b/files.go
@@ -133,11 +133,21 @@ type GetFilesParameters struct {
 	Page          int
 }
 
+// ListFilesParameters contains all the parameters necessary (including the optional ones) for a ListFiles() request
+type ListFilesParameters struct {
+	Limit   int
+	User    string
+	Channel string
+	Types   string
+	Cursor  string
+}
+
 type fileResponseFull struct {
 	File     `json:"file"`
 	Paging   `json:"paging"`
-	Comments []Comment `json:"comments"`
-	Files    []File    `json:"files"`
+	Comments []Comment        `json:"comments"`
+	Files    []File           `json:"files"`
+	Metadata ResponseMetadata `json:"response_metadata"`
 
 	SlackResponse
 }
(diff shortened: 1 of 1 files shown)


files.go
```diff
@@ -196,6 +206,37 @@ func (api *Client) GetFiles(params GetFilesParameters) ([]File, *Paging, error)
 	return api.GetFilesContext(context.Background(), params)
 }
 
+// ListFiles retrieves all files according to the parameters given. Uses cursor based pagination.
+func (api *Client) ListFiles(params ListFilesParameters) ([]File, *ResponseMetadata, error) {
```
james-lawrence: return a single type here with a Next field that is ListFilesParameters.

james-lawrence (changes requested):
please mark the old methods as deprecated if it makes sense.

files.go
jasonhancock: That seems different than what other functions with cursor based paging already have implemented in the package. For example, the [GetUsersInConversationContext](https://godoc.org/github.com/nlopes/slack#Client.GetUsersInConversationContext) returns the cursor as a string as the second return argument.

Based on that finding, should it be this instead?

```
func (api *Client) ListFiles(params ListFilesParameters) ([]File, string, error)
```

jasonhancock: @james-lawrence any thoughts on ^^?

james-lawrence: sorry for the delays, I prefer the approach taken by [GetUsersPaginated](https://godoc.org/github.com/nlopes/slack#Client.GetUsersPaginated)

jasonhancock: implemented `return a single type here with a Next field that is ListFilesParameters.`, however left the name as `Cursor` to more closely align with the API parameter names.

james-lawrence (approved):
good enough, thanks.

PR #518, opened 2019-04-21
Add `auth.test` to supported APIs in the test server. 

warrengray:
I'm implementing a slack bot that makes use of the [auth.test](https://api.slack.com/methods/auth.test) API. For testing purposes in my own project (and perhaps the convenience of others), I've added support for that API call to the test server.

There are no behavioural changes to the core library, and I've added tests to cover the new handler in the style of the existing tests.

Please let me know if I've missed anything.

### DIFF
--- a/slacktest/handlers.go
+++ b/slacktest/handlers.go
@@ -26,7 +26,12 @@ func contextHandler(server *Server, next http.HandlerFunc) http.Handler {
 	})
 }
 
-func usersInfoHandler(w http.ResponseWriter, r *http.Request) {
+// handle auth.test
+func authTestHandler(w http.ResponseWriter, _ *http.Request) {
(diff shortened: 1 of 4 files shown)


slacktest/handlers.go
```diff
@@ -26,7 +26,12 @@ func contextHandler(server *Server, next http.HandlerFunc) http.Handler {
 	})
 }
 
-func usersInfoHandler(w http.ResponseWriter, r *http.Request) {
+// handle auth.test
+func authTestHandler(w http.ResponseWriter, _ *http.Request) {
```
warrengray, line 30: The `r` variables were unused in some of the handler methods, so I changed them to the blank identifier.

james-lawrence:
ahh wonderful, thanks for the imrpovement.

PR #523, opened 2019-04-25
ADDED: pagination suppor fot GetChannels

jescarri:
Related to: https://github.com/nlopes/slack/issues/522

##### Pull Request Guidelines

- [x] is it application specific (caching, retry logic, rate limiting, etc)

##### API changes

### DIFF
--- a/channels.go
+++ b/channels.go
@@ -293,12 +293,20 @@ func (api *Client) GetChannelsContext(ctx context.Context, excludeArchived bool,
 			return nil, err
 		}
 	}
-
-	response, err := api.channelRequest(ctx, "channels.list", config.values)
-	if err != nil {
-		return nil, err
+	var channels []Channel
+	for {
+		response, err := api.channelRequest(ctx, "channels.list", config.values)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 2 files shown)


jescarri:
@nlopes @james-lawrence hey guys any input here?

Thanks!

channels.go
```diff
@@ -293,12 +293,20 @@ func (api *Client) GetChannelsContext(ctx context.Context, excludeArchived bool,
 			return nil, err
 		}
 	}
-
-	response, err := api.channelRequest(ctx, "channels.list", config.values)
-	if err != nil {
-		return nil, err
+	var channels []Channel
```
james-lawrence, line 296: going to ask you to follow the apgination pattern set elsewhere in the library. see user pagination for an example.

james-lawrence (changes requested):

channels.go
jescarri: thanks for the review, yep will try to submit something in the upcoming weeks, it's been hectic over here.

ollieparsley:
@jescarri Channels pagination support was added in #641, so I'll close this PR.

PR #525, opened 2019-05-01
Added style parameter to ButtonBlockElement

matt-harding:
The Style property of the block button ([https://api.slack.com/reference/messaging/block-elements#button](url)) was not available in the BlockButtonElement struct. This is used to decorate buttons with alternative visual color schemes.


james-lawrence:
thanks

PR #528, opened 2019-05-10
Fix WithStyle method on ButtonBlockElement

rpunkfu:
Signature for `WithStyle` method looks like:
```go
func (s ButtonBlockElement) WithStyle(style Style)
```

This prevents us from changing style on ButtonBlockElement after defining it and makes method pretty much useless.
Fix is to define this method on pointer to ButtonBlockElement.


james-lawrence:
ha, good catch. thanks.

PR #529, opened 2019-05-13
Fix Blocks marshalling

olivierlemasle:
Custom marshalling was provided for `*Blocks` instead of `Blocks`, which means that a `Message` containing blocks was marshalled with an extraneous `blocks` value.

Cf `examples/blocks`:

```
--- Begin Example One ---
{
    "replace_original": false,
    "delete_original": false,
    "blocks": {
        "blocks": [
            {
                "type": "section",
                "text": {
                    "type": "mrkdwn",
                    "text": "You have a new request:\n*\u003cfakeLink.toEmployeeProfile.com|Fred Enriquez - New device request\u003e*"

...
```

This commit fixes `Block`'s marshalling.

After the commit, `examples/blocks` prints:

```
--- Begin Example One ---
{
    "replace_original": false,
    "delete_original": false,
    "blocks": [
        {
            "type": "section",
            "text": {
                "type": "mrkdwn",
                "text": "You have a new request:\n*\u003cfakeLink.toEmployeeProfile.com|Fred Enriquez - New device request\u003e*"
...
```

### DIFF
--- a/block_conv.go
+++ b/block_conv.go
@@ -12,7 +12,7 @@ type sumtype struct {
 
 // MarshalJSON implements the Marshaller interface for Blocks so that any JSON
 // marshalling is delegated and proper type determination can be made before marshal
-func (b *Blocks) MarshalJSON() ([]byte, error) {
+func (b Blocks) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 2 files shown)


james-lawrence:
thanks.

Issue #533, opened 2019-05-23
Add option to optional input


alexfernandessd:
Somes inputs must be optional.

PR #534, opened 2019-05-23
add optional inputText

alexfernandessd:
Related: #533 

##### Pull Request Guidelines

##### API changes

### DIFF
--- a/dialog_text.go
+++ b/dialog_text.go
@@ -3,8 +3,8 @@ package slack
 // TextInputSubtype Accepts email, number, tel, or url. In some form factors, optimized input is provided for this subtype.
 type TextInputSubtype string
 
-// TextInputOption hndle to optional inputs
-type TextInputOption func(*TextInputOption)
+// TextInputOption handle to extra inputs options.
+type TextInputOption func(*TextInputElement)
 
 const (
 	// InputSubtypeEmail email keyboard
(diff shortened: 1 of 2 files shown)


dialog_text.go
```diff
@@ -26,12 +26,13 @@ type TextInputElement struct {
 }
 
 // NewTextInput constructor for a `text` input
-func NewTextInput(name, label, text string) *TextInputElement {
+func NewTextInput(name, label, text string, optional bool) *TextInputElement {
```
james-lawrence: prefer `options ...TextInputOption` instead of just concating fields to the function signature.

```
type TextInputOption func(*TextInputOption)
var options = []TextInputOption{}
t := &TextInputElement{}
for _, opt := range options {
  opt(&t)
}
```

james-lawrence (changes requested):
library api consistency.

dialog_text.go
alexfernandessd: Thank you for your comment! Can you say me if i understood your idea with my last commit? (i'm new in language).

james-lawrence:
that is precisely what I meant.

james-lawrence (approved):
LGTM, sorry for the delay, life was busy and wasnt sure fi you were done =)

PR #535, opened 2019-05-26
Block fixes pt.1

alyosha:
## Summary
- When originally adding the dynamic unmarshalling method for block elements, I only included `static_select`, but there are a number of other select menus that need to be supported. This PR adds them.
- Also currently unable to properly unmarshal the callback from `block_actions` type interaction callbacks because of a lot of missing fields.
- Added new interaction type for convenience in handling interaction callbacks.

Assume there are probably more fixes to come, will keep an eye out.

### DIFF
--- a/block.go
+++ b/block.go
@@ -30,12 +30,22 @@ type Blocks struct {
 
 // BlockAction is the action callback sent when a block is interacted with
 type BlockAction struct {
-	ActionID string          `json:"action_id"`
-	BlockID  string          `json:"block_id"`
-	Text     TextBlockObject `json:"text"`
-	Value    string          `json:"value"`
-	Type     actionType      `json:"type"`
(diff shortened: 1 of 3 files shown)


alyosha:
@james-lawrence my fault for missing this in original implementation, if you have a minute 🙏

alyosha:
@james-lawrence sorry for bugging you, just wanted to clarify that while #538 and #539 are lower priority, the issues addressed in this PR will block anyone who is trying to handle callbacks from a `select_menu` that uses a type other than `static_select`

know how busy you are, thanks in advance :pray:

james-lawrence (approved):
LGTM.

james-lawrence:
thanks for doing the heavy lifting on this Alyosha.

PR #538, opened 2019-05-30
Add block ID parameter to NewSectionBlock method

alyosha:
## Summary
This change is ultimately a nice to have and may not be worth it. Up to discussion on this one, but my thought process is as follows:

- `SectionBlock` can take any `BlockElement` type under the `Accessory` field, meaning that an action callback can potentially be triggered off of the block 
  - e.g. if providing a `DatePickerBlockElement` as `Accessory` 

- In my projects using `ActionBlocks` blocks so far, I've found it easiest to handle the callback using a switch case on `BlockID`

- Currently, using the `NewSectionBlock` method to generate a new `SectionBlock` results in an empty `BlockID` that is then transformed into a random string at the time of post. 
  - This means that it is not possible to implement a switch like the one described above unless I modify the block value returned from `NewSectionBlock`

I find it most convenient to pre-declare all my layout blocks as vars in a single file and then just pick and choose the ones I need, so right now I am just using a wrapper method around `NewSectionBlock`, but it would be nice to just set it when calling `NewSectionBlock` itself

### DIFF
--- a/block_section.go
+++ b/block_section.go
@@ -16,13 +16,27 @@ func (s SectionBlock) BlockType() MessageBlockType {
 	return s.Type
 }
 
+// SectionBlockOption allows configuration of options for a new section block
+type SectionBlockOption func(*SectionBlock)
+
+func SectionBlockOptionBlockID(blockID string) SectionBlockOption {
+	return func(block *SectionBlock) {
+		block.BlockID = blockID
+	}
+}
+
 // NewSectionBlock returns a new instance of a section block to be rendered
-func NewSectionBlock(blockID string, textObj *TextBlockObject, fields []*TextBlockObject, accessory *Accessory) *SectionBlock {
-	return &SectionBlock{
+func NewSectionBlock(textObj *TextBlockObject, fields []*TextBlockObject, accessory *Accessory, options ...SectionBlockOption) *SectionBlock {
+	block := SectionBlock{
 		Type:      MBTSection,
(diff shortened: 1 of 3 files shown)


block_section.go
```diff
@@ -17,10 +17,11 @@ func (s SectionBlock) BlockType() MessageBlockType {
 }
 
 // NewSectionBlock returns a new instance of a section block to be rendered
-func NewSectionBlock(textObj *TextBlockObject, fields []*TextBlockObject, accessory *Accessory) *SectionBlock {
+func NewSectionBlock(blockID string, textObj *TextBlockObject, fields []*TextBlockObject, accessory *Accessory) *SectionBlock {
```
james-lawrence: this method is getting to the point where you want to do the following:
```
type SectionBlockOption func(*SectionBlock)
func NewSectionBlock(textObj *TextBlockObject, fields []*TextBlockObject, accessory *Accessory, options ...SectionBlockOption) *SectionBlock {
s := &SectionBlock{}
for _, opt := range options {
  opt(s)
}
// ...
```


*edit* I write english gud.

james-lawrence (changes requested):
minor suggestion.

alyosha:
@james-lawrence sorry for the delay on this one, updated 👍

james-lawrence (approved):

james-lawrence:
thanks!

PR #539, opened 2019-06-01
Make label tag in OptionGroupBlockObject optional

alyosha:
## Summary
When populating an `external_select` menu's choices via the options callback, we need to respond with an array of options/single option without the `label` field. If we specify the omitempty encoding option here, we can reuse this struct for this response as well (fails to populate menu if `label` is encoded as `null`)

https://api.slack.com/reference/messaging/block-elements#external-select


james-lawrence (approved):
LGTM.

PR #540, opened 2019-06-02
handle tokens_revoked event

KavenC:
## Summary

Implemented support for [token_revoked](https://api.slack.com/events/tokens_revoked) inner event. This event is necessary for managing access tokens among different teams.

## Typical Use-Case

For a slack app with a bot user, the app maintains access tokens for bot user permission to the team that installed the app.

When a team revoked the tokens (usually also uninstalled app), slack app should also remove the `BotID, AccessToken` pair in the maintained token list.


james-lawrence (approved):
LGTM.

PR #541, opened 2019-06-02
Update the slash command example to use the new request signature verification method

rking788:
I was looking to start using this package and noticed the slash command example was still using the old verification token method of verifying Slack requests. When I started updating the example, I noticed that it was a little tricky to verify the signature using the request body and still parse the form later in the `SlashCommandParse` method.

This change adds a new `WriteBody` method on the secrets verifier. It will read the body and add it to the HMAC contents and replace the request body with a new buffer so it can be read again later.

### DIFF
--- a/examples/slash/slash.go
+++ b/examples/slash/slash.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 
 	"github.com/nlopes/slack"
@@ -11,20 +13,28 @@ import (
 
 func main() {
 	var (
-		verificationToken string
+		signingSecret string
 	)
 
-	flag.StringVar(&verificationToken, "token", "YOUR_VERIFICATION_TOKEN_HERE", "Your Slash Verification Token")
+	flag.StringVar(&signingSecret, "secret", "YOUR_SIGNING_SECRET_HERE", "Your Slack app's signing secret")
 	flag.Parse()
 
 	http.HandleFunc("/slash", func(w http.ResponseWriter, r *http.Request) {
+
+		verifier, err := slack.NewSecretsVerifier(r.Header, signingSecret)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &verifier))
 		s, err := slack.SlashCommandParse(r)
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
 
-		if !s.ValidateToken(verificationToken) {
+		if err = verifier.Ensure(); err != nil {
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}


examples/slash/slash.go
```diff
@@ -11,24 +11,36 @@ import (
...
-		s, err := slack.SlashCommandParse(r)
+
+		verifier, err := slack.NewSecretsVerifier(r.Header, signingSecret)
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
 
-		if !s.ValidateToken(verificationToken) {
+		_, err = verifier.WriteBody(r)
```
james-lawrence: new method should be unnecessary.

```suggestion
		if _, err = io.Copy(verifier, r.Body); err != nil {
```

james-lawrence (changes requested):

examples/slash/slash.go
```diff
@@ -11,24 +11,36 @@ import (
...
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		if err = verifier.Ensure(); err != nil {
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
 
+		s, err := slack.SlashCommandParse(r)
```
james-lawrence: this also will likely not work the way you expect. as you've already read from r.

security.go
```diff
@@ -79,6 +81,17 @@ func (v *SecretsVerifier) Write(body []byte) (n int, err error) {
 	return v.hmac.Write(body)
 }
 
+// WriteBody will write the contents of the request body to the HMAC digest. The request body is then replaced with a new reader
+// so that future uses of the request body will succeeed
+func (v *SecretsVerifier) WriteBody(r *http.Request) (int, error) {
```
james-lawrence: io.Copy does this for you.

rking788: It does one portion of this, writing the bytes to the verifier but it does not replace the request body with a new buffer so that it can be read again later. I have tested this locally and io.Copy consumes the request body and an empty slash command is returned afterwards. 

The unit test I have written shows that the same body can be read again (for slash command parsing) after request verification.

examples/slash/slash.go
rking788: this does not work with io.Copy. i have tested locally and on a remote server taking real slack commands and the new method correctly verifies the request and parses the command from the request afterwards.

security.go
james-lawrence: using either io.MultiWriter or io.TeeReader resolves these problems, regardless this method is unnecessary. happy for a more up to date example but not extraneous code. =)

james-lawrence: just to make it explicit:

```
r.Body = io.TeeReader(r.Body, verifier)
if cmd, err = slack.SlashCommandParse(r); err != nil {}
if err = verifier.Ensure(); err != nil {}
```

james-lawrence: [the original issue](https://github.com/nlopes/slack/issues/353) has a whole discussion around the Write method and the io.Tee etc.

rking788: ah, that makes sense. thanks for clarifying. updated the example to use a TeeReader.

james-lawrence (approved):
LGTM.

PR #543, opened 2019-06-09
Use int for MinQueryLength

bradleyboy:
This updates `MinQueryLength` in the new Blocks implementation of a select menu to be an `int`. Using a string causes an `invalid_blocks error`, as the Slack API expects this to be an integer.


james-lawrence:
LGTM. thanks.

PR #546, opened 2019-06-18
Allow custom http client on webhooks

snebel29:
This is just a convenience function to allow passing a custom `http.Client` without breaking backward compatibility which is quite useful for customising client parameters and testing your own functions.

There seems to have been some effort https://github.com/nlopes/slack/pull/230 into allow custom `http.Client` in the past, but webhooks were not covered by.

I hope this makes sense


james-lawrence:
LGTM.

PR #549, opened 2019-07-03
Correct select dialog items' SelectedOptions type

asf-stripe:
According to https://api.slack.com/dialogs#select_default_values, this field can't be a string: it may hold an array with at most one value, giving an option that comes pre-selected on a select field with an "external" data source.

I tried sending just a value (as I would expect from the `string` type), but slack returns a validation error when you do that. Using this type (and a single element on it) does work.

This is an API change, but I don't expect that anyone using `SelectOptions` on that struct has seen it work. Insofar, it might be a "breaking" change in the sense of "fixing things incompatibly" (:

I didn't add a test, but can't quite tell if we test struct fields here. Let me know if you want one!

### DIFF
--- a/dialog_select.go
+++ b/dialog_select.go
@@ -21,7 +21,7 @@ type DialogInputSelect struct {
 	DialogInput
 	Value           string               `json:"value,omitempty"`            //Optional.
 	DataSource      SelectDataSource     `json:"data_source,omitempty"`      //Optional. Allowed values: "users", "channels", "conversations", "external".
-	SelectedOptions string               `json:"selected_options,omitempty"` //Optional. Default value for "external" only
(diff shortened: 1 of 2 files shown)


rizzza:
@asf-stripe Could you please fix `dialog_test.go`?

asf-stripe:
Whoops, sorry for missing these tests, @rizzza! Fixed now.

rizzza:
> Whoops, sorry for missing these tests, @rizzza! Fixed now.

Hey thanks! I'm not a maintainer though... I hit this bug yesterday. 😉

james-lawrence (approved):
@asf-stripe I'm always happy to take tests with code. =P thanks for the fix.

PR #560, opened 2019-07-24
fix some invalid payload from last Slack API

gghez:
##### Pull Request Guidelines

##### API changes


james-lawrence:
could use a better description of what your trying to fix and the tests should pass.

kanata2:
Please feel free to reopen if you need.

PR #565, opened 2019-08-09
GetUsersContext Bug Fixes

### LINKED ISSUE #557: (api *Client) GetUsersContext() doesn't return error
Actual Behavior: 
The `(api *Client) GetUsersContext()` method stops paginating if it encounters any errors during pagination, and returns a nil error.

Expected Behavior:
The method would return the actual error encountered so that we know that the returned data is incomplete.

Happy to provide more info if necessary, but it seems to be a simple issue w/ the for loop in `GetUsersContext`:
```go
for p = api.GetUsersPaginated(options...); !p.Done(err); p, err = p.Next(ctx) {
	results = append(results, p.Users...)
}
```
`p.Next(ctx)` receives an error from `t.c.userRequest()` on line 328 and returns an error (that isn't the `errPaginationComplete`) without setting the `t.previousResp`. Since we're not done (`!p.Done(err)`), we continue paginating and hit `p.Next(ctx)` again. This time, line 314 evaluates to true (since we have a non-nil `t.previousResp` but an empty `t.previousResp.Cursor` b/c it was never set on line 334 when we returned early from the previous iteration. We then fire off an `errPaginationComplete` error, effectively ending pagination but not letting the user know that their data is incomplete.

A quick fix might be as simple as changing the for loop in `GetUsersContext` to:
```go
for p = api.GetUsersPaginated(options...); !p.Done(err); p, err = p.Next(ctx) {
	if err != nil {
		break
	}
	results = append(results, p.Users...)
}
```

### LINKED ISSUE #563: Auto-Paginated API Methods Can Include Duplicates If Errors Occur
Affects version 0.5.0

Similar to #557 

The method GetUsersContext internally performs pagination. However, if one of the paginated API calls to Slack results in an error (like exceeding a rate limit), then the error is ignored and the results from the previous page are appended _again_ to the results list.

The resulting behavior is that the full results list can contain many many duplicates, especially in the case of exceeding rate limits, since the rate limit backoff is not respected by the pagination code in GetUsersContext.

Suggestion: GetUsersContext should be modified so that the results list is only appended to when there is no error, and should additionally perform backoff when the error is a rate limit error.

Alternate suggestion: The UserPagination object should set its Users property to nil when there is an error.

armsnyder:
Fix #563 (Auto-Paginated API Methods Can Include Duplicates If Errors Occur)
Fix #557 (GetUsersContext() doesn't return error)

Implements the changes discussed in #563 to fix the GetUsersContext bugs above.

The two new tests captured the defective behavior described in the issues. They pass after the change to the GetUsersContext function.

### DIFF
--- a/users.go
+++ b/users.go
@@ -355,7 +355,7 @@ func (api *Client) GetUsersContext(ctx context.Context) (results []User, err err
 		} else if rateLimitedError, ok := err.(*RateLimitedError); ok {
 			select {
 			case <-ctx.Done():
-				return nil, ctx.Err()
+				err = ctx.Err()
 			case <-time.After(rateLimitedError.RetryAfter):
 				err = nil
 			}
--- a/users.go
+++ b/users.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"net/url"
 	"strconv"
+	"time"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


james-lawrence (approved):
Thanks, especially for the tests.

PR #566, opened 2019-08-09
enable setting status for other users than the authenticated one

finfinack:
Adding a function to set the user status for a user other than the authenticated one.

See details in https://github.com/nlopes/slack/issues/562 as well.

In order to retain backwards compatibility, I added a new function instead of changing the interface of the existing ones.

Note: I had a look at the existing tests to see whether I can meaningfully test the user being set correctly but I don't think this can be done easily?


james-lawrence (approved):
we do have a test package you can use to mock out slack responses. its used mainly by the [RTM code](https://github.com/nlopes/slack/blob/master/websocket_managed_conn_test.go#L60).

mock out server responses is acceptable testing as it exercises the code.

i'll accept as is, but if you feel very generous to follow up with a test that'd be wonderful.

PR #569, opened 2019-08-21
Add support for responseURL

Primetalk:
The implementation by @gtrindade has initially been taken from https://github.com/nlopes/slack/issues/387#issuecomment-480961473

Ref https://github.com/nlopes/slack/issues/387

### DIFF
--- a/slack_response_url.go
+++ b/slack_response_url.go
@@ -5,28 +5,38 @@ import (
 	"encoding/json"
 )
 
-// The implementation has initially be taken from https://github.com/nlopes/slack/issues/387#issuecomment-480961473
+// The implementation has initially been taken from https://github.com/nlopes/slack/issues/387#issuecomment-480961473
 // author: gtrindade
 
 // DeleteEphemeral deletes the ephemeral message, use the responseURL from the action payload in your ephemeral message. Only works for the interactive ones.
 func (api *Client) DeleteEphemeral(responseURL string) (*SlackResponse, error) {
+	return api.DeleteEphemeralWithContext(context.Background(), responseURL)
+}
+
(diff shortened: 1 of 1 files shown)


armsnyder:
This is a great addition! Some comments--

1. Could you add versions of these functions which take a `context.Context` as a parameter?
2. I think `slack. Msg` is the wrong type to use here, since it includes some fields which are not accepted when using `resposeUrl`. `slackevents.MessageActionResponse` is more specific.
3. Add unit tests please.

Primetalk:
Hey, @armsnyder. Thank you for the review.
1. Will do.
2. Unfortunately, this code is in `slack`, and I cannot add dependency on `slackevents`.
3. What kind of unit test do you think of? These functions exist only for their side effects, as far as I can see.

armsnyder:
Thanks. I think it would be worthwhile to define a new type for the message in the SendResponse functions. From my own experience, I have had frustrations using responseurl for things like it not supporting attachments. It's a confusing behavior, and I think we have an opportunity here to clarify it.

The reason I asked for tests is because this API is different in that it uses the json content type instead of url encoding the parameters. Verifying that the new functions create the expected http request would be good.

I'm not an approver for this project however. ;)

Primetalk:
@armsnyder if you could suggest a data structure for the message, that would be great. I didn't see in Slack exact specification of messages that are accepted at `responseURL`.

james-lawrence:
Thanks for your PR, I'm going to merge in a different way for handling the urls as it extends the current SendMessage API to support the feature instead of creating new standalone methods.

Feel free to take a look and let me know if it doesn't support parts of your usecase.

PR #573, opened 2019-08-23
Remove no longer populated items from Info struct

ghost:
Slack is no longer sending this information on connect and it causes user confusion.

See #532

### DIFF
--- a/info.go
+++ b/info.go
@@ -156,17 +156,12 @@ type Icons struct {
 	Image72 string `json:"image_72,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


james-lawrence (approved):
LGTM.

in response to your question regarding API compatibility:
actually I believe returning nil is the wrong behavior here. its actually better in this
case to break users of the library as the behavior has changed explicitly and they can no longer rely on it.

I also believe this code interacts with RTM connect (vs start) and can still be populated by that method.
however i'm still going to accept this PR as is due to connect's deprecation by slack and i think this is a good cleanup.

thanks.

PR #574, opened 2019-08-24
Make filename optional for uploading files

mzjp2:
To make the library mirror the API more closely, this removes the filename being a mandatory parameter, _unless_ the primary method of uploading the file uses the Reader field, in which case we throw an error.

This was mentioned in https://github.com/nlopes/slack/issues/568

Filename tests pass without changing them and this also makes the example code in `examples/files.go` consistent with the library, since the example code doesn't specify a Filename (and so running the example file results in an error).

### DIFF
--- a/files.go
+++ b/files.go
@@ -288,9 +288,6 @@ func (api *Client) UploadFileContext(ctx context.Context, params FileUploadParam
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


james-lawrence (approved):
i'd still love a test for this case to prevent it from occurring in the future. technically examples should be run as tests by go.... I wonder why they are not.

PR #575, opened 2019-08-24
Remove omitempty on SelectBlockElement.MinQueryLength

olivierlemasle:
In new Block API, default MinQueryLength is 3. It seems not to be documented; however, it was confirmed to me by Slack support. It means that by default, for an `external_select`, the query to populate the select menu will be sent by Slack only if the user has typed at least 3 characters.

To change this value to 0 (in order to populate the select menu even if no character is typed), setting MinQueryLength = 0 does not work currently because 0 is Go default value for int and is omitted from JSON.

This change removes "omitempty" which means that setting `MinQueryLength = 0` adds `"min_query_length": 0` in JSON serialization.

### DIFF
--- a/block_element.go
+++ b/block_element.go
@@ -148,7 +148,7 @@ type SelectBlockElement struct {
 	InitialUser         string                    `json:"initial_user,omitempty"`
 	InitialConversation string                    `json:"initial_conversation,omitempty"`
 	InitialChannel      string                    `json:"initial_channel,omitempty"`
-	MinQueryLength      int                       `json:"min_query_length,omitempty"`
+	MinQueryLength      int                       `json:"min_query_length"` // Slack defaults to 3.
 	Confirm             *ConfirmationBlockObject  `json:"confirm,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


block_element.go
```diff
@@ -148,7 +148,7 @@ type SelectBlockElement struct {
 	InitialUser         string                    `json:"initial_user,omitempty"`
 	InitialConversation string                    `json:"initial_conversation,omitempty"`
 	InitialChannel      string                    `json:"initial_channel,omitempty"`
-	MinQueryLength      int                       `json:"min_query_length,omitempty"`
+	MinQueryLength      int                       `json:"min_query_length"` // Optional. Default value is 3.
```
james-lawrence:
```suggestion
	MinQueryLength      int                       `json:"min_query_length"` // Optional. Slack defaults to 3.
```

james-lawrence (commented):
only problem I see with this is since we don't mimic slacks default we unconditionally send 0, which may negatively impact slacks systems.

i'd feel more comfortable with setting the value to 3 and sending 3 everytime instead.

otherwise this looks good.

block_element.go
olivierlemasle: Ok, done.

olivierlemasle:
Issue already fixed + conflict. Closing!

PR #582, opened 2019-08-31
Add pagination support to ListStars

atreya2011:
You were right when you said that `ListStarsPaginationContext` isn't actually paginated so I changed the function name to `ListAllStarsContext`. Thank you for taking your time in reviewing my code.

### DIFF
--- a/stars.go
+++ b/stars.go
@@ -158,3 +158,96 @@ func (api *Client) GetStarredContext(ctx context.Context, params StarsParameters
 	}
 	return starredItems, paging, nil
 }
+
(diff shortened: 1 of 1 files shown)


james-lawrence (approved):
looks good, will likely need followup for rate limiting but not a concern to get it in the app. thanks!

atreya2011:
I will do the followup on rate limiting. If I am not mistaken, it is 15 requests per second, so 15 pages per second for the `ListAllStars` method.

atreya2011:
Thanks for merging my PR!

PR #583, opened 2019-08-31
Add message buttons example

mzjp2:
Add example buttons file. Not sure if I should have included the HandleFunc stuff, but I initially struggled with parsing the payload, so though it'd be useful to someone else. Let me know if the example should be somewhere else, or a different filename or anything and I'll be happy to make changes.

From issue https://github.com/nlopes/slack/issues/578


james-lawrence:
fine as is. =) thanks.

james-lawrence (approved):

Issue #559, opened 2019-07-19
Dynamically creating contexts and blocks results in bad JSON


rifelpet:
On the current master branch I'm seeing messages get rejected due to incorrect JSON. We need to add a dynamic number of section blocks as well as text blocks in a context block:

```go
contextElements := make([]slack.MixedElement, 0)
contextElements = append(contextElements, slack.NewTextBlockObject(slack.MarkdownType, "*hello*", false, false))
contextElements = append(contextElements, slack.NewTextBlockObject(slack.MarkdownType, "*world*", false, false))
context := slack.NewContextBlock("context", contextElements...)

blocks = make([]slack.Block, 0)
blocks = append(blocks, slack.NewSectionBlock(slack.NewTextBlockObject(slack.MarkdownType, "foobar", false, false), nil, nil))
blocks = append(blocks, *context)

_, _, err := client.PostMessage(channel, slack.MsgOptionBlocks(blocks...), slack.MsgOptionText(message, false))
```

This is generating the following `blocks` json:
```json
[
  {
    "type": "section",
    "text": {
      "type": "mrkdwn",
      "text": "foobar"
    }
  },
  {
    "type": "context",
    "block_id": "context",
    "elements": {
      "Elements": [
        {
          "type": "mrkdwn",
          "text": "*hello*"
        },
        {
          "type": "mrkdwn",
          "text": "*world*"
        }
      ]
    }
  }
]
```

and logging this error:

`parseResponseBody {"ok":false,"error":"invalid_blocks","response_metadata":{"messages":["[ERROR] must provide an object [json-pointer:\/blocks\/1\/elements\/0]"]}}`

[...]

PR #584, opened 2019-08-31
Add test for posting a message with blocks

armsnyder:
Adding a new test. No changes to library made. Test was asked for in #559.


james-lawrence:
hero. thanks.

PR #587, opened 2019-09-09
Add omitempty to OptionBlockObject.URL, fixes invalid_blocks for select menus

lhallam:
The URL field is only valid for overflow menus: https://api.slack.com/reference/messaging/composition-objects#option


james-lawrence:
ha apparently a often encountered issue right here =p thanks.

PR #588, opened 2019-09-11
add blocks attribute to attachments, too

novas0x2a:
blocks can be embedded within attachments, to work around the lack of
colorbar support in blocks, or as a migration path to blocks

- https://api.slack.com/messaging/attachments-to-blocks#finding_equivalents_for_older_visual_elements
- https://api.slack.com/reference/messaging/attachments#fields

This also extends the existing postMessage test to include an attachment
with a block

### DIFF
--- a/attachments.go
+++ b/attachments.go
@@ -61,7 +61,7 @@ type ConfirmationField struct {
 // Attachment contains all the information for an attachment
 type Attachment struct {
 	Color    string `json:"color,omitempty"`
-	Fallback string `json:"fallback"`
+	Fallback string `json:"fallback,omitempty"`
 
 	CallbackID string `json:"callback_id,omitempty"`
 	ID         int    `json:"id,omitempty"`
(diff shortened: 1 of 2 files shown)


0gajun:
Thanks for good PR! I wanna use colorbar in blocks, too 😉 

When trying to post message with blocks in attachments using this PR, I got an error from slack api, `invalid_attachments`.

When using blocks in attachments, it seems that attachment shouldn't contain legacy fields.

This is a sample commit to fix the issue. If this looks good to you, please feel free to cherry-pick :)


https://github.com/0gajun/slack/commit/b9a497361026ae11ec8856765066508b92e7057a

novas0x2a:
Good catch, I had that in my local tree but apparently I forgot to push it. (I also fixed the test for that).

james-lawrence (approved):
LGTM thanks.

PR #589, opened 2019-09-15
Add fields to CtxChannel to get search API's response

suzuki-shunsuke:
See the official reference's Response section.

* https://api.slack.com/methods/search.all
* https://api.slack.com/methods/search.messages

```json
"channel": {
  "id": "C12345678",
  "is_ext_shared": false,
  "is_mpim": false,
  "is_org_shared": false,
  "is_pending_ext_shared": false,
  "is_private": false,
  "is_shared": false,
  "name": "general",
  "pending_shared": []
}
```

In this commit `"pending_shared"` is ignored because I don't know the data structure.

##### Pull Request Guidelines

##### API changes

### DIFF
--- a/search.go
+++ b/search.go
@@ -23,8 +23,14 @@ type SearchParameters struct {
 }
 
 type CtxChannel struct {
-	ID   string `json:"id"`
-	Name string `json:"name"`
+	ID                 string `json:"id"`
+	Name               string `json:"name"`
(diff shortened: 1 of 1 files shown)


ZainPatelQB:
Looks this PR: https://github.com/nlopes/slack/pull/592 has pending_shared.

james-lawrence (approved):
LGTM thanks.

PR #592, opened 2019-09-16
Add fields to search.messages response struct

### LINKED ISSUE #591: search.messages response is missing fields
The response to the search.messages API function is missing some fields.

The `CtxChannel` object only has two: `id` and `name`, but the real API response includes much more which we should add:

```
"channel": {
  "id": "C12345678",
  "is_ext_shared": false,
  "is_mpim": false,
  "is_org_shared": false,
  "is_pending_ext_shared": false,
  "is_private": false,
  "is_shared": false,
  "name": "general",
  "pending_shared": []
}
```

https://api.slack.com/methods/search.messages

armsnyder:
Add some fields to the response struct which were missing. Test is based on the search.messages documentation here: https://api.slack.com/methods/search.messages

Closes #591

### DIFF
--- a/search.go
+++ b/search.go
@@ -23,8 +23,15 @@ type SearchParameters struct {
 }
 
 type CtxChannel struct {
-	ID   string `json:"id"`
-	Name string `json:"name"`
+	ID                 string        `json:"id"`
+	Name               string        `json:"name"`
+	IsExtShared        bool          `json:"is_ext_shared"`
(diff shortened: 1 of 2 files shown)


kanata2:
@armsnyder Sorry for slow response. Please resolve conflict 🙏

armsnyder:
@kanata2 No worries. It looks like all of these fields except pending_shared have been added in master, so I'm going to close this PR.

PR #593, opened 2019-09-16
adds symmetry to json (un)marshaling of InteractionCallbacks

samolds:
this will allow json.Marshal to generate output that can then be
json.Unmarshaled for InteractionCallbacks.


james-lawrence (approved):
we should rethink this. but lg for now.

PR #598, opened 2019-09-29
Send blocks on SendMessage when MsgOptionResponseURL applied

pavlo-v-chernykh:
Found that blocks do not passed to responseURLSender through
sendConfig. Fixed implementation by adding separate field
for blocks to responseURLSender and sendConfig.


james-lawrence (approved):
LGTM, thanks.

PR #601, opened 2019-10-06
Add Unread property to History

acaloiaro:
This adds support for retrieving unread message counts with `conversations.history` and `channels.history`. 

I've contacted Slack support about the missing `Unreads` parameter in their [conversations.history documentation](https://api.slack.com/methods/conversations.history), but if you go to [their tester](https://api.slack.com/methods/conversations.history/test) and add `unreads=1` in the "extra args" field, you can see that it successfully returns message counts just like [conversations.history](https://api.slack.com/methods/channels.history/test). 

##### Pull Request Guidelines

##### API changes

### DIFF
--- a/conversation.go
+++ b/conversation.go
@@ -578,7 +578,7 @@ type GetConversationHistoryResponse struct {
 	HasMore            bool   `json:"has_more"`
 	PinCount           int    `json:"pin_count"`
 	Latest             string `json:"latest"`
-	UnreadCountDisplay int    `json:"unread_count_display"`
+	Unread int    `json:"unread_count_display"`
 	ResponseMetaData   struct {
 		NextCursor string `json:"next_cursor"`
 	} `json:"response_metadata"`
--- a/conversation.go
+++ b/conversation.go
@@ -570,16 +570,14 @@ type GetConversationHistoryParameters struct {
 	Latest    string
 	Limit     int
 	Oldest    string
-	Unreads   bool
 }
 
 type GetConversationHistoryResponse struct {
 	SlackResponse
-	HasMore            bool   `json:"has_more"`
-	PinCount           int    `json:"pin_count"`
-	Latest             string `json:"latest"`
-	Unread int    `json:"unread_count_display"`
-	ResponseMetaData   struct {
+	HasMore          bool   `json:"has_more"`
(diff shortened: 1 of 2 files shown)


history.go
```diff
@@ -19,9 +19,10 @@ type HistoryParameters struct {
 
 // History contains message history information needed to navigate a Channel / Group / DM history
 type History struct {
-	Latest   string    `json:"latest"`
-	Messages []Message `json:"messages"`
-	HasMore  bool      `json:"has_more"`
+	Latest             string    `json:"latest"`
+	Messages           []Message `json:"messages"`
+	HasMore            bool      `json:"has_more"`
+	UnreadCountDisplay int       `json:"unread_count_display"`
```
james-lawrence:
```suggestion
	Unread int       `json:"unread_count_display"`
```

conversation.go
```diff
@@ -570,14 +570,16 @@ type GetConversationHistoryParameters struct {
...
 type GetConversationHistoryResponse struct {
 	SlackResponse
-	HasMore          bool   `json:"has_more"`
-	PinCount         int    `json:"pin_count"`
-	Latest           string `json:"latest"`
-	ResponseMetaData struct {
+	HasMore            bool   `json:"has_more"`
+	PinCount           int    `json:"pin_count"`
+	Latest             string `json:"latest"`
+	UnreadCountDisplay int    `json:"unread_count_display"`
```
james-lawrence:
```suggestion
	Unread int    `json:"unread_count_display"`
```

james-lawrence (commented):
minor thing on naming.

acaloiaro:
I've removed support for unread count from `conversations.history`, in this PR because Slack doesn't officially support it. 

Response from Slack's support team
> The unreads parameter is indeed not officially supported by the conversations.history endpoint, and we do not recommend using it.


However, I've retained `Unread` on the `History` struct so it's available for other API requests such as `channel.history`.

acaloiaro:
@ollieparsley would you be interested in merging this? I'd like to move  https://github.com/acaloiaro/slack-term to `go-slack/slack`, and this is the last thing I depend on in my fork that hasn't been merged upstream at this point.

ollieparsley:
Merged! I'll do a release shortly

PR #605, opened 2019-10-15
Signal for read loop teardown on close.

### LINKED ISSUE #604: `goodbye` handling causes connect-disconnect loop due to incorrect signalling
As part of the cleanup in #492, I believe another race condition was introduced in the disconnect logic, particularly in the interaction between the [handling of Slack's `goodbye` message][1] and the [teardown of the `receiveIncomingEvent` loop][2].

It seems that the only way for the read loop to be torn down is from a recv on `rtm.disconnected`, or an error from `ReadJSON` which triggers a send on `killChannel`. Since a `goodbye` triggers a send on `killChannel`, and the subsequent effect of `killConnection` is to close the socket being polled by `ReadJSON`, the end effect is that two `false` values are sent to the `killChannel`: the first from the [`goodbye` handling][1] and the second from [`receiveIncomingEvent`][3]. However, the second send always blocks waiting for a recv, since the normal place `killChannel` is polled is in the [`handleEvents` flow][3], which has already exited due to handling a `goodbye`. What makes this bug so insidious, though, is the fact that this connect-disconnect loop does eventually end, and a normal connection can be established.

After 10 tries (each connect-disconnect loop takes about a second to complete), Slack begins ratelimiting the connection. At this point, we are saved by the [recv from `killChannel`][4] as part of the `RTM.connect` loop logic, which completely drains the `killChannel` (and also causes us to disregard the backoff cooldown time completely, triggering an immediate reconnect despite being ratelimited), and allows a normal connection to begin.

What you'll see in the debug log is something like this:
```
2019/10/12 23:35:42 websocket_managed_conn.go:369: Incoming Event: {"type": "goodbye", "source": "gateway_server"}
2019/10/12 23:35:42 websocket_managed_conn.go:211: killing connection
2019/10/12 23:35:42 websocket_managed_conn.go:173: Connecting to RTM
2019/10/12 23:35:42 rtm.go:72: Using URL: [redacted]
2019/10/12 23:35:42 websocket_managed_conn.go:189: Dialing to websocket on url [redacted]
2019/10/12 23:35:42 websocket_managed_conn.go:59: RTM connection succeeded on try 1
2019/10/12 23:35:42 websocket_managed_conn.go:211: killing connection
2019/10/12 23:35:42 websocket_managed_conn.go:173: Connecting to RTM
2019/10/12 23:35:42 rtm.go:72: Using URL: [redacted]
2019/10/12 23:35:42 websocket_managed_conn.go:189: Dialing to websocket on url [redacted]
2019/10/12 23:35:42 websocket_managed_conn.go:59: RTM connection succeeded on try 2
2019/10/12 23:35:42 websocket_managed_conn.go:211: killing connection
2019/10/12 23:35:42 websocket_managed_conn.go:173: Connecting to RTM
...
2019/10/12 23:35:44 websocket_managed_conn.go:173: Connecting to RTM
2019/10/12 23:35:44 rtm.go:68: Failed to connect to RTM: slack rate limit exceeded, retry after 58s
2019/10/12 23:35:44 websocket_managed_conn.go:177: Failed to start or connect to RTM: slack rate limit exceeded, retry after 58s
2019/10/12 23:35:44 websocket_managed_conn.go:143: reconnection 1 failed: slack rate limit exceeded, retry after 58s reconnecting in 58s
2019/10/12 23:35:44 websocket_managed_conn.go:173: Connecting to RTM
2019/10/12 23:35:44 rtm.go:68: Failed to connect to RTM: slack rate limit exceeded, retry after 58s
2019/10/12 23:35:44 websocket_managed_conn.go:177: Failed to start or connect to RTM: slack rate limit exceeded, retry after 58s
2019/10/12 23:35:44 websocket_managed_conn.go:143: reconnection 2 failed: slack rate limit exceeded, retry after 58s reconnecting in 58s
2019/10/12 23:36:42 websocket_managed_conn.go:173: Connecting to RTM
2019/10/12 23:36:42 rtm.go:72: Using URL: [redacted]
2019/10/12 23:36:42 websocket_managed_conn.go:189: Dialing to websocket on url [redacted]
2019/10/12 23:36:42 websocket_managed_conn.go:59: RTM connection succeeded on try 11
2019/10/12 23:36:42 websocket_managed_conn.go:369: Incoming Event: {"type": "hello"}
```

I believe the best course of action to fix this would be to add a mechanism to either preemptively tear down the `receiveIncomingEvent` loop at the same time as receiving a `goodbye` or signal that we are already in the process of tearing down the RTM connection. We can either 1) have `receiveIncomingEvent` tear itself down when it reads a `goodbye` off the socket, or 2) signal ongoing teardown by doing something like passing it a unique channel (per-connection, per-attempt) that can be closed by `killConnection` so we can ignore errors during teardown (or immediately return).

I can open up a PR but I'd like to hear your feedback before deciding on an approach (particularly from @james-lawrence).

Thanks!

[1]: https://github.com/nlopes/slack/blob/master/websocket_managed_conn.go#L261-L263
[2]: https://github.com/nlopes/slack/blob/master/websocket_managed_conn.go#L353-L365
[3]: https://github.com/nlopes/slack/blob/master/websocket_managed_conn.go#L237
[4]: https://github.com/nlopes/slack/blob/master/websocket_managed_conn.go#L150-L154

nolanlum:
This prevents errors from the previous read loop from propagating to the next connection attempt and triggering a disconnect. This is just one possible fix (though I've tested it in my environment and it works perfectly), and I'm willing to adopt it to match convention as necessary. Fixes #604

### DIFF
--- a/websocket_managed_conn.go
+++ b/websocket_managed_conn.go
@@ -59,10 +59,14 @@ func (rtm *RTM) ManageConnection() {
 		rtm.Debugf("RTM connection succeeded on try %d", connectionCount)
 
 		// we're now connected so we can set up listeners
-		go rtm.handleIncomingEvents()
+		shutdown := make(chan interface{})
+		go rtm.handleIncomingEvents(shutdown)
 
 		// this should be a blocking call until the connection has ended
 		rtm.handleEvents()
+
+		// signal listeners to teardown
+		close(shutdown)
 
 		select {
 		case <-rtm.disconnected:
@@ -272,9 +276,9 @@ func (rtm *RTM) handleEvents() {
 //
 // This will stop executing once the RTM's keepRunning channel has been closed
 // or has anything sent to it.
-func (rtm *RTM) handleIncomingEvents() {
+func (rtm *RTM) handleIncomingEvents(shutdown <-chan interface{}) {
 	for {
-		if err := rtm.receiveIncomingEvent(); err != nil {
+		if err := rtm.receiveIncomingEvent(shutdown); err != nil {
 			return
 		}
 	}
@@ -335,8 +339,14 @@ func (rtm *RTM) ping() error {
 
 // receiveIncomingEvent attempts to receive an event from the RTM's websocket.
 // This will block until a frame is available from the websocket.
-// If the read from the websocket results in a fatal error, this function will return non-nil.
-func (rtm *RTM) receiveIncomingEvent() error {
+// If the read from the websocket results in a fatal error, this function
+// will return non-nil.
+//
+// If `shutdown` is closed, we will break out of the read loop as soon as
+// able, and eat any event or error; since the incoming events channel is shared
+// between connection attempts, this prevents any lingering events or errors
+// from being processed by the following attempted connection.
+func (rtm *RTM) receiveIncomingEvent(shutdown <-chan interface{}) error {
 	event := json.RawMessage{}
 	err := rtm.conn.ReadJSON(&event)
 	switch {
@@ -349,6 +359,9 @@ func (rtm *RTM) receiveIncomingEvent() error {
 		select {
 		case rtm.forcePing <- true:
 		case <-rtm.disconnected:
+			return err
+		case <-shutdown:
+			return err
 		}
 	case err != nil:
 		// All other errors from ReadJSON come from NextReader, and should
@@ -360,6 +373,7 @@ func (rtm *RTM) receiveIncomingEvent() error {
 		select {
 		case rtm.killChannel <- false:
 		case <-rtm.disconnected:
+		case <-shutdown:
 		}
 
 		return err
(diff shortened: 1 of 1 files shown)


james-lawrence:
I'd love a test for this, so I can examine different approaches to the problem. trying to limit the number of things that need to be tracked when shutting down. this isn't a bad approach just want to think about it a bit

james-lawrence:
possible alternative solutions:
1. remove the goodby check and let ping handle eventual disconnect. (not ideal due to slowness)
2. pass the connection down into handle incoming events, part of the problem is that `receiveIncomingEvent` determines what connection to use from the RTM. (this has issues where it'll keep attempting to read a close conn potentially?)

regardless if you write a test for this, I'll accept the PR.

james-lawrence:
another potential avenue, pass in a raw events channel to both methods. removing it from the RTM struct. this can become your signaling channel.

james-lawrence:
closing in favor of the other pr. no need to add new signaling.

PR #608, opened 2019-10-21
Add RTM example that responds in different ways

veloek:
Demonstrating how to respond in channels vs. threads.

Ref #331


ollieparsley (approved):

ollieparsley:
Thank you @veloek

PR #610, opened 2019-10-22
Add BlockAction missing SelectedOptions field

ariefap:
Today I'm facing problem when retrieving `multi_static_select` components until realized that we missed `selected_options` to be parsed.


james-lawrence:
man slack is the worst, should have just gone with an array to begin with.

PR #618, opened 2019-10-30
fix unmarshallable `Message` from new WYSIWYG Slack feature

### LINKED ISSUE #616: Message event fails to parse new WYSIWYG messages
I was having all messages from only certain users get mysteriously dropped and did some digging. I'm using RTM.
Recently Slack changed how messages work: https://api.slack.com/changelog/2019-09-what-they-see-is-what-you-get-and-more-and-less
I'm not sure if this is some kind of A/B test right now but the WYSIWYG interface is only happening in the web client for me at the moment.
This library doesn't support the `rich_text` block so message events fail to parse.
Here's an example of a failing parse:
> Data: (*slack.UnmarshallingErrorEvent)(0xc00031adc0)(RTM Error: Could not unmarshall event "message": {"client_msg_id":"f6e92b76-00e7-xxxx-xxxx-56dd5761fa89","suppress_notification":false,"type":"message","text":"its over","user":"U0DCQ8UQ5","team":"T0D14XXXX","blocks":[{"type":"rich_text","block_id":"GoMVS","elements":[{"type":"rich_text_section","elements":[{"type":"text","text":"its over"}]}]}],"user_team":"T0D14XXXX","source_team":"T0D14XXXX","channel":"C8JDY8A3Y","event_ts":"1572008617.046900","ts":"1572008617.046900"})

Supporting the new blocks would be great. But because this Slack API change is mostly backwards compatible, I think we should ignore new block types instead of erroring out completely (as in here: https://github.com/nlopes/slack/blob/e5749f13b5af3c139165ab4180a95bb06a60128b/block_conv.go#L63-L64). It's likely that Slack will continue to add block types, and getting most of a message is better than failing to parse it whenever Slack changes their API.

rubensayshi:
Slack is rolling out their WYSIWYG chat messages on web and desktop, the result is that the RTM `Message`s we receive now contain a `rich_text` block which has significant list of new blocks that it be be composed of as you can see here; https://api.slack.com/changelog/2019-09-what-they-see-is-what-you-get-and-more-and-less

I suggest (and this PR implements just that), that we for now ignore the content of rich_text blocks until we can fully support it because atm we're unable to parse messages at all...  

Slack wrote in their blog;
> Once we release WYSIWYG editing, the text field found in message objects your app encounters will become an approximation of a user's more richly formatted message.

So we can continue using the `text` property just fine...

fixes https://github.com/nlopes/slack/issues/616

### DIFF
--- a/block_conv.go
+++ b/block_conv.go
@@ -60,6 +60,9 @@ func (b *Blocks) UnmarshalJSON(data []byte) error {
 			block = &ImageBlock{}
 		case "section":
 			block = &SectionBlock{}
+		case "rich_text":
+			// for now ignore the (complex) content of rich_text blocks until we can fully support it
+			continue
 		default:
 			return errors.New("unsupported block type")
 		}


nod:
I vote for this getting merged into the mainline.  Makes it easier for deploys, etc.

anatoliyfedorenko:
I support merging this since our code works 50% now in some slacks because of the changes related to WYSIWYG

rubensayshi:
ye our slackbot completely broke down :/

tobby17:
our slackbot broke down too😥

ArcticSnowman:
@james-lawrence  - this needs some urgency as Slack push this RTM update widely and most bots will stop working.

[edit] https://github.com/GrantStreetGroup/slack updated with this PR and replaced all references of nlopes.   Confirm that our bot is back and working..

jrnt30:
Thanks for the fix, can confirm this helped our deployment.

nod:
@nlopes Would you please merge this into master?  Many of us would rather not track a specific branch of the lib for slack bots.  Thanks!

SmilingNavern:
Yeah, thanks for this update. It would be great if there would be release for this fix. Because simple update of golang dependencies doesn't work.

SmilingNavern:
Currently you can update to latest nlopes/slack if you use modules like this: 
```
go get github.com/nlopes/slack@d06c2a2b3249b44a9c5dee8485f5a87497beb9ea
```
It works well.

PR #621, opened 2019-11-02
do not send multiple disconnect events

james-lawrence:
might be interrelated to: https://github.com/nlopes/slack/pull/605

### DIFF
--- a/websocket_managed_conn_test.go
+++ b/websocket_managed_conn_test.go
@@ -16,6 +16,35 @@ const (
 	testToken   = "TEST_TOKEN"
 )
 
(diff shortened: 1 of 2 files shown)


websocket_managed_conn_test.go
```diff
@@ -16,6 +16,35 @@ const (
 	testToken   = "TEST_TOKEN"
 )
 
+func TestRTMBeforeEvents(t *testing.T) {
```
james-lawrence, line 19: test is unrelated. but it test that disconnect will work even when manage connection is run after.

PR #622, opened 2019-11-02
simplify rtm

james-lawrence:
remove raw events from RTM. simplifying the code.


james-lawrence:
@nolanlum this might help with the issue, though I suspect not.

PR #633, opened 2019-11-11
Fix typo in SendAuthRevokeContext doc string

houqp:
Just a simple doc fix.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -27,7 +27,7 @@ func (api *Client) SendAuthRevoke(token string) (*AuthRevokeResponse, error) {
 	return api.SendAuthRevokeContext(context.Background(), token)
 }
 
-// SendAuthRevokeContext will retrieve the satus from api.test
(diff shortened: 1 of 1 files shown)


auth.go
```diff
@@ -27,7 +27,7 @@ func (api *Client) SendAuthRevoke(token string) (*AuthRevokeResponse, error) {
 	return api.SendAuthRevokeContext(context.Background(), token)
 }
 
-// SendAuthRevokeContext will retrieve the satus from api.test
+// SendAuthRevokeContext will retrieve the satus from api.revoke
```
kanata2: I think this comment is wrong in itself... 🤔
The `SendAuthRevokeContext` method sends a revocation request for the token with context.

houqp: Sorry, missed my local change :P I have pushed an updated version.

ollieparsley:
Thank you :)

PR #638, opened 2019-11-29
Bugfix: fix json unmarshal error for blocks in message attachments

malkhamis:
These changes are needed so we can unmarshal interaction-callback payloads, which are sent from Slack if the user clicks a button, select items etc.., into `slack.InteractionCallback` type. The payload may contain the original message with original blocks included in the `attachments` part of that payload's message.

Attempting to unmarshal such payloads into `slack.InteractionCallback` results in the following error:
```
json: cannot unmarshal object into Go struct field Attachment.message.attachments.blocks of type slack.Block
```

This is because field `Blocks` in type `Attachment` is of type `Block`, which is an interface and we cannot unmarshal things into interfaces. [Apparently](https://github.com/nlopes/slack/blob/d06c2a2b3249b44a9c5dee8485f5a87497beb9ea/messages.go#L98), field `Msg.Blocks` is of type `Blocks` rather than `[]Block`, but that is not the case with `Attachment.Blocks`.

For your convenience, I made a code snippet if you would like to do local testing, which can be found in the [Go Playground](https://play.golang.org/p/kK5RFWpSGje). Note that this snippet won't fail in the Go Playground since the unmarshal operation succeeds in the 0.6,0 release, but testing against `nlopes/slack@master` returns the aforementioned error.


kendung:
When are we going to merge this PR to master?

SophisticaSean:
Can we please merge this? Conversations API is super busted until this is fixed.

gvirtanen:
@james-lawrence Are you able to help get this merged?

bitsofinfo:
Please! can this be merged?

PR #648, opened 2020-01-10
fix channel_rename events: timestamps are ints, not strings

jschauma:
Without this, any channel_rename event will trigger an RTM error:

```{unmarshalling_error RTM Error: Could not unmarshall event "channel_rename": {"type":"channel_rename","channel":{"id":"CS7HLV9A5","is_channel":true,"name":"jt100","name_normalized":"jt100","created":1578681371},"event_ts":"1578682060.068000"}```


ollieparsley:
@jschauma thanks for the fix. I'll close this as it was also fixed in this: https://github.com/slack-go/slack/pull/672/files#diff-4efd52dcdbf2f01ebef1b6737a4c7011R48

PR #649, opened 2020-01-14
implement simple attachments parsing in MessageEvent

sawman:
##### Pull Request Guidelines

##### API changes


sawman:
I would ideally like to do a more thorough refactoring to separate the types into another package, so the common types can be used across RTM/Webhooks/Events, but I don't have the luxury of availability currently 😢 . This is to just add fields that are easy to parse, and they happen to be sufficient for the problem I am trying to solve.

ollieparsley:
@sawman Apologies for being slow to get to your PR. Could you refresh your branch from master and resolve the `go.sum` conflict. Then can you run `make pr-prep` to run the linter, formatter and tests please?

kanata2:
It seems that a lot of time has passed. Please feel free to re-open this!

PR #650, opened 2020-01-20
Add input block and plain-text input element

jojo43:
### What this PR does

- Add [input block](https://api.slack.com/reference/block-kit/blocks#input)
- Add [plain-text input element](https://api.slack.com/reference/block-kit/block-elements#input)

### Note

Currently, the modal is not supported in master branch.
It will be available after #632 is merged.

### DIFF
--- a/block.go
+++ b/block.go
@@ -14,6 +14,7 @@ const (
 	MBTImage   MessageBlockType = "image"
 	MBTAction  MessageBlockType = "actions"
 	MBTContext MessageBlockType = "context"
+	MBTInput   MessageBlockType = "input"
 )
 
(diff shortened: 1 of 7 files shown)


jojo43:
Hi @james-lawrence !
Could you check this?

createchange:
Any idea on when this will be merged? Looking to make use of this functionality in a simple app that is in-progress. Don't want to make further moves forward with dialogs if this is right on the horizon.

Thanks for your hard work!

jojo43:
I'm in the similar situation too, but have no idea...
I hope someone in the maintainer will see this.

PR #661, opened 2020-02-17
MsgOptionResponseURL: make argument name easy

nu50218:
The argument name `rt` is not easy to understand as `responseType` (at least for me).

could you update it to `responseType` or add comment?


ollieparsley (approved):
Thanks for the PR @nu50218 :)

PR #668, opened 2020-02-28
Fix update view request

### LINKED ISSUE #667: UpdateView failed due to error "Cannot provide both `view_id` and `external_id`"
### Steps to reproduce

1. Call [`UpdateView`](https://github.com/slack-go/slack/blob/8f71a596d3982be7ec4abdaecce3d4aafb2eb0a5/views.go#L190) with valid arguments (only `externalID` is blank)
2. Receive error response like below

```json
{
    "ok": false,
    "error": "invalid_arguments",
    "warning": "missing_charset",
    "response_metadata": {
        "messages": [
            "[ERROR] Cannot provide both `view_id` and `external_id`"
        ],
        "warnings": [
            "missing_charset"
        ]
    }
}
```

jojo43:
### What this PR does

- [x] Add `omitempty` to optional fields

### Reference

- https://api.slack.com/methods/views.update
- fixes #667

### DIFF
--- a/views.go
+++ b/views.go
@@ -81,9 +81,9 @@ type pushViewRequest struct {
 
 type updateViewRequest struct {
 	View       ModalViewRequest `json:"view"`
-	ExternalID string           `json:"external_id"`
-	Hash       string           `json:"hash"`
-	ViewID     string           `json:"view_id"`
+	ExternalID string           `json:"external_id,omitempty"`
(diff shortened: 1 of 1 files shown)


kanata2 (approved):
Thanks! Looks good to me :)
(It seems that a part of [my changes](https://github.com/slack-go/slack/pull/632) was lost without knowing ... 🤔 )

PR #670, opened 2020-02-28
Add example of dialog

Donngi:
Hi!
I added an example of dialog.
I was troubled when I implemented it myself for the first time, so I wrote a sample code.

There are many examples of other features and they are heavily helpful. I think this sample will help users as well!

### DIFF
--- a/examples/dialog/dialog.go
+++ b/examples/dialog/dialog.go
@@ -1,0 +1,98 @@
+package main
+
+import (
+	"encoding/json"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


ollieparsley:
@nicoJN Thank you very mich to taking the time to create the example :)

Donngi:
Thank you! 
This is my first PR of oss in my life. I’m really happy to commit this great oss!

kanata2:
Congrats!

PR #674, opened 2020-03-02
removes unused fields from modal view requests

samolds:
these fields are optional, but if they're not included the api requests fail

### DIFF
--- a/views.go
+++ b/views.go
@@ -38,8 +38,8 @@ type ModalViewRequest struct {
 	Type            ViewType         `json:"type"`
 	Title           *TextBlockObject `json:"title"`
 	Blocks          Blocks           `json:"blocks"`
-	Close           *TextBlockObject `json:"close"`
-	Submit          *TextBlockObject `json:"submit"`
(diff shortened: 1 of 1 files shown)


block_element.go
```diff
@@ -244,7 +244,7 @@ func NewDatePickerBlockElement(actionID string) *DatePickerBlockElement {
 // More Information: https://api.slack.com/reference/messaging/block-elements#input
 type PlainTextInputBlockElement struct {
 	Type         MessageElementType `json:"type"`
-	ActionID     string             `json:"action_id"`
+	ActionID     string             `json:"action_id,omitempty"`
```
kanata2: According to https://api.slack.com/reference/block-kit/block-elements#input, `action_id` is required.

ollieparsley: @samolds did you see @kanata2's comment? ^

samolds: oops, i missed this. thanks for looking out!

samolds:
This has already been resolved by dea0811c 🙌Closing this PR

PR #680, opened 2020-03-10
Unmarshal multi select input block

jojo43:
Multi-select menus are also available in input blocks.
https://api.slack.com/reference/block-kit/block-elements#multi_select


usmonster:
Related: #690.

ollieparsley (approved):

ollieparsley:
@jojo43 thanks for the contribution :)

PR #685, opened 2020-03-27
Add support for checkboxes

eliot-stripe:
I added blockkit support for checkbox groups. You can find the documentation at https://api.slack.com/reference/block-kit/block-elements#checkboxes

I manually tested this by creating a modal that had checkboxes. I wasn't sure if there was a way to add automated tests for these types of definitions, but I'm definitely open to suggestions.

### DIFF
--- a/block_element.go
+++ b/block_element.go
@@ -3,6 +3,7 @@ package slack
 // https://api.slack.com/reference/messaging/block-elements
 
 const (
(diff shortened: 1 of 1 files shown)


ollieparsley (approved):
@eliot-stripe Thanks for the contribution. It's more difficult to automate the tests for these. So I'll approve and merge. Then put out an issue for automated testing of block elements.

PR #687, opened 2020-03-30
Removed unused dependency

mjwhitta:
##### PR preparation

Done, but some fail. I verified they failed before-hand, so not my doing.

I almost made this an issue, but decided it was more polite to do and ask for merge rather than say "hey, can you do this?". If you'd rather it be an issue, feel free to close this request. I won't be offended.


ollieparsley (approved):
@mjwhitta thank you for the fix and for taking the time to actually make the change

PR #689, opened 2020-04-03
Add missing omitempty for `client_msg_id`

DanielleB-R:
Unlike all the other fields in the `Msg` struct, the `ClientMsgID` field does not have `omitempty` set in its json tag. This PR adds it back so that the JSON generation is consistent.


ollieparsley (approved):
@DanielleB-R Thank you for the fix!

PR #694, opened 2020-04-09
Adding const for Interaction Type Shortcut

cynicaljoy:
With the latest release, we're now able to make use of [Shortcuts](https://api.slack.com/interactivity/shortcuts/using), adding the const def for new Type.


ollieparsley (approved):
@cynicaljoy Thanks for the contribution!

PR #698, opened 2020-04-12
Removed separate TeamID field from OAuth2 response.

mzduke:
Based on [official documentation](https://api.slack.com/methods/oauth.v2.access#response) and actual behavior OAuth2 response object has no TeamID field (only Team field with nested ID field).

It does not break something explicitly but disguiding during development.


ollieparsley (approved):
@mzduke Thank you for spotting the issue! I'll approve and merge it.

PR #699, opened 2020-04-14
Add support for groups.replies 

hkousha:
groups are being used for private channels.

Slack API: https://api.slack.com/methods/groups.replies

This is similar to how channel function is doing it.


ollieparsley (approved):
Hey @hkousha I'll merge this. Though it is worth noting that the API endpoint is being deprecated in Nov 2020 and apps created after June 2020 won't be able to use it.

hkousha:
yes saw that will use `conversation` moving forward. Thanks for the merge.

PR #700, opened 2020-04-16
handle unexpected integer ip from team.accessLogs

evrenios:
when i tried to use `GetAccessLogs` we've seen a panic 

`panic: json: cannot unmarshal number into Go struct field Login.logins.ip of type string`

when i tried to see the response i've seen that the API returns 0 for some cases and when tried to bind to a string field, it fails. 

Example response from slack api shown from REST API 

image

here is a link that mocks the situation; https://play.golang.org/p/KWf363iOnNx

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,7 @@
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/evrenios/slack v0.6.7 h1:Owaw4BdxojykM9pdyLHPo6MChqJq06yFon/28nRrcao=
+github.com/evrenios/slack v0.6.7/go.mod h1:QSRiBHK5IQLOF9dRDDbj+yLriccGPKa/Y6KqTS9OFAs=
 github.com/go-test/deep v1.0.4 h1:u2CU3YKy9I2pmu9pX0eq50wCgjfGIt539SqR7FbHiho=
 github.com/go-test/deep v1.0.4/go.mod h1:wGDj63lr65AM2AQyKZd/NYHGb0R+1RLqB8NKt3aSFNA=
 github.com/gorilla/websocket v1.2.0 h1:VJtLvh6VQym50czpZzx07z/kw9EgAxI3x1ZB8taTMQQ=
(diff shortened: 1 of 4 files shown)


ollieparsley (changes requested):
Hi @evrenios! Thank you very much for spotting the issue and fixing it. Could you convert the your code in play.golang.org into test in `team_test.go` please. Shouldn't take long as you've got the code ready on that link :). That way we can ensure we don't have any regressions if someone doesn't understand the issue in the future and tries to revert this change.

You can then run `make pr-prep` to check linting, formatting and ensure all tests pass.

evrenios:
added test cases @ollieparsley;
ran `make pr-prep` all tests passed

kanata2 (approved):

ollieparsley (approved):
Thanks @evrenios! :)

go.sum
```diff
@@ -1,5 +1,7 @@
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/evrenios/slack v0.6.7 h1:Owaw4BdxojykM9pdyLHPo6MChqJq06yFon/28nRrcao=
```
ollieparsley: Could you remove these references to your fork please. `go mod tidy` should do it.

ollieparsley (changes requested):

go.sum
evrenios: just ran in @ollieparsley

team.go
```diff
@@ -134,6 +136,20 @@ func (api *Client) GetAccessLogsContext(ctx context.Context, params AccessLogPar
...
+	for i, login := range response.Logins {
+		switch vv := login.IP.(type) {
+		case string:
+			response.Logins[i].RealIP = vv
+		default:
+			response.Logins[i].RealIP = "0.0.0.0"
+		}
+	}
+
+	for _, login := range response.Logins {
```
ollieparsley: Apologies for not spotting this in previous reviews! Was the printng here just for debugging?

ollieparsley (changes requested):

team.go
evrenios: ahh yes totally my bad; forgot to remove it; apologies for not removing in the first place : )

nevins-b:
This seems like it will cause issues for folks who want to look at the access logs directly. Would something like this be a nicer solution?

```
func (l *Login) UnmarshalJSON(data []byte) error {
	s := make(map[string]interface{})
	if err := json.Unmarshal(data, &s); err != nil {
		return err
	}
	l.UserID = s["user_id"].(string)
	l.Username = s["username"].(string)
	l.DateFirst = int(s["date_first"].(float64))
	l.DateLast = int(s["date_last"].(float64))
	l.Count = int(s["count"].(float64))
	l.UserAgent = s["user_agent"].(string)
	if v, ok := s["isp"]; ok && v != nil {
		l.ISP = v.(string)
	}
	if v, ok := s["country"]; ok && v != nil {
		l.Country = v.(string)
	}
	if v, ok := s["region"]; ok && v != nil {
		l.Region = v.(string)
	}
	if v, ok := s["ip"]; ok && v != nil {
		switch v.(type) {
		case string:
			l.IP = v.(string)
		default:
			l.IP = "0.0.0.0"
		}
	}
	return nil
}
```

kanata2:
@evrenios 
I am going to close this one for now. If you are still having similar problems, could you please submit the PR again? 🙏

PR #693, opened 2020-04-05
Add support for replacing and deleting original message with `ResponseURL`

frozenbonito:
I added new msg options for replacing and deleting original message with `ResponseURL`.

PR #701, opened 2020-04-18
Add delete original message option

alyosha:
## Summary
- [From what I can tell](https://api.slack.com/interactivity/handling#updating_message_response#deleting_message_response), `chat.delete` works for normal messages, but it is only possible to delete an ephemeral message by using `response_url`  
- The aim of this PR is to add a message option which can be used for ephemeral/non-ephemeral messages alike, but which is required when trying to remove the former (please point to any existing means of removing ephemeral messages if I am missing something)
- I have confirmed behavior on a personal project and added a test, please let me know if I can do anything else to get the PR ready for review :wave:


alyosha:
closing this as I see it is a dupe of #693 which also adds support for `ReplaceOriginal` :100:

@ollieparsley any chance we could get someone to take a look at #693? Not being able to delete/replace ephemeral messages is pretty big

ollieparsley:
The other PR is merged and will be included in the latest release

PR #703, opened 2020-04-19
Add go 1.14 tests to travis

ollieparsley:


ollieparsley:
There is an issue with the SingleRTMConnect test which always fails with 1.14, but no other versions. Hmmm

ghost:
https://github.com/slack-go/slack/blob/9d8728cd25ff74f4b98d18d994d1f8370f5608e4/slacktest/server_test.go#L51 needs a `* time.Second` like in all other tests.

ollieparsley:
@ttgmpsn thanks! I completely overlooked the units. Go 1.14 must have changed the implementation to not default to seconds if not specifying a unit of time.

PR #706, opened 2020-04-20
WithStyle fluent interface for ButtonBlockElement

jace-ys:
##### Pull Request Guidelines

##### PR preparation

##### API changes

##### Description

This adds a backwards-compatible fluent interface for ButtonBlockElement, mainly for convenience, so we can do in-line chaining like this:

```go
suggestButtonTextBlock := slack.NewTextBlockObject(slack.PlainTextType, "Suggest a question", false, false)
suggestButtonBlockEle := slack.NewButtonBlockElement("start.block.suggest_question", "", suggestButtonTextBlock).WithStyle(slack.StylePrimary)
suggestButtonBlock := slack.NewActionBlock("", suggestButtonBlockEle)
```

### DIFF
--- a/block_element.go
+++ b/block_element.go
@@ -152,9 +152,10 @@ func (s ButtonBlockElement) ElementType() MessageElementType {
 	return s.Type
 }
 
-// add styling to button object
-func (s *ButtonBlockElement) WithStyle(style Style) {
+// WithStyling adds styling to the button object and returns the modified ButtonBlockElement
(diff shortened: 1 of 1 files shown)


kanata2 (approved):

ollieparsley:
Thanks for the contribution @jace-ys. I'll get this into the next release.

ollieparsley (approved):

PR #707, opened 2020-04-21
Added Blocks struct to WebhookMessage

### LINKED ISSUE #594: Add `Blocks` to WebHooks
We are using the webhooks calls.  We would like to add in the ability to send the new `Block` Messages.   I will PR the update later this week. 

https://github.com/nlopes/slack/blob/50e524da56af7c1b27cc221baaa70cfe3d926804/webhooks.go#L11-L20

Dalabad:
Replaces [PR-595](https://github.com/slack-go/slack/pull/595) and adds the missing tests.
Closes [Issue-594](https://github.com/slack-go/slack/issues/594)

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/webhooks_test.go
+++ b/webhooks_test.go
@@ -70,8 +70,8 @@ func TestWebhookMessage_WithBlocks(t *testing.T) {
 	textBlockObject := NewTextBlockObject("plain_text", "text", false, false)
 	sectionBlock := NewSectionBlock(textBlockObject, nil, nil)
 
-	singleBlock := Blocks{BlockSet: []Block{sectionBlock}}
-	twoBlocks := Blocks{BlockSet: []Block{sectionBlock, sectionBlock}}
+	singleBlock := &Blocks{BlockSet: []Block{sectionBlock}}
+	twoBlocks := &Blocks{BlockSet: []Block{sectionBlock, sectionBlock}}
 
 	msgSingleBlock := WebhookMessage{Blocks: singleBlock}
 	assert.Equal(t, 1, len(msgSingleBlock.Blocks.BlockSet))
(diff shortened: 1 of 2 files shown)


webhooks_test.go
```diff
@@ -63,3 +65,27 @@ func TestPostWebhook_NotOK(t *testing.T) {
...
+
+	msgTwoBlocks := WebhookMessage{Blocks: twoBlocks}
+	assert.Equal(t, 2, len(msgTwoBlocks.Blocks.BlockSet))
+
+	msgJsonTwoBlocks, _ := json.Marshal(msgTwoBlocks)
+	assert.Equal(t, `{"blocks":[{"type":"section","text":{"type":"plain_text","text":"text"}},{"type":"section","text":{"type":"plain_text","text":"text"}}]}`, string(msgJsonTwoBlocks))
+
+	msgNoBlocks := WebhookMessage{Text: "foo"}
+	msgJsonNoBlocks, _ := json.Marshal(msgNoBlocks)
+	assert.Equal(t, `{"text":"foo","blocks":null}`, string(msgJsonNoBlocks))
```
kanata2: I think it is better to omit `blocks` property from encoded json if `blocks` is null.

Dalabad: The `blocks` property is now omitted if it is empty. Sadly the travis build crashed due to an error, which based on what I see is not related to the latest commit. Could you please restart the build?

kanata2 (approved):
Thanks! LGTM!

Phylu:
Is there anything missing here for it to be merged?

kanata2:
merged!

Phylu:
Thanks! :)

PR #709, opened 2020-04-23
Add SelectedUsers field to BlockAction

DanielleB-R:
The `users_multi_select` block element (https://api.slack.com/reference/block-kit/block-elements#users_multi_select) returns its results in the JSON key `selected_users` as an array of strings. This PR adds that key to the `BlockAction` struct so that we can use it to receive data from that input element.


DanielleB-R:
Superseded by #717

PR #710, opened 2020-04-26
Add `omitempty struct tag for Title field of ImageBlock

mzduke:
As per [API documentation](https://api.slack.com/reference/block-kit/blocks#image), the `title` of ImageBlock is not required.

In the current library implementation, an empty field of `*TextBlockObject` type leads to `null` in the encoded JSON representation. Slack API rejects such payload with an error. The proper behavior: do not send the `title` field if it's empty. where `omitempty`  struct tag helps.


kanata2 (approved):
Thanks!

PR #711, opened 2020-04-26
Added missed fields for Container struct

mzduke:
As per [API documentation](https://api.slack.com/reference/interaction-payloads/block-actions#examples), `container` subtype of interaction callback can contain additional fields like:

- message_ts
- attachment_id
- channel_id
- is_ephemeral
- is_app_unfurl

They are missed in the current version of the library.

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -59,8 +59,11 @@ type InteractionCallback struct {
 }
 
 type Container struct {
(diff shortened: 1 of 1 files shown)


kanata2 (approved):
Thanks!

PR #712, opened 2020-04-27
Add thread_ts to to LinkSharedEvent

smgt:
I was doing some unfurling and noticed that `thread_ts` was missing in the `LinkSharedEvent` struct. This is a little cumbersome if my bot wants to respond in the thread where the URL was mentioned.

https://api.slack.com/events/link_shared


kanata2 (approved):
Thank you!

Issue #714, opened 2020-05-01
In debug mode log http request body while talking to message api


sfzylad:
At the moment it is possible to initialise the Slack client with debug turned on:

`slack.New(TOKEN, slack.OptionDebug(true))`

This results in some of the requests to be logged. However it doesn't log the body of the requests that are sent to the message api. Having this information logged is very useful for debugging.

kanata2:
close by #715

PR #715, opened 2020-05-01
Add debug message in SendMessageContext.

sfzylad:
This PR implements the #714

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -208,13 +208,14 @@ func (api *Client) SendMessageContext(ctx context.Context, channelID string, opt
 		return "", "", "", err
 	}
 
-	reqBody, err := ioutil.ReadAll(req.Body)
-	if err != nil {
-		return "", "", "", err
+	if api.Debug() {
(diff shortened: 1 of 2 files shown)


chat.go
```diff
@@ -206,6 +208,14 @@ func (api *Client) SendMessageContext(ctx context.Context, channelID string, opt
 		return "", "", "", err
 	}
 
+	reqBody, err := ioutil.ReadAll(req.Body)
```
kanata2: I think it is better for debugging but it may affect performance by `iotuil.ReadAll`.
How about you use `if api.Debug() { ... }` ?

sfzylad: It makes perfect sense. Change is on the way! 👍

sfzylad: For the record, I realise the `api.Debugf()` checks if the debug is enabled but I think it's more readable if it's inside of the if. :)

kanata2 (approved):
Thank you :)

PR #717, opened 2020-05-02
Add missing omitempty to actionID of some elements. Add multi select components property to the BlockAction.

### LINKED ISSUE #713: The lost of the "selected_users" field when using modals with multi_users_select input blocks.
Hello,
 I noticed that there is a field lost problem when using a modal with “multi_users_select” input blocks.

Thanks for the past contribution #680, we can unmarshal multi select input block (Great thank you!).   
However when I receive a payload of a view.submission event, the input value field “selected_users” is lost.

Original payload (view.state.values).
```
"block-id": {
    "action-id": {
        "type": "multi_users_select",
        "selected_users": [
            "USER1",
            "USER2"
        ]
    }
},
```

Result of unmarshall (view.state.values)
```
"block-id": {
    "action-id": {
        "action_id": "",
        "block_id": "",
        "type": "multi_users_select",
        "text": {
            "type": "",
            "text": ""
        },
        "value": "",
        "action_ts": "",
        "selected_option": {
            "text": null,
            "value": ""
        },
        "selected_options": null,
        "selected_user": "",
        "selected_channel": "",
        "selected_conversation": "",
        "selected_date": "",
        "initial_option": {
            "text": null,
            "value": ""
        },
        "initial_user": "",
        "initial_channel": "",
        "initial_conversation": "",
        "initial_date": ""
    }
},
```

I tried fixing it but I couldn't understand a proper approach (I'm a beginner of Go ...).
Would you fix this problem?

Donngi:
I fixed two parts.

## 1. Add missing omitempty to actionID of some elements
I noticed that there are some Block Elements whose actionID doesn't have omitempty property.

- DatePickerBlockElement
- DatePickerBlockElement
- CheckboxGroupsBlockElement
- RadioButtonsBlockElement

Like any other Block Elements, "actionID" is not required. (When we create modal's json by [BlockKit Builder](https://api.slack.com/tools/block-kit-builder?mode=modal), actionID is not attached initially.)

## 2.Add  selected_xxxxxs params to the BlockAction struct
I noticed that there aren't some Multi-Select components properties.
I added the params below to the struct `BlockAction`.

- SelectedUsers 
- SelectedChannels
- SelectedConversations

This change will resolve https://github.com/slack-go/slack/issues/713 :+1:

===

In addition, as boyscout, I fix a mistake of a comment of "NewCheckboxGroupsBlockElement".

### DIFF
--- a/block.go
+++ b/block.go
@@ -49,11 +49,8 @@ type BlockAction struct {
 	SelectedDate          string              `json:"selected_date"`
 	InitialOption         OptionBlockObject   `json:"initial_option"`
 	InitialUser           string              `json:"initial_user"`
-	InitialUsers          []string            `json:"initial_users"`
 	InitialChannel        string              `json:"initial_channel"`
-	InitialChannels       []string            `json:"initial_channels"`
 	InitialConversation   string              `json:"initial_conversation"`
-	InitialConversations  []string            `json:"initial_conversations"`
 	InitialDate           string              `json:"initial_date"`
 }
 
--- a/block.go
+++ b/block.go
@@ -32,23 +32,29 @@ type Blocks struct {
 
 // BlockAction is the action callback sent when a block is interacted with
 type BlockAction struct {
-	ActionID             string              `json:"action_id"`
-	BlockID              string              `json:"block_id"`
-	Type                 actionType          `json:"type"`
-	Text                 TextBlockObject     `json:"text"`
-	Value                string              `json:"value"`
-	ActionTs             string              `json:"action_ts"`
-	SelectedOption       OptionBlockObject   `json:"selected_option"`
-	SelectedOptions      []OptionBlockObject `json:"selected_options"`
-	SelectedUser         string              `json:"selected_user"`
-	SelectedChannel      string              `json:"selected_channel"`
-	SelectedConversation string              `json:"selected_conversation"`
-	SelectedDate         string              `json:"selected_date"`
-	InitialOption        OptionBlockObject   `json:"initial_option"`
-	InitialUser          string              `json:"initial_user"`
-	InitialChannel       string              `json:"initial_channel"`
-	InitialConversation  string              `json:"initial_conversation"`
-	InitialDate          string              `json:"initial_date"`
+	ActionID              string              `json:"action_id"`
+	BlockID               string              `json:"block_id"`
+	Type                  actionType          `json:"type"`
+	Text                  TextBlockObject     `json:"text"`
+	Value                 string              `json:"value"`
+	ActionTs              string              `json:"action_ts"`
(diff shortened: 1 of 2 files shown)


kanata2 (changes requested):
As documentation, it seems that `action_id` is required. (ref: https://api.slack.com/reference/block-kit/block-elements#datepicker )
However, the documentation is sometimes outdated, so I want you to verify the actual behavior.

Donngi:
Hi, @kanata2 

I verified the behavior of the slack.

## Test item
I send a modal view request with the json below.

```json
{
	"type": "modal",
	"title": {
		"type": "plain_text",
		"text": "My App",
		"emoji": true
	},
	"submit": {
		"type": "plain_text",
		"text": "Submit",
		"emoji": true
	},
	"close": {
		"type": "plain_text",
		"text": "Cancel",
		"emoji": true
	},
	"blocks": [
		{
			"type": "input",
			"element": {
				"type": "plain_text_input"
			},
			"label": {
				"type": "plain_text",
				"text": "Singleline Text Input",
				"emoji": true
			}
		}
	]
}
```

Then I inputed text field and pushed a submit button.

## Test result
I got a json like below (Excerpt only some fields).

```
"view": {
        ...
        "blocks": [
            {
                "type": "input",
                "block_id": "Kzpss",
                "label": {
                    "type": "plain_text",
                    "text": "Singleline Text Input",
                    "emoji": true
                },
                "optional": false,
                "element": {
                    "type": "plain_text_input",
                    "action_id": "wCK"
                }
            }
        ],
        ...
        "state": {
            "values": {
                "Kzpss": {
                    "wCK": {
                        "type": "plain_text_input",
                        "value": "TEST"
                    }
                }
            }
        },
        ...
    },
```

### Conclusion
If we don't attach action_id and block_id to some elements, slack automatically add some IDs.

## Further discussion
I think the biggest problem is that It's difficult to analyze the problem for slack-go/slack users with the current specifications.

If we don't attach action_id, slack-go/slack automatically add action_id field with empty string.

```
action_id: ""
```

And then, If we send a modal with `action_id: ""` (func Openview), it returns an error.

```
err invalid_arguments
```

The root cause of the error is that the action_id field must be contain more than 0 characters.

image

Unfortunately, return values of func Openview don't contain this information. So users can see the error reason by only using Block Kit Builder. 
It makes it difficult to analyze the problem.

If we add `omitempty` to action_id field, users can successfully send a modal view and get input values like mentioned above, and can notice that they should add action_id field expressly to make  easy to parse the values.

I think it's a more user friendly behavior.

Donngi:
Hi, @kanata2 !
Is there anything else which I should validate? Feel free to offer me additional test👍

ollieparsley (approved):
Thank you @nicoJN

PR #718, opened 2020-05-04
Support for decoding "checkboxes" block element

mversluys:
See the "todo app" template example in the block kit builder https://api.slack.com/tools/block-kit-builder for an example of an action block that contains a checkbox group block element.

The library allows constructing an `ActionBlock` that contains a `CheckBoxGroupBlockElement`, but didn't handle decoding when Slack responds.

Updated error if the type is unsupported to include the name of the unsupported type to make it easier to track down similar errors in the future.

### DIFF
--- a/block_conv.go
+++ b/block_conv.go
@@ -2,6 +2,7 @@ package slack
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


Jonatan-McE:
I'm seeing the same thing. I have created a checkbox element in an input block for my View Modul which works fine. But when the users submits the form, my code complains that its not able to unmarshal the returnd json. Error message: "Could not parse action response JSON: unsupported block element type"

ollieparsley (approved):

ollieparsley:
Thank you @mversluys

PR #725, opened 2020-05-11
Add check in OpenView to verify that BlockID is unique for InputBlock's

sspaink:
The goal of this pull request is to validate input blocks before sending a request to be able to return a useful error message. 

I ran into this issue were I had duplicate block id's but the only error message I got back from Slack API was "invalid_arguments". The requirement that these block id's need to be unique is described in the slack documentation with the description ["unique identifier for a block"](https://api.slack.com/reference/block-kit/blocks) but I overlooked it. I thought maybe other people could run into this same issue because while block ID is optional in the slack API it is required when you use `slack.NewInputBlock("BlockID", label, element)`.

Hopefully this is an appropriate change, could it be considered a convenience method?
I made a unit test case for OpenView and ran `make pr-prep`

Thanks.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -9,6 +9,7 @@ const (
 	ErrRTMGoodbye           = errorsx.String("goodbye detected")
 	ErrRTMDeadman           = errorsx.String("deadman switch triggered")
 	ErrParametersMissing    = errorsx.String("received empty parameters")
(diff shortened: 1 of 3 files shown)


ollieparsley (approved):

ollieparsley:
Thank you @BattleBas

PR #727, opened 2020-05-12
Addition of BotID to AppMentionEvent

snarkyheelz:
If a bot triggers the `app_mention` event, a `bot_id` attribute is provided as part of the event.  This allows consumers of the event to distinguish between a user and a bot.

_Note: This seems to be undocumented in the Slack Events API; however, it is still present._

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,9 +1,11 @@
 module github.com/slack-go/slack
 
 require (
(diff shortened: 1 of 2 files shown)


ollieparsley (approved):
Thank you @vaporofnuance

PR #728, opened 2020-05-14
Fix potential security vuln in eventsapi example

bored-engineer:
Without `return` the handler will continue to process events even if `slackevents.ParseEvent` returns an error (such as an invalid verify token)


kanata2 (approved):
Good catch! Thanks!

PR #729, opened 2020-05-17
Added missed fields for Select block element struct

mzduke:
Based on official documentation SelectBlockElement struct has additional properties: `default_to_current_conversation` and `response_url_enabled` which is useful in conjunction with modals.

See for details: https://api.slack.com/reference/block-kit/block-elements#conversation_select


kanata2 (approved):
LGTM.
(I thought it might be good to use an omitempty bool value as a pointer.)

PR #734, opened 2020-05-22
Add Style to ConfirmationBlockObject

d15ky:
Added Style attribute and WithStyle func similar to ButtonBlockElement


ollieparsley (approved):
Thank you @d15ky

PR #738, opened 2020-06-01
Add support for unmarshalling of input blocks containing checkbox blo…

gdsmith7:
Unmarshalling an input block containing a checkbox block element currently returns this error `unsupported block element type`. 

I have added the checkboxes case to the custom unmarshaller for input blocks so this should no longer happen.


eliot-stripe:
Thanks for taking care of this.

kanata2 (approved):
Thanks :)

Jonatan-McE:
Thanks George. This was blocking me and this fixed my issue! Thanks alot!

PR #741, opened 2020-06-04
basic conversations.info endpoint for slacktest

gak:
Added a basic endpoint to slacktest for `conversations.info`.

It only returns an ID and Name. Name is a cropped version of the channel ID, removing the `C` prefix.


ollieparsley (approved):
Thank you @gak :)

PR #745, opened 2020-06-10
Add client_msg_id to MessageEvent

themmj:
A message event has a `client_msg_id` attribute to identify the message. The RTM `Msg` already contains it. I've added it to the EventsAPI `MessageEvent`.


kanata2 (approved):
LGTM :)

PR #755, opened 2020-07-05
Add SetUserRealName for change user's realName

felizi:
Added SetUserRealName to permit user's change the realName.


kanata2 (approved):
Thanks! (I would appreciate if you can set up other fields as well in a method.)

ollieparsley (approved):
Thank you @felizi :)

PR #756, opened 2020-07-07
Modal example with different kinds of fields

karl-johan-grahn:


ollieparsley (approved):
Thank you for this @karl-johan-grahn :)

PR #765, opened 2020-07-23
Add support for blocks to slacktest package

jamiehannaford:
This PR adds support for blocks when using the `slacktest` package. We're currently using it to run integration tests for our Slackbot, but the tests fail because the mock server does not support deserializing blocks. This PR adds support for that, in the same way it does for attachments.


ollieparsley (approved):
Thank you for this @jamiehannaford :)

PR #768, opened 2020-07-30
Add Attachments in MessageEvents

utkarshmani1997:
This commit adds the missing Attachments field in the MessageEvents.

Signed-off-by: Utkarsh <utkarshm@jfrog.com>

- [x] exposes a previously private type, const, method, etc.


utkarshmani1997:
Hi @kanata2 @ollieparsley could you please take a look at this PR?

ollieparsley (approved):
Thank you @utkarshmani1997 :)

PR #770, opened 2020-08-05
Allow EMPTY Text fields on attachments (required field)

CodeMonk:
This patch makes `text` a required field on attachments. We have code that creates a button as an attachment with empty text field. The regression caused the `text` json attribute to be removed when empty, which breaks the slack API.

This change moves it back to required, so even if empty, a field is still sent: `"text":""`

##### Pull Request Guidelines

##### PR preparation

Maybe? It was definitely a regression from 0.6.0. Without this change, you cannot post a button attachment w/o having some text included.

##### API changes

- fixed existing tests

### DIFF
--- a/attachements_test.go
+++ b/attachements_test.go
@@ -12,6 +12,7 @@ func TestAttachment_UnmarshalMarshalJSON_WithBlocks(t *testing.T) {
 
 	originalAttachmentJson := `{
     "id": 1,
+    "text":"",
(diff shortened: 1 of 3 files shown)


ollieparsley (approved):
Thank you for fixing this @CodeMonk :)

sumakumardey:
this should have omitempty text is not a required field.

PR #771, opened 2020-08-09
Update examples/eventsapi to return after error

nu50218:


ollieparsley (approved):
Thank you for fixing this @nu50218 :)

PR #772, opened 2020-08-09
Add response metadata to slack response

alyosha:
## Summary of Changes
The Web API response represented by the `SlackResponse` struct can contain an additional JSON field `"response_metadata"` in such cases as block validation failure, etc. It would be helpful to include the field in the struct so that these extra descriptors can be returned in existing models like `ViewResponse` when something goes wrong.

## Example
Below are two examples of what the `ViewResponse` looks like when dumped before/after the addition.
 
**Before**
```
(*slack.ViewResponse)(0xc000419320)({
   SlackResponse: (slack.SlackResponse) {
       Ok: (bool) false,
       Error: (string) (len=17) "invalid_arguments"
    },
    ...
}
```

**After**
```
(*slack.ViewResponse)(0xc000778500)({
    SlackResponse: (slack.SlackResponse) {
        Ok: (bool) false,                                                                                                                                            
        Error: (string) (len=17) "invalid_arguments",                                                                                                                
        ResponseMetadata: (slack.ResponseMetadata) {                                                                                                                 
           Cursor: (string) "",                                                                                                                                        
           Messages: ([]string) (len=3 cap=4) {                                                                                                                        
              (string) (len=64) "[ERROR] failed to match all allowed schemas [json-pointer:/view]",                                                                      
              (string) (len=70) "[ERROR] failed to match all allowed schemas [json-pointer:/view/title]",                                                                
              (string) (len=71) "[ERROR] must be less than 25 characters [json-pointer:/view/title/text]"                                                                
           },                                                                                                                                                          
           Warnings: ([]string) (len=1 cap=4) {                                                                                                                        
             (string) (len=15) "missing_charset"                                                                                                                        
           }                                                                                                                                                           
        } 
    },
    ...
}
```


ollieparsley (changes requested):
Hi @alyosha. Thank you for this. Is `godebug` actually used? I can't see any references to it other than in the modules and vendor list. If it isn't used (could be that I'm missing it), could you remove it please. Then I can get this merged.

misc.go
```diff
@@ -22,8 +22,9 @@ import (
 
 // SlackResponse handles parsing out errors from the web api.
 type SlackResponse struct {
-	Ok    bool   `json:"ok"`
-	Error string `json:"error"`
+	Ok               bool                   `json:"ok"`
+	Error            string                 `json:"error"`
+	ResponseMetadata map[string]interface{} `json:"response_metadata"`
```
kanata2: Maybe you can use `struct { Message []string }` as the type of `response_metadata` instead of `map[string]interface{}`.

e.g. official Java Implementation: https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/ResponseMetadata.java

kanata2 (commented):
In addition to @ollieparsley's comments, I commented a bit :)

alyosha:
hey @ollieparsley it's used in test but this is purely a personal preference (just wanted to check all fields match in the expected/received struct, where before only a few were being checked)

I will drop in lieu of reflect though, just to avoid introducing new deps :+1:

misc.go
alyosha: hey @kanata2 thanks for your feedback! I would be open to using a struct instead of a map here, will update

alyosha:
@ollieparsley @kanata2 

Thanks again for the reviews and sorry for the delay in updating :wave: 

I made the move to use a struct rather than map as @kanata2 suggested, but I noticed there are already multiple `ResponseMetadata` types floating around in the library. I decided to just add the new metadata fields (messages/warnings) to one of the existing structs, please let me know if you have any concerns here.

I confirmed the behavior using a personal project and updated the PR description with an example of what the metadata looks like following these changes

kanata2 (approved):

stevekuznetsov:
Fixes https://github.com/slack-go/slack/issues/805

Would love to see this merge, otherwise errors are not debuggable :)

Issue #785, opened 2020-08-19
When I send a message to the bot, it replies via the default slackbot


descholar-ceo:
### Here are my codes of handling MessageEvent:
```go
func handleMsgFromSlack(event *slack.MessageEvent) {
	user, err := slackClient.GetUserInfo(event.User)
	attachment := slack.Attachment{
		Pretext: "Hello @" + user.Name + "",
		Text:    "I am happy to see you here!",

		Fields: []slack.AttachmentField{
			slack.AttachmentField{
				Title: "Title of the attachment",
				Value: "This is the body",
			},
		},
	}

	channelID, timestamp, err := slackClient.PostMessage(
		event.User,
		slack.MsgOptionText("Hello there!", true),
		slack.MsgOptionAttachments(attachment),
	)

	if err != nil {
		fmt.Printf("Ooops! There is an error: %s\n", err)
		return
	}
	fmt.Printf("Message successfully sent to channel %s at %s", channelID, timestamp)
}
```

The above codes work, but in a wrong way; When a user sends a message to the bot, it replies using the default `slackbot`. The response is not coming via the bot itself. I think that's bug.

kanata2:
You can use `slack.MsgOptionAsUser(true)`.

PR #787, opened 2020-08-20
Add explanation how the message could be sent in a proper way

descholar-ceo:
### What does this PR do?
- [x] It adds a concise way of how a bot would post a message to the user as a user, not as a default slackbot `(examples/messages/messages.go)`

- [x] It fixes [#785 ](https://github.com/slack-go/slack/issues/785)


kanata2 (approved):
Thanks!

PR #788, opened 2020-08-20
Resolve: Revert changes to json tags in Attachment's Text field

prgres:
Resolve: Revert changes to json tags in Attachment's Text field. #784 

Documentation of API says that `text` field is required in `Attachment`, but when you want to send it with `blocks` field as well, it returns `invalid_attachment` error. This is because `blocks` cannot be sent along with `text` (even with empty `"text":""`)

PR #789, opened 2020-08-28
Added omitempty tag for attachment text field

mzduke:
Based on official documentation (https://api.slack.com/reference/messaging/attachments#legacy_fields):

> All of these fields are optional if you're including blocks as above. If you aren't, one of either fallback or text are required

It means the Text field of Attachment struct cannot be required. The current implementation leads to 500 responses from Slack API when you are sending Attachment without Text but with Blocks field.


kanata2:
duplicate of #788

PR #790, opened 2020-08-29
Added select block element conversations filter field

mzduke:
Based on official docs (https://api.slack.com/reference/block-kit/block-elements#conversation_select) it's possible to restrict types of conversations listed in a select block element.

In order to do this it's required to specify `filter` property with the following structure:
https://api.slack.com/reference/block-kit/composition-objects#filter_conversations

PR adds a field for SelectBlockElement struct and defines a new struct for filter options representation.


kanata2 (approved):
Thanks!

PR #791, opened 2020-09-02
Make GetConversationsParameters.ExcludeArchived optional

9bt:
We don't have to set GetConversationsParameters.ExcludeArchived  as default,
because exclude_archived in conversations.list is optional from the start.

see also https://api.slack.com/methods/conversations.list

Verification please.


kanata2 (approved):
LGTM :)

PR #793, opened 2020-09-07
Add Root field to MessageEvent to support thread_broadcast subtype

### LINKED ISSUE #792: slackevents: MessageEvent missing "Root" field for thread_broadcast SubType
Ran into an issue with the Slackbot for the community workspace, where it blew up with endless replies because it was unable to determine that a message with SubType `thread_broadcast` originated from itself. This looks to be because the `root` field isn't present on the `slackevents.MessageEvent` type, and I need to use that UserID to properly identify who sent it:

- https://api.slack.com/events/message/thread_broadcast

I believe this could be fixed by adding the following to the `MessageEvent` struct:

```Go
// Root is the message that was broadcast to the channel when the SubType is thread_broadcast. If this is not a thread_broadcast this value is likely nil.
Root *MessageEvent `json:"root"`
```

theckman:
The `thread_broadcast` subtype includes extra data in the MessageEvent that
we're currently losing, the `root` field. This is where it puts the message that
is being shared to a channel as a reply to a thread.

Fixes #792

### DIFF
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -115,6 +115,11 @@ type MessageEvent struct {
 	Files  []File `json:"files"`
 
 	Attachments []slack.Attachment `json:"attachments,omitempty"`
+
+	// Root is the message that was broadcast to the channel when the SubType is
+	// thread_broadcast. If this is not a thread_broadcast message event, this
+	// value is nil.
+	Root *MessageEvent `json:"root"`
 }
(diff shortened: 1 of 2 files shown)


kanata2 (approved):
Thanks!
(BTW, we should have some additional fields(e.g. reply_count, reply_users, etc...) in MessageEvent 😢 )

PR #800, opened 2020-09-12
Unify variables to "config"

YuzuruS:
The variable names of sendConfig existed as "c" and "config".
So Unify variables to "config".


kanata2:
Thank you :)

PR #803, opened 2020-09-22
Include GetOAuth* methods in Client's method set

### LINKED ISSUE #744: Allow mocking requests to GetOAuth* methods
In version `v0.4.0` I used to be able to mock requests to GetOAuth* methods by overriding the `SLACK_API` variable: `slack.SLACK_API = "some-test-url"`

Now this isn't possible anymore, because these methods build the request with `APIURL` which is a constant:
```
// GetOAuthResponseContext retrieves OAuth response with custom context
func GetOAuthResponseContext(...) (resp *OAuthResponse, err error) {
        ...
	if err = postForm(ctx, client, APIURL+"oauth.access", ...); err != nil {
	...
}
```

Please would you consider adding back some support for overriding the url? Alternatively, is there a better way to test this?

Thank you

vibridi:
Fixes #744 .

This PR moves the previously static `GetOAuth*` methods into `Client`'s method set. This permits to use the configurable API endpoint instead of the global constant, and the existing HTTP client that is set on the `Client`.

This PR currently doesn't include unit tests, as no such tests existed before either. Please advise if you want me to add tests.

**Note:** this PR breaks backwards compatibility.

Verification please.

### DIFF
--- a/oauth.go
+++ b/oauth.go
@@ -66,13 +66,13 @@ type OAuthV2ResponseAuthedUser struct {
 }
 
 // GetOAuthToken retrieves an AccessToken
-func GetOAuthToken(client httpClient, clientID, clientSecret, code, redirectURI string) (accessToken string, scope string, err error) {
-	return GetOAuthTokenContext(context.Background(), client, clientID, clientSecret, code, redirectURI)
+func (api *Client) GetOAuthToken(clientID, clientSecret, code, redirectURI string) (accessToken string, scope string, err error) {
+	return api.GetOAuthTokenContext(context.Background(), clientID, clientSecret, code, redirectURI)
 }
 
 // GetOAuthTokenContext retrieves an AccessToken with a custom context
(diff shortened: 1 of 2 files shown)


kanata2:
Thank you for your PR.

I agree that the testability is low due to the dependency on APIURL constant value, but this function itself does not seem to be a feature that slack.Client should have 🤔 
So, it would be nice to have some other ideas to solve this problem...

vibridi:
@kanata2 
thanks for your feedback

> does not seem to be a feature that slack.Client should have

why not? 
Even if this is the OAuth flow, it still belongs to the Slack API with the same base path, as usage of `APIURL` constant shows.

An alternative approach is to declare a different client, let's say `slack.OAuthClient`, and add these funcs to its method set, but that would require duplicating a bunch of code...

As a matter of fact, I'm not sure why you designed these OAuth funcs as static.

nlopes:
I'm going to take a different approach and use variadic arguments to allow you to set the url.

PR #811, opened 2020-10-01
Implement app-level tokens + event authorizations

cjdenio:
Hello! This PR implements Slack's brand-new flow for retrieving event authorizations, [outlined here](https://api.slack.com/changelog/2020-09-15-events-api-truncate-authed-users).

Here are two unique things about the new [`apps.event.authorizations.list`](https://api.slack.com/methods/apps.event.authorizations.list) method:
1. It requires an [app-level token](https://api.slack.com/authentication/token-types#app) instead of a standard OAuth one. I added an `OptionAppLevelToken` client option, to allow developers to optionally set one. It's not required, and won't affect any existing apps.
2. This method requires that the token be presented in the `Authorization` header, regardless of the content-type. The `postMethodWithBearerToken` function is identical to `postMethod`, except it allows the setting of a bearer token.

This change passes all linting/tests.

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/apps.go
+++ b/apps.go
@@ -19,7 +19,7 @@ type EventAuthorization struct {
 }
 
 // ListEventAuthorizations lists authed users and teams for the given event_context. You must provide an app-level token to the client using OptionAppLevelToken. More info: https://api.slack.com/methods/apps.event.authorizations.list
-func (api *Client) ListEventAuthorizations(event_context string) ([]EventAuthorization, error) {
+func (api *Client) ListEventAuthorizations(eventContext string) ([]EventAuthorization, error) {
 	resp := &listEventAuthorizationsResponse{}
 
 	err := api.postMethodWithBearerToken(context.Background(), "apps.event.authorizations.list", url.Values{
--- a/apps.go
+++ b/apps.go
@@ -1,0 +1,37 @@
+package slack
+
+import (
+	"context"
+	"net/url"
+)
+
+type listEventAuthorizationsResponse struct {
+	SlackResponse
+	Authorizations []EventAuthorization `json:"authorizations"`
+}
+
+type EventAuthorization struct {
+	EnterpriseID        string `json:"enterprise_id"`
+	TeamID              string `json:"team_id"`
+	UserID              string `json:"user_id"`
+	IsBot               bool   `json:"is_bot"`
+	IsEnterpriseInstall bool   `json:"is_enterprise_install"`
+}
+
+// ListEventAuthorizations lists authed users and teams for the given event_context. You must provide an app-level token to the client using OptionAppLevelToken. More info: https://api.slack.com/methods/apps.event.authorizations.list
+func (api *Client) ListEventAuthorizations(event_context string) ([]EventAuthorization, error) {
+	resp := &listEventAuthorizationsResponse{}
+
+	err := api.postMethodWithBearerToken(context.Background(), "apps.event.authorizations.list", url.Values{
+		"event_context": {event_context},
+	}, &resp, api.appLevelToken)
+
+	if err != nil {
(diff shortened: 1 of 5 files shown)


apps.go
```diff
@@ -0,0 +1,37 @@
...
+type EventAuthorization struct {
+	EnterpriseID        string `json:"enterprise_id"`
+	TeamID              string `json:"team_id"`
+	UserID              string `json:"user_id"`
+	IsBot               bool   `json:"is_bot"`
+	IsEnterpriseInstall bool   `json:"is_enterprise_install"`
+}
+
+// ListEventAuthorizations lists authed users and teams for the given event_context. You must provide an app-level token to the client using OptionAppLevelToken. More info: https://api.slack.com/methods/apps.event.authorizations.list
+func (api *Client) ListEventAuthorizations(event_context string) ([]EventAuthorization, error) {
```
kanata2:
```suggestion
func (api *Client) ListEventAuthorizations(eventContext string) ([]EventAuthorization, error) {
```

kanata2: We might want to add ListEventAuthorizationsContext too 😄

misc.go
```diff
@@ -246,13 +246,16 @@ func postJSON(ctx context.Context, client httpClient, endpoint, token string, js
 }
 
 // post a url encoded form.
-func postForm(ctx context.Context, client httpClient, endpoint string, values url.Values, intf interface{}, d debug) error {
+func postForm(ctx context.Context, client httpClient, endpoint string, values url.Values, intf interface{}, d debug, token ...string) error {
```
kanata2: `postForm` method is not suitable for requests that involve `Content-Type: application/json` headers.
 How about using postJSON from misc.go instead?

kanata2 (changes requested):

misc.go
cjdenio: I must have missed that function! This API method does, in fact, support `application/x-www-form-urlencoded` (as long as you pass the token in the `Authorization` header), but I'll switch it over anyway.

kanata2: Thank you!

( FYI: https://api.slack.com/web#slack-web-api__basics__post-bodies__url-encoded-bodies )

cjdenio: I understand that this is Slack's convention, though there's an issue with this particular method: it requires a bearer token in the `Authorizaion` header, regardless of whether or not you're using `application/json`.

cjdenio:
I've implemented `ListEventAuthorizationsContext`, and switched to `postJSON`! :tada:

kanata2 (approved):
LGTM! Thanks!!

slack.go
```diff
@@ -57,11 +57,12 @@ type authTestResponseFull struct {
 type ParamOption func(*url.Values)
 
 type Client struct {
-	token      string
-	endpoint   string
-	debug      bool
-	log        ilogger
-	httpclient httpClient
+	token         string
+	appLevelToken string
```
kanata2, line 61: I'm sorry. I have one more question.
Is it better to set the app level token and normal token separately?Currently there is no distinction between bot and user tokens, but I don't think this is a problem because they have a lot in common.

cjdenio: I'm assuming that there will be a fair number of cases in which someone would want to reuse a single client with both an OAuth token and an app-level token. App-level tokens are completely different from standard bot/user tokens, and I would lean towards keeping it as a separate parameter.

It also helps clear up confusion where a user might try to call the `ListEventAuthorizations` without being aware that they're using the wrong token.

If you think it should be one parameter, though, I'll definitely make that change!

kanata2: Thanks! Looks good to me!

PR #812, opened 2020-10-01
reminders.list api call implemented

KarolisKl:
##### Pull Request Guidelines

##### PR preparation
Run `make pr-prep` from the root of the repository to run formatting, linting and tests. (passed)

##### API changes

Description:

Implementation for https://api.slack.com/methods/reminders.list

Some changes in reminders.go:
- remindersResp struct for reminders.list response
- doReminders function for sending API request expecting reminders response mentioned above
- ListReminders function which calls the api request and returns reminders for the user 
- One test added (since the call takes only the token parameter not much you can test there)

### DIFF
--- a/reminders.go
+++ b/reminders.go
@@ -34,7 +34,7 @@ func (api *Client) doReminder(ctx context.Context, path string, values url.Value
 	return &response.Reminder, response.Err()
 }
 
-func (api *Client) doReminders(ctx context.Context, path string, values url.Values) (*[]Reminder, error) {
+func (api *Client) doReminders(ctx context.Context, path string, values url.Values) ([]*Reminder, error) {
 	response := &remindersResp{}
 	if err := api.postMethod(ctx, path, values, response); err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


reminders.go
```diff
@@ -29,6 +34,24 @@ func (api *Client) doReminder(ctx context.Context, path string, values url.Value
 	return &response.Reminder, response.Err()
 }
 
+func (api *Client) doReminders(ctx context.Context, path string, values url.Values) (*[]Reminder, error) {
```
kanata2:
```suggestion
func (api *Client) doReminders(ctx context.Context, path string, values url.Values) ([]*Reminder, error) {
```

kanata2 (changes requested):

reminders.go
```diff
@@ -29,6 +34,31 @@ func (api *Client) doReminder(ctx context.Context, path string, values url.Value
...
 }
 
+func (api *Client) doReminders(ctx context.Context, path string, values url.Values) ([]*Reminder, error) {
+	response := &remindersResp{}
+	if err := api.postMethod(ctx, path, values, response); err != nil {
+		return nil, err
+	}
+
+	// create an array of pointers to reminders
+	var reminders = make([]*Reminder, 0)
```
kanata2: You can get the length of slice, so it's best to allocate it in advance.

```suggestion
	var reminders = make([]*Reminder, 0, len(response.Reminders)
```

kanata2 (approved):
LGTM👍

PR #819, opened 2020-10-08
[Patch] Completed RTM Info

KarimNahas:
PR Content:
- Added missing information collected from `rtm.start` (info is generated when creating a new rtm using `RTMOptionUseStart` option)

### DIFF
--- a/info.go
+++ b/info.go
@@ -431,6 +431,15 @@ type Icons struct {
 
 // Info contains various details about the authenticated user and team.
 // It is returned by StartRTM or included in the "ConnectedEvent" RTM event.
+// It should be noted that in order to optimise the sync speed, RTMConnect
+// (which uses "rtm.connect") does not get all the account information. Namely,
+// channels, groups, mpims, ims, users and bots are left empty.
+// If you want to have access these fields populated, you should use "rtm.start"
+// instead. Sample code bellow:
+// ```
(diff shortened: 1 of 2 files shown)


info.go
```diff
@@ -432,9 +432,15 @@ type Icons struct {
...
-	Team *Team        `json:"team,omitempty"`
+	URL      string       `json:"url,omitempty"`
+	User     *UserDetails `json:"self,omitempty"`
+	Team     *Team        `json:"team,omitempty"`
+	Channels []*Channel   `json:"channels",omitempty`
+	Groups   []*Channel   `json:"groups",omitempty`
+	MPIMs    []*Channel   `json:"mpims",omitempty`
+	IMs      []*Channel   `json:"ims",omitempty`
+	Users    []*User      `json:"users,omitempty"`
+	Bots     []*User      `json:"bots,omitempty"`
```
kanata2: These fields are included only when you use `rtm.start`, so it would be nice if you could add that to comments.

KarimNahas:
@kanata2 
I've rebased the code and added a description as per your review (I tried to be as explicit as possible). 
Also made sure that all test are passing.
Tell me if any other modifications are needed.

nlopes:
The `rtm.Start` has since been deprecated and shouldn't be used so I'm not adding that. I'll add a PR for the attachments change briefly.

PR #822, opened 2020-10-13
Hotfix for block_actions unmarshal

ghost:
Despite being announced as introduced 29th of September, at least my browser app still doesn't include a "state" parameter for block_actions (mobile beta does). This leads to a `unexpected end of JSON input` error on unmarshalling.

This simple check only triggers the newly introduced special unmarshalling only when there actually is a `RawState` to handle.


jmiller101:
I'm running into the same issue and was about to open a PR before seeing yours!

kanata2 (approved):
Thank you for reporting and sending PR!

> at least my browser app still doesn't include a "state" parameter for block_actions (mobile beta does). 

Hmm, I did not know this case...!

PR #828, opened 2020-10-16
reactions.list: support for cursor-based pagination

### LINKED ISSUE #825: reactions.list not works on page-based pagination
reactions.list (https://api.slack.com/methods/reactions.list ) has parameters for pagination both of page-based and cursor-based.
But, it seems like not working on page-based because reactions.list. (it can be confirmed on official tester)

- on request, result is not affected `page` parameter.
- on response, `paging` is not returned

I fixed current code on my fork, it works well on my environment.
https://github.com/storz/slack/commit/4c187e3

- Cloud I apply this (breaking) change?
  - Or should I create another method for backward compatibility?

storz:
## WHAT

Fixes: #825 

SSIA
Note that this PR includes breaking changes about an argument of method.

ref. [Paginating through collections | Slack](https://api.slack.com/docs/pagination)

## WHY

A page-based pagination on the API is not working now.
See #825 for more details.

### DIFF
--- a/reactions.go
+++ b/reactions.go
@@ -59,28 +59,24 @@ func (res getReactionsResponseFull) extractReactions() []ItemReaction {
 }
 
 const (
-	DEFAULT_REACTIONS_USER  = ""
-	DEFAULT_REACTIONS_COUNT = 100
-	DEFAULT_REACTIONS_PAGE  = 1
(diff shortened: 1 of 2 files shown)


kanata2 (changes requested):
https://github.com/slack-go/slack/issues/825#issuecomment-722012498

PR #832, opened 2020-10-20
Change type of `ExcludeArchived` field to bool in `GetConversationsParameters`

mzduke:
I believe the type `GetConversationsParameters` by mistake has a `string` type for `ExcludeArchived` property while in the style of this library is to use `bool` values for flags and then convert it using `strconv.FormatBool()`.

The fix is quite simple, but it's a breaking change, so I'm not sure what would be the proper flow to incorporate this PR in the terms of the repository.

### DIFF
--- a/conversation.go
+++ b/conversation.go
@@ -455,7 +455,7 @@ func (api *Client) GetConversationRepliesContext(ctx context.Context, params *Ge
 
 type GetConversationsParameters struct {
 	Cursor          string
-	ExcludeArchived string
+	ExcludeArchived bool
 	Limit           int
 	Types           []string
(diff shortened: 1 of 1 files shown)


kanata2 (commented):
It seems good, but I am going to wait until next minor version release to merge due to contain breaking changes as you said.

jaysonesmith:
@kanata2 Thanks for the reply! Happy to help out with other stuff as needed. I asked in Gitter about it, too.

joshleeb:
Bumping this PR as it looks like there have been a few releases since it's creation 🙂

PR #838, opened 2020-10-30
Make possible to get authorized user's profile providing empty userID to GetUserProfileContext

2nick:
The thing is that until some time api considered empty userID value as lack of value but currently for successful execution you must to not provide query string argument at all.


kanata2 (approved):
Looks good to me :)
Thanks.

PR #839, opened 2020-11-05
Allow message unfurling with auth url

kautsig:
This PR adds a method for unfurling a message with an authentication url to initiate a custom authentication flow with the unfurling component.

See: https://api.slack.com/reference/messaging/link-unfurling#authenticated_unfurls

The slack user will be presented with the message:

```
Hi there!  Linking your slack account to <app> you to use <app> in slack (and it only takes a click or two). Would you like to set it up?

- Yes allow. - No
```


kautsig:
I closed this for now, because we already have unrelated changes in my master. I will have to clean this and potentially re-open separate PRs for both changes.

PR #843, opened 2020-11-07
reminders time field fix

KarolisKl:
##### PR preparation

Done

It's a bug fix

##### API changes

It's a bug fix for the Reminder struct.

Currently, reminders add and list functions return a parsing error.
This is because the Reminder's struct Time field is set to time.time currently.

If you take a look at reminders calls, the responses are sending us back Unix timestamp (int), not a time object.

```
Response
(Note: only non-recurring reminders will have time and complete_ts field.)

{
	"ok": true,
	"reminder": {
		"id": "Rm12345678",
		"creator": "U18888888",
		"user": "U18888888",
		"text": "eat a banana",
		"recurring": false,
		"time": 1602288000,
		"complete_ts": 0
	}

```

### DIFF
--- a/reminders.go
+++ b/reminders.go
@@ -3,17 +3,16 @@ package slack
 import (
 	"context"
 	"net/url"
-	"time"
 )
 
 type Reminder struct {
-	ID         string    `json:"id"`
-	Creator    string    `json:"creator"`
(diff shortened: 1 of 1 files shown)


kanata2 (approved):
Thanks.

PR #847, opened 2020-11-16
Added coverage for audit API endpoint

JustinJudd:
Includes structs and functions for using the [Slack Audit endpoint](https://api.slack.com/admins/audit-logs).

Test is provided based on the example output in the documentation,

### DIFF
--- a/audit.go
+++ b/audit.go
@@ -1,0 +1,103 @@
+package slack
+
+import (
+	"context"
+	"net/url"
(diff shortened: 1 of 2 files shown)


kanata2 (approved):
Looks good to me :)
(However, I don't have an owner of Slack Enterprise Grid workspace, so I couldn't even confirm if it works correctly...)

PR #850, opened 2020-11-19
Make private channel handleable as well as public channel in usergroup's default channels

atatkin:
Slack treats private channels (Groups) differently from public channels (Channels) when getting usergroup details (https://api.slack.com/types/usergroup) storing them in two separate slices (prefs.channels and prefs.groups), but then treats them as a single value ("channels") when pushing an update to the usergroup via the usergroup.update method (https://api.slack.com/methods/usergroups.update).

2 years ago when I added this, I mistakenly only accounted for the public channels which are stored in the Prefs.Channels value when combining to update the default channels in a usergroup.  Really what needs to be done is to combine the public channels and private channel ID's (Prefs.Channels and Prefs.Groups) together before pushing that back to Slack or you could inadvertently remove private channels (groups) from a usergroup.

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/usergroups.go
+++ b/usergroups.go
@@ -208,7 +208,7 @@ func (api *Client) UpdateUserGroupContext(ctx context.Context, userGroup UserGro
 	}
 
 	if len(userGroup.Prefs.Channels) > 0 {
-		values["channels"] = []string{strings.Join(userGroup.Prefs.Channels, ",")}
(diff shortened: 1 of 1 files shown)


kanata2 (commented):
Looks good! Thanks.
Could you add testing for behavior about L210-211?

PR #853, opened 2020-11-24
Add conversations.list to the slacktest server

hobbeswalsh:
This commit adds the `conversations.list` endpoint to the testing server.
This endpoint returns JSON shaped the same as `channels.list`, so I just
reused that handler.

I also found what appeared to be a bug with error handling (a misnamed error)
so I fixed that as well.


kanata2 (approved):
👍

PR #860, opened 2020-12-07
Use custom data type for unfurl

ledmonster:
13b47a4 breaks unfurl (`text` shouldn't exist in unfurl parameters), so I defined custom data type for unfurl.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -84,6 +84,10 @@ func NewPostMessageParameters() PostMessageParameters {
 	}
 }
 
+type Unfurl struct {
+	Blocks Blocks `json:"blocks,omitempty"`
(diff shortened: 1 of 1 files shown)


kanata2 (commented):
Thanks!

kanata2:
I merged https://github.com/slack-go/slack/pull/788. So this PP is no longer needed.(I'm sorry...)
If you want to use custom data type for another reasons, please feel free to tell us.

ledmonster:
Thank you. I noticed that unfurl accepts not only `blocks` but single `Attachment`. So reverting 13b47a4 is better. I'll close this PR.

PR #862, opened 2020-12-07
Avoid returning correct signature

### LINKED ISSUE #849: Error from "Ensure" should not disclose expected signature
Hello! First of all, a big thank-you to the maintainers and contributors for this library.

I'm surfacing this issue because I wonder if the error message located [here](https://github.com/slack-go/slack/blob/master/security.go#L90) is too descriptive. Specifically, it reads:

```
return fmt.Errorf("Expected signing signature: %s, but computed: %s", hex.EncodeToString(v.signature), hex.EncodeToString(computed))
```

My concern is that since Go errors tend to get blindly passed upwards, this could be eventually returned directly to a caller making a request. Because it directly discloses the expected signature, the caller could simply update to the expected signature, and re-execute the call successfully.

Is returning the expected signature actually useful for debugging? If it's not, then perhaps we could strike it completely. If it _is_, perhaps instead we could do something like:

```
log.Debugf(""Expected signing signature: %s, but computed: %s", hex.EncodeToString(v.signature), hex.EncodeToString(computed)")
return fmt.Errorf("Computed unexpected signature of: %s", hex.EncodeToString(v.signature), hex.EncodeToString(computed))
```

Am I missing something here? Let me know. If this seems reasonable, I'm happy to open a pull request, just want to check with maintainers first. Thank you!

tyrannosaurus-becks:
Closes #849 

This PR updates the code to avoid returning the correct signature. This prevents a malicious caller from viewing the error, updating their signature to the now known correct one, and re-executing the call.

A maintainer stated that if we were to remove the explicitly correct signature, it would be nice to also log the correct one at the debug level. To do that without changing the signature of the `NewSecretsVerifier` method, I added a `WithDebug` method that could be optionally called when creating the verifier, which follows [the Builder pattern](https://en.wikipedia.org/wiki/Builder_pattern). It would have been strange to take an argument that was an unexported interface, so I also exported the `Debug` interface. That makes this change is 100% backwards compatible. 

Happy to change my approach, add tests, or anything else that might help. A big thank-you to this repo's maintainers! 😄

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -18,12 +18,12 @@ type ilogger interface {
 	Println(...interface{})
 }
 
-type debug interface {
+type Debug interface {
 	Debug() bool
 
-	// Debugf print a formatted debug line.
+	// Debugf print a formatted Debug line.
 	Debugf(format string, v ...interface{})
-	// Debugln print a debug line.
+	// Debugln print a Debug line.
 	Debugln(v ...interface{})
 }
 
@@ -53,8 +53,8 @@ func (t discard) Debug() bool {
 	return false
 }
 
-// Debugf print a formatted debug line.
+// Debugf print a formatted Debug line.
 func (t discard) Debugf(format string, v ...interface{}) {}
 
-// Debugln print a debug line.
+// Debugln print a Debug line.
 func (t discard) Debugln(v ...interface{}) {}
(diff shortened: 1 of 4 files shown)


kanata2 (approved):
👍
Thanks!

tyrannosaurus-becks:
ありがとうございます

PR #865, opened 2020-12-10
Reminders list fix

KarolisKl:
##### PR preparation

Done.

Fix for a bug that appeared in https://github.com/slack-go/slack/pull/812

##### API changes

At the moment the reminders list array is being populated with the same data, this is because of Golang behaviour inside range loops - it copies the values inside its loop and gets rid of it when it's done. This PR fixes it so the data returned is correct.

Updated the test to check for this issue (that's how I confirmed the fix works)

### DIFF
--- a/reminders_test.go
+++ b/reminders_test.go
@@ -161,29 +161,30 @@ func TestSlack_DeleteReminder(t *testing.T) {
 type mockRemindersListHTTPClient struct{}
 
 func (m *mockRemindersListHTTPClient) Do(*http.Request) (*http.Response, error) {
-	responseString := "{\n" +
-		"\"ok\": true,\n" +
-		"\"reminders\": [\n" +
-		"{\n" +
-		"\"id\": \"Rm12345678\",\n" +
-		"\"creator\": \"U18888888\",\n" +
-		"\"user\": \"U18888888\",\n" +
-		"\"text\": \"eat a banana\",\n" +
-		"\"recurring\": false,\n" +
-		"\"time\": 1458678068,\n" +
-		"\"complete_ts\": 0\n" +
-		"},\n" +
-		"{\n" +
-		"\"id\": \"Gm12345678\",\n" +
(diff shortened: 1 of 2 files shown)


KarolisKl:
Hey @kanata2, the latest merge you did: https://github.com/slack-go/slack/commit/106d33d9f59459bb6111d67b0c04536e9482a796

Introduced a race error and because of that all of the following PR's will fail.
Could you fix that?

kanata2:
Fixed in #866.

kanata2:
merged

KarolisKl:
Thanks for the quick fix!

reminders_test.go
```diff
@@ -156,17 +158,53 @@ func TestSlack_DeleteReminder(t *testing.T) {
...
+		"\"id\": \"Gm12345678\",\n" +
+		"\"creator\": \"U18888888\",\n" +
+		"\"user\": \"U18888888\",\n" +
+		"\"text\": \"drink some water\",\n" +
+		"\"recurring\": false,\n" +
+		"\"time\": 1458678090,\n" +
+		"\"complete_ts\": 0\n" +
+		"}\n" +
+		"]\n" +
+		"}"
```
kanata2:
```suggestion
	responseString := `{
		"ok": true,
		"reminders": [
	        {
				"id": "Rm12345678",
				"creator": "U18888888",
				"user": "U18888888",
				"text": "eat a banana",
				"recurring": false,
				"time": 1458678068,
				"complete_ts": 0,
			},
			{
				"id": "Gm12345678",
				"creator": "U18888888",
				"user": "U18888888",
				"text": "drink some water",
				"recurring": false,
				"time": 1458678090,
				"complete_ts": 0,
			},
		],
	}`

```

kanata2 (commented):
Thank you!

kanata2:
@KarolisKl Ah, I'm sorry. Please remove trailing comma in  `responseString`.🙏 🙏

PR #868, opened 2020-12-11
Remove the unnecessary parameter from SetUserRealName and update optional param handling in profile setting

### LINKED ISSUE #809: SetUserCustomStatus results in user_not_found error
Recently `SetUserCustomStatus` started failing with a `user_not_found` error.

It looks like the `users.profile.set` no longer accepts an empty string for the `user` parameter. When I remove this parameter from the API call, it seems to work.

As a workaround, I'm now calling `SetUserCustomStatusContext` with my own User ID (which works even though the API docs say "this argument may only be specified by team admins")

### LINKED ISSUE #867: SetUserRealName doesn't work due to the unnecessary parameter included
### What happend
`SetUserRealName` returned `user_not_found`

I guess that's because `SetUserRealName` puts `realName` to the second parameter of `SetUserRealNameContextWithUser`.
`SetUserRealNameContextWithUser` uses this as `user` field in the request to `users.profile.set`.
But, according to [Slack API doc](https://api.slack.com/methods/users.profile.set), this field should have `ID of user to change`, not the real name to set.

```.go
func (api *Client) SetUserRealName(realName string) error {
	return api.SetUserRealNameContextWithUser(context.Background(), realName, realName)
}

// SetUserRealNameContextWithUser will set a real name for the provided user with a custom context
func (api *Client) SetUserRealNameContextWithUser(ctx context.Context, user, realName string) error {
	profile, err := json.Marshal(
		&struct {
			RealName string `json:"real_name"`
		}{
			RealName: realName,
		},
	)

	if err != nil {
		return err
	}

	values := url.Values{
		"user":    {user},
		"token":   {api.token},
		"profile": {string(profile)},
	}

	response := &userResponseFull{}
	if err = api.postMethod(ctx, "users.profile.set", values, response); err != nil {
		return err
	}

	return response.Err()
}
```

### Expected behavior
1. Ensure an empty string is set to the second parameter of `SetUserRealNameContextWithUser` in `SetUserRealName`

As `user` field in  `users.profile.set` is optional, `SetUserRealName` can set an empty string to the second parameter of `SetUserRealNameContextWithUser`.

Actually, I could confirm it works even with an empty value [here](https://api.slack.com/methods/users.profile.set/test).

2. Ensure `user` field should be added only when a non-empty value is passed to the parameter in `SetUserRealNameContextWithUser`

Also, I found `https://slack.com/api/users.profile.set` doesn't receive an empty string in `user` field. 
In association with the above change, we need to change the handling of this field in `SetUserRealNameContextWithUser`

Without `user` param
```
$ curl -X POST "https://slack.com/api/users.profile.set?token=xxx&profile=%7B%22real_name%22%3A%20%22User%20Name%22%7D&pretty=1"
-> success
```

With an empty `user` param
```
curl -X POST "https://slack.com/api/users.profile.set?token=xxx&profile=%7B%22real_name%22%3A%20%22User%20Name%22%7D&user=%22%22&pretty=1"
-> receive "user_not_found"
```

### Steps to reproduce
You should probably get the error by running the following.

```.go
package main
import (
	"fmt"
	"github.com/slack-go/slack"
)
func main() {
	api := slack.New(
		"xxx",
	)
	err := api.SetUserRealName("User Name")
	fmt.Println(err)
}
```

### Versions
- Go: 1.15.5
- slack-go/slack: 0.7.0

sryoya:
##### PR preparation
Fixes #867 
Fixes #809

### DIFF
--- a/users.go
+++ b/users.go
@@ -513,7 +513,7 @@ func (api *Client) DeleteUserPhotoContext(ctx context.Context) (err error) {
 //
 // For more information see SetUserRealNameContextWithUser
 func (api *Client) SetUserRealName(realName string) error {
-	return api.SetUserRealNameContextWithUser(context.Background(), realName, realName)
+	return api.SetUserRealNameContextWithUser(context.Background(), "", realName)
 }
 
 // SetUserRealNameContextWithUser will set a real name for the provided user with a custom context
@@ -531,9 +531,13 @@ func (api *Client) SetUserRealNameContextWithUser(ctx context.Context, user, rea
 	}
 
 	values := url.Values{
-		"user":    {user},
 		"token":   {api.token},
 		"profile": {string(profile)},
+	}
+
+	// optional field. It should not be set if empty
+	if user != "" {
+		values["user"] = []string{user}
 	}
 
 	response := &userResponseFull{}
@@ -598,9 +602,13 @@ func (api *Client) SetUserCustomStatusContextWithUser(ctx context.Context, user,
 	}
 
 	values := url.Values{
-		"user":    {user},
 		"token":   {api.token},
 		"profile": {string(profile)},
+	}
+
+	// optional field. It should not be set if empty
+	if user != "" {
+		values["user"] = []string{user}
 	}
 
 	response := &userResponseFull{}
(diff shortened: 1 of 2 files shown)


kanata2 (approved):
Looks good. Thanks!

PR #869, opened 2020-12-12
remove .gometalinter.json

suzuki-shunsuke:
gometalinter isn't used now.

https://github.com/slack-go/slack/pull/840

##### PR preparation

> Run `make pr-prep` from the root of the repository to run formatting, linting and tests.

I did.

##### API changes

No change.


kanata2 (approved):
Thanks!

PR #873, opened 2020-12-18
Fix button style defenition and add test for that

### LINKED ISSUE #871: [BUG]Button default style is wrong
Due to the introduction for slack button, default style should be omitted.
And there will be an invalid block error when set to default.

```
{
	"blocks": [
		{
			"type": "actions",
			"block_id": "actionblock789",
			"elements": [
				{
					"type": "button",
					"text": {
						"type": "plain_text",
						"text": "Primary Button"
					},
should be deleted --->			"style": "default",
					"value": "click_me_456"
				}
			]
		}
	]
}
```

code is here, I think default style="" should fix this issue.
https://github.com/slack-go/slack/blob/e96a24877c12e04ce3692a3096559b2ccfbeb26d/block_element.go#L133

sryoya:
# Description
- Fixes #871 
- Creates new tests to confirm `Style` values
- adds and updates related comments to be user-friendly and compliant with Go standard conventions

# Note
I found a minor difference of the default value handling between Block Element and Confirmation Dialog and had some concern about this `Style` type.
But, if I had done the related update, I would have updated an exported API and constant and made that backward incompatible.
So, raised the suggestion issue #874 instead.

### DIFF
--- a/block_element.go
+++ b/block_element.go
@@ -127,10 +127,12 @@ func NewImageBlockElement(imageURL, altText string) *ImageBlockElement {
 	}
 }
 
+// Style is a style of Button element
+// https://api.slack.com/reference/block-kit/block-elements#button__fields
 type Style string
 
 const (
-	StyleDefault Style = "default"
+	StyleDefault Style = ""
 	StylePrimary Style = "primary"
 	StyleDanger  Style = "danger"
 )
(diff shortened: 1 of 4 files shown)


kanata2 (commented):
Sorry for slow response. LGTM!

kanata2 (approved):

PR #875, opened 2020-12-21
Add missing `Image512` field in `UserProfile`

p1ass:
🙆 

##### API changes

> Since API changes have to be maintained they undergo a more detailed review and are more likely to require changes.

Added `Image512` field in `UserProfile`.

ref: https://api.slack.com/types/user


kanata2 (approved):
Looks good! Thanks.

PR #879, opened 2021-01-05
Adding apps.uninstall method

gsavit:
##### Changes
Adding the `apps.uninstall` method, which completely removes a Slack app installation from its workspace (revoking any and all tokens in the process).

##### Docs
https://api.slack.com/methods/apps.uninstall

### DIFF
--- a/apps.go
+++ b/apps.go
@@ -3,6 +3,7 @@ package slack
 import (
 	"context"
 	"encoding/json"
+	"net/url"
 )
 
 type listEventAuthorizationsResponse struct {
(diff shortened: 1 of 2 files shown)


apps.go
```diff
@@ -41,3 +42,20 @@ func (api *Client) ListEventAuthorizationsContext(ctx context.Context, eventCont
 
 	return resp.Authorizations, nil
 }
+
+func (api *Client) UninstallApp(clientID string, clientSecret string) error {
```
kanata2:
```suggestion
func (api *Client) UninstallApp(clientID, clientSecret string) error {
```

kanata2 (commented):
Looks good! Thank you.

PR #880, opened 2021-01-10
Add support for "channel_created" in the Events API.

cure:
The events API support is missing a number of callback events. This PR adds support for the "channel_created" event.


slackevents/inner_events.go
```diff
@@ -290,6 +306,8 @@ const (
 	AppHomeOpened = "app_home_opened"
 	// AppUninstalled Your Slack app was uninstalled.
 	AppUninstalled = "app_uninstalled"
+	// ChannelCreated
```
kanata2:
```suggestion
	// ChannelCreated is sent when a new channel is created.
```

Issue #878, opened 2021-01-05
statusCodeError struct should be accessible outside package


gsavit:
### Description
Currently the `statusCodeError` struct is private. It should be made public/accessible outside the package so that callers are able to inspect status codes on error responses. 

By comparison, I see the similar `RateLimitedError` struct is indeed public.

gsagula:
I can take this one.

PR #885, opened 2021-01-14
Socket Mode support

### LINKED ISSUE #883: Feature request: support new Socket Mode feature
https://api.slack.com/apis/connections/socket

mumoshu:
This is a ~~WIP~~ implementation of Socket Mode support for this library based on the official documentation at https://api.slack.com/apis/connections/socket and some reverse engineering on https://github.com/slackapi/node-slack-sdk/blob/main/packages/socket-mode.

Please read [the official annoucement](https://medium.com/slack-developer-blog/socket-to-me-3d122f96d955) for more information about Socket Mode.

Resolves #883

Notes for reviewers:

- [x] This PR introduces NO API changes to existing types
- [x] `socketmode.Client` is added, which is a bit similar to the existing `slack.RTM`
- [ ] Unit tests for the new feature
- [ ] Documentation update (where should I add this to?)

##### Pull Request Guidelines

##### PR preparation

##### API changes


socket_mode_client.go
```diff
@@ -0,0 +1,115 @@
...
+	URL  string                 `json:"url,omitempty"`
+	Data map[string]interface{} `json:"-"`
+}
+
+// SocketModeMessage maps to each message received via the WebSocket connection of SocketMode
+type SocketModeMessage struct {
+	Type       string                   `json:"type"`
+	Reason     string                   `json:"reason"`
+	Payload    SocketModeMessagePayload `json:"payload"`
+	EnvelopeID string                   `json:"envelope_id"`
```
seratch: FYI: https://github.com/slackapi/python-slack-sdk/blob/v3.2.0/slack_sdk/socket_mode/request.py#L7-L12

A Socket Mode envelope can have the following attributes.

* AcceptsResponsePayload bool
* RetryAttempt *int // only for Events API
* RetryReason string // only for Events API

examples/socketmode/events.go
```diff
@@ -0,0 +1,48 @@
...
+		"YOUR TOKEN HERE",
+		slack.OptionDebug(true),
+		slack.OptionLog(log.New(os.Stdout, "slack-bot: ", log.Lshortfile|log.LstdFlags)),
+	)
+
+	client := api.NewSocketModeClient()
+	go client.ManageConnection()
+
+	for evt := range client.IncomingEvents {
+		eventsAPIEvent, ok := evt.Data.(slackevents.EventsAPIEvent)
```
seratch: How does handing of "interactive"/"slash_commands" type envelopes look like?

socket_mode_managed_conn.go
```diff
@@ -0,0 +1,476 @@
...
+
+	if err := rtm.sendWithDeadline(msg); err != nil {
+		rtm.IncomingEvents <- rtm.internalEvent("outgoing_error", &OutgoingErrorEvent{
+			Message:  msg,
+			ErrorObj: err,
+		})
+	}
+}
+
+// ping sends a 'PING' message to the RTM's websocket. If the 'PING' message
```
seratch: I don't think this method will be used for ping-pong communication.

examples/socketmode/events.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/slack-go/slack"
+	"github.com/slack-go/slack/slackevents"
+)
+
+// You more than likely want your "Bot User OAuth Access Token" which starts with "xoxb-"
+var api = slack.New("TOKEN")
```
seratch: I know it's a bit confusing but the token for `apps.connections.open` is not a bot/user token (xoxb-/xoxp-). The token type for it is "App-Level Token" (xapp-). Thus, Socket Mode apps usually use two tokens, app token (xapp-) for `apps.connections.open` plus a bot/user token.

socket_mode.go
```diff
@@ -0,0 +1,92 @@
...
+	SlackResponse
+	SocketModeConnection
+}
+
+// StartSocketModeContext calls the "apps.connections.open" endpoint and returns the provided URL and the full Info block with a custom context.
+//
+// To have a fully managed Websocket connection, use `NewRTM`, and call `ManageConnection()` on it.
+func (api *Client) StartSocketModeContext(ctx context.Context) (info *SocketModeConnection, websocketURL string, err error) {
+	response := &openResponseFull{}
+	err = api.postMethod(ctx, "apps.connections.open", url.Values{"token": {api.token}}, response)
```
seratch: The token here needs to be an app-level token.

seratch (commented):
@mumoshu You are so quick! I am happy to be of help. Please feel free to ask me or other folks in the github.com/SlackAPI team any questions 👋

socket_mode.go
cjdenio: @mumoshu app-level tokens were PRed in a while back, so `api.appLevelToken` should work. The user sets it by passing `OptionAppLevelToken` to the client definition.

examples/socketmode/events.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	botToken := os.Getenv("SLACK_BOT_TOKEN")
+	if botToken == "" {
+		fmt.Fprintf(os.Stderr, "SLACK_BOT_TOKEN must be set.\n")
+		os.Exit(1)
+	}
+
+	if !strings.HasPrefix(botToken, "xoxb-") {
+		fmt.Fprintf(os.Stderr, "SLACK_APP_TOKEN must have the prefix \"xoxb-\".")
```
seratch:
```suggestion
		fmt.Fprintf(os.Stderr, "SLACK_BOT_TOKEN must have the prefix \"xoxb-\".")
```

slacksocketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,469 @@
...
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/gorilla/websocket"
+	"github.com/slack-go/slack/internal/errorsx"
+	"github.com/slack-go/slack/internal/timex"
+)
+
+// Run can be called on a Slack RTM instance returned by the
```
seratch: We may want to replace "RTM" with "Socket Mode" in comments.

slacksocketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+}
+
+// ack tells Slack that the we have received the SocketModeRequest denoted by the envelope ID,
+// by sending back the envelope ID over the WebSocket connection.
+func (smc *Client) ack(envelopeID string) error {
+	smc.Debugln("Sending ACK ", envelopeID)
+
+	// See https://github.com/slackapi/node-slack-sdk/blob/c3f4d7109062a0356fb765d53794b7b5f6b3b5ae/packages/socket-mode/src/SocketModeClient.ts#L417
+	msg := map[string]interface{}{"envelope_id": envelopeID}
```
seratch: The response message can have "payload" in addition to "envelop_id" for interactive / slash_commands types, while events_api requires only "envelope_id".

* https://api.slack.com/apis/connections/socket-implement#acknowledge
* https://github.com/slackapi/java-slack-sdk/tree/v1.5.1/slack-api-client/src/main/java/com/slack/api/socket_mode/response

`slack.OutgoingMessage` is not compatible with Socket Mode.

slacksocketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,469 @@
...
+		if err != nil {
+			return ""
+		}
+
+		smc.IncomingEvents <- smc.externalEvent(eventsAPIEvent.Type, eventsAPIEvent)
+
+		// We automatically ack the message.
+		// TODO Should there be any way to manually ack the msg, like the official nodejs client?
+		smc.ack(event.EnvelopeID)
+	case "disconnect":
```
seratch: FYI https://api.slack.com/apis/connections/socket-implement#disconnect

slacksocketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,469 @@
...
+
+	default:
```
seratch: others are interactive and slash_commands

internal/misc/misc.go
```diff
@@ -0,0 +1,29 @@
+package misc
+
+import (
+	"fmt"
+	"net/http"
+)
+
+// StatusCodeError represents an http response error.
+// type httpStatusCode interface { HTTPStatusCode() int } to handle it.
+type StatusCodeError struct {
```
kanata2, line 10: Some people want to use `StatusCodeError` in their code, so it might be better not to move it to internal.

See: https://github.com/slack-go/slack/issues/878

kanata2: Also, I would like to make the package name more specific instead of `misc`.

examples/socketmode/socketmode.go
```diff
@@ -0,0 +1,80 @@
+package main
+
+import (
+	"fmt"
+	"github.com/slack-go/slack/slacksocketmode"
```
kanata2: How about removing `slack` prefix from package name?

slacksocketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,441 @@
...
+	req := &Request{}
+	err := json.Unmarshal(wsMsg, req)
+	if err != nil {
+		smc.IncomingEvents <- smc.internalEvent("unmarshalling_error", &slack.UnmarshallingErrorEvent{err})
+		return ""
+	}
+
+	smc.Debugf("Handling WebSocket message: %s", wsMsg)
+
+	// See https://github.com/slackapi/node-slack-sdk/blob/main/packages/socket-mode/src/SocketModeClient.ts#L533
```
kanata2: You might want to specify the revision or version since the reference line may shift depending on the update of the source code.

slacksocketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,441 @@
...
+	defer smc.conn.SetWriteDeadline(time.Time{})
+
+	if err := smc.conn.WriteJSON(msg); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (smc *Client) internalEvent(tpe string, data interface{}) ClientEvent {
```
kanata2: What does `tpe` mean ? 🤔

kanata2: What is the difference between `internalEvent` and `externalEvent`?

kanata2 (commented):
Great work! Thank you.
I've add a few comments, but I will do another detailed review when you finish implementing.

examples/socketmode/socketmode.go
mumoshu: I definitely prefer removing the prefix. I just thought you'd prefer it with the prefix as you already do so in e.g. [`slackevents`](https://github.com/slack-go/slack/tree/master/slackevents) 😄

internal/misc/misc.go
mumoshu: @kanata2 I agree. But can we do it in another PR? It was [already private before this change](https://github.com/slack-go/slack/blob/2133f59a0871dcda0ab3d627b96d28feb5ceff29/misc.go#L45-L50), and I don't want to introduce another public API that is irrelevant to this PR here.

examples/socketmode/socketmode.go
mumoshu: Addressed in 852061eaff58f9acb2a18f70ac871341d0f57982

slacksocketmode/socket_mode_managed_conn.go
mumoshu: `tpe` is another form of `type` you occasionally use when you want a variable for some kind of type but you need to avoid conflict with the reserved keyword `type`

mumoshu: Re `tpe`, I'm fine to hear your opinion for alternative name but anyway, you can see it used widely https://github.com/search?l=Go&q=%22tpe%22&type=Code

mumoshu: > What is the difference between internalEvent and externalEvent?

Thanks for pointing it out.

`internalEvent` vs `externalEvent` was some artifact left while I was scratching my head structuring code. It's not necessary anymore and removed in 11f20f31b34a0c2b7a80a21c794f4ea96865bc4a!

slacksocketmode/socket_mode_managed_conn.go
mumoshu: Ah thanks! First of all, I didn't realize till now that `slash_commands` and `interactive` are top-level types that are comparable to `events_api` and `hello` and so on.

And it's really good to know that the way we ack the socket mode "request" (or "event", which do you prefer btw? I thought you call it `SocketModeRequest` in python-sdk and `event` in node-sdk)

I will soon try modifying `examples/socketmode` and the library and test it accordingly and circle back to you.

socket_mode.go
mumoshu: @seratch @cjdenio Thanks to your comments I updated it to use `appLevelToken` for `apps.connection.open` and it's now working 👍

socket_mode_client.go
mumoshu: Thanks! I've added a bunch of fields according to your info and my observation

https://github.com/mumoshu/slack/blob/852061eaff58f9acb2a18f70ac871341d0f57982/socketmode/request.go#L15-L38

slacksocketmode/socket_mode_managed_conn.go
mumoshu: Indeed! Fixed as part of 88722e9

internal/misc/misc.go
mumoshu: Note that I made it under `internal`. Even though its name is capitalized our users are still unable to use it yet.

slacksocketmode/socket_mode_managed_conn.go
mumoshu: The first changeset for this: https://github.com/slack-go/slack/pull/885/commits/88722e9b1ed603fbea5d962aa0dd9a0f0b7cc639#diff-f5d9e76c79011a908909fa0f14332b25faba7ef70c64f12682937d35382f3f47R395-R425

examples/socketmode/events.go
mumoshu: They look like https://github.com/mumoshu/slack/blob/0be002ac0a943e28b8651abe5c65a0f00097c465/examples/socketmode/socketmode.go#L79-L96 now

mumoshu:
This should be good enough to be reviewed now, as I have manually verified this to work with `events_api`, `interactive`, and `slash_commands`.

That is, I was able to respond to a mention to the bot by handling `events_api` events, then respond with a blockkit button against a slash command executed by me handling the `slash_commands` event, and then  print some log message on button click by handling the `interactive` event.

examples/socketmode/socketmode.go
```diff
@@ -0,0 +1,141 @@
...
+				case slackevents.CallbackEvent:
+					innerEvent := eventsAPIEvent.InnerEvent
+					switch ev := innerEvent.Data.(type) {
+					case *slackevents.AppMentionEvent:
+						_, _, err := api.PostMessage(ev.Channel, slack.MsgOptionText("Yes, hello.", false))
+						if err != nil {
+							fmt.Printf("failed posting message: %v", err)
+						}
+					}
+				case slackevents.MemberJoinedChannel:
```
seratch: It seems this pattern matching does not work. Probably, matching with `innerEvent.Data.(type)` instead of  `eventsAPIEvent.Type` is correct.

socketmode/socketmode.go
```diff
@@ -0,0 +1,107 @@
...
+	EventTypeConnected        = EventType("connected")
+	EventTypeErrorWriteFailed = EventType("write_error")
+
+	//
+	// The following event types are guaranteed to not change unless Slack changes
+	//
+
+	EventTypeEventsAPI    = EventType("events_api")
+	EventTypeInteractive  = EventType("interactive")
+	EventTypeSlashCommand = EventType("slash_command")
```
seratch: Although this string value held in `EventType` is not used by any code at this point , I suggest using consistent values here.
```suggestion
	EventTypeSlashCommand = EventType("slash_commands")
```

mumoshu: Good point! Fixed in 5a2d8db

examples/socketmode/socketmode.go
mumoshu: You have an eagle eye! Fixed in 18c4fd2

mumoshu:
Added proper support (as I think) for `disconnect` handling - `Client.Run()` has been redesigned to handle reconnection when Slack sends `disconnect`-type request over the Socket Mode connection.

Also note that the `Client` struct has much less fields = states now, which I believe to help maintainability.

internal/misc/misc.go
kanata2: > But can we do it in another PR?

👍

slacksocketmode/socket_mode_managed_conn.go
kanata2: > tpe is another form of type you occasionally use when you want a variable for some kind of type but you need to avoid conflict with the reserved keyword type

That was very educational. Thank you :)

socketmode/client.go
```diff
@@ -0,0 +1,63 @@
...
+// but the structure and the design has been adapted as much as possible to that of our RTM client for consistency
+// within the library.
+//
+// You can instantiate the socket mode client with
+// Client's New() or NewSocketModeClientWithOptions(*SocketModeClientOptions)
+type Client struct {
+	// Client is the main API, embedded
+	apiClient slack.Client
+
+	idGen slack.IDGenerator
```
seratch: This ID generator is not used anywhere. We may want to remove this at this time.

socketmode/request.go
```diff
@@ -0,0 +1,38 @@
...
+	// Reason can be "warning" or else
+	Reason string `json:"reason"`
+
+	// `hello` and `disconnect` types only
+	DebugInfo DebugInfo `json:"debug_info"`
+
+	// `events_api` type only
+	EnvelopeID string `json:"envelope_id"`
+	// TODO Can it really be a non-object type?
+	// See https://github.com/slackapi/python-slack-sdk/blob/3f1c4c6e27bf7ee8af57699b2543e6eb7848bcf9/slack_sdk/socket_mode/request.py#L26-L31
```
seratch, line 33: Ah, never mind about L30-31 in the Python SDK (the code should be improved). You can safely assume the "payload" is an object.

socketmode/socketmode.go
```diff
@@ -0,0 +1,111 @@
...
+	EventTypeDisconnect   = EventType("disconnect")
+	EventTypeEventsAPI    = EventType("events_api")
+	EventTypeInteractive  = EventType("interactive")
+	EventTypeSlashCommand = EventType("slash_commands")
+
+	websocketDefaultTimeout = 10 * time.Second
+	defaultMaxPingInterval  = 30 * time.Second
+)
+
+// Open calls the "apps.connection.open" endpoint and returns the provided URL and the full Info block.
```
seratch:
```suggestion
// Open calls the "apps.connections.open" endpoint and returns the provided URL and the full Info block.
```

socketmode/socketmode.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+}
+
+// OptionPingInterval determines how often we expect Slack to deliver WebSocket ping to us.
+// If no ping is delivered to us within this interval after the last ping, we assumes the WebSocket connection
+// is dead and needs to be reconnected.
+func OptionPingInterval(d time.Duration) Option {
+	return func(rtm *Client) {
+		rtm.maxPingInterval = d
+	}
```
seratch: There are still many "RTM" names and comments like this:
```suggestion
	return func(client *Client) {
		client.maxPingInterval = d
	}
```

socketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,549 @@
...
+			if err != nil {
+				smc.Events <- newEvent(EventTypeErrorBadMessage, &ErrorBadMessage{
+					Cause:   err,
+					Message: message,
+				})
+			} else if evt != nil {
+				if evt.Type == EventTypeDisconnect {
+					// We treat the `disconnect` request from Slack as an error internally,
+					// so that we can tell the consumer of this function to reopen the connection on it.
+					return ErrorRequestedDisconnect{}
```
seratch: How does the handling code look like? Do developers need to write a handler on their own? If so, having the code in the example should be helpful. The disconnect requests can be regularly sent to all Socket Mode apps. This is not a rare case.

If this library can have the reconnection logic (= acquiring a new WSS URL and establishing a new connection to the endpoint under the hood), that's amazing!

seratch (commented):
I just left a few more comments. This pull request looks almost done 👍  

(Please feel free to mention me whenever you and the maintainers of this project need my help.)

socketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,549 @@
...
+	}
+
+	var evt Event
+
+	// See below two links for all the available message types.
+	// - https://github.com/slackapi/node-slack-sdk/blob/c3f4d7109062a0356fb765d53794b7b5f6b3b5ae/packages/socket-mode/src/SocketModeClient.ts#L533
+	// - https://api.slack.com/apis/connections/socket-implement
+	switch req.Type {
+	case RequestTypeHello:
+		evt = newEvent("hello", nil, req)
```
kanata2:
```suggestion
		evt = newEvent(EventTypeHello, nil, req)
```

socketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,549 @@
...
+		Max: 5 * time.Minute,
+	}
+
+	for {
+		var (
+			backoff time.Duration
+		)
+
+		// send connecting event
+		smc.Events <- newEvent("connecting", &slack.ConnectingEvent{
```
kanata2: It might be good to define connecting event type too.
(I thought it would be nice to be able to handle this event properly when running examples/socketmode/socketmode.go.)

socketmode/socket_mode_managed_conn.go
```diff
@@ -0,0 +1,549 @@
...
+	case len(event) == 0:
+		smc.Debugln("Received empty event")
+	default:
+		buf := &bytes.Buffer{}
+		d := json.NewEncoder(buf)
+		d.SetIndent("", "  ")
+		if err := d.Encode(event); err != nil {
+			smc.Debugln("Failed encoding decoded json:", err)
+		}
+		reencoded := buf.String()
```
kanata2: Prefer to apply only when debug mode.

socket_mode.go
```diff
@@ -0,0 +1,35 @@
...
+	SocketModeConnection
+}
+
+// StartSocketModeContext calls the "apps.connections.open" endpoint and returns the provided URL and the full Info block with a custom context.
+//
+// To have a fully managed Websocket connection, use `NewRTM`, and call `Run()` on it.
+func (api *Client) StartSocketModeContext(ctx context.Context) (info *SocketModeConnection, websocketURL string, err error) {
+	response := &openResponseFull{}
+	err = postJSON(ctx, api.httpclient, api.endpoint+"apps.connections.open", api.appLevelToken, nil, response, api)
+	//err = api.postMethod(ctx, "apps.connections.open", url.Values{"token": {api.appLevelToken}}, response)
```
kanata2: ✂️

socket_mode.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+type openResponseFull struct {
+	SlackResponse
+	SocketModeConnection
+}
+
+// StartSocketModeContext calls the "apps.connections.open" endpoint and returns the provided URL and the full Info block with a custom context.
+//
+// To have a fully managed Websocket connection, use `NewRTM`, and call `Run()` on it.
```
kanata2:
```suggestion
// To have a fully managed Websocket connection, use `New()`, and call `Run()` on it.
```

socketmode/client.go
```diff
@@ -0,0 +1,63 @@
...
+
+type ConnectionInfo struct {
+	AppID string `json:"app_id"`
+}
+
+type SocketModeMessagePayload struct {
+	Event json.RawMessage `json:"´event"`
+}
+
+// Client allows allows programs to communicate with the
```
kanata2:
```suggestion
// Client allows programs to communicate with the
```

kanata2 (commented):
Added few comments, but looks good to me!

socketmode/client.go
mumoshu: Thanks 👍 Removed in a606305

socketmode/request.go
mumoshu: Got it. Thanks for clarification 😃

socketmode/socketmode.go
mumoshu: Thanks! I believe we're all clear now.

socketmode/socket_mode_managed_conn.go
mumoshu: Thanks! Addressed in 2cbdf3b

socketmode/socket_mode_managed_conn.go
mumoshu: Added EventTypeConnection in 2cbdf3b and some code to the example in 04ead06 👍

socketmode/socket_mode_managed_conn.go
mumoshu: Good point! Done in e9e7822

socket_mode.go
mumoshu: 😊 

Addressed in c0cd3da

socket_mode.go
mumoshu: Fixed in fcf2b1e. I've also added mention to `slackmode.Client` and edited the message to refer to Socket Mode

socketmode/client.go
mumoshu: Thanks!
I've also brushed up other parts of this comment in 0f4bd7d

socketmode/socket_mode_managed_conn.go
mumoshu: > If this library can have the reconnection logic (= acquiring a new WSS URL and establishing a new connection to the endpoint under the hood), that's amazing!

Good question - Yes I've made the client to handle the reconnection on its own, following all the awesome Socket Mode clients you've written for other languages 😄 

For your reference, the below is where we implement the infinite reconnection loop:

https://github.com/mumoshu/slack/blob/e9e7822646b5341295e0854167d774e4fb5cd33b/socketmode/socket_mode_managed_conn.go#L36-L42

mumoshu: BTW In Go anything that is NOT under `internal` package whose name starts with an uppercase letter is considered public - so `ErrorRequestedDisconnect` is fairly confusing as it can give users misconception that one might want to use this outside of this package.

I'm going to make it private so that it is clear that the consumer of the client does not need to know about this internal error happens on disconnect request.

mumoshu:
Thanks to everyone for reviewing! I believe I've addressed all the feedback we had so far.

kanata2 (approved):
LGTM!!!

@seratch How do you think?

socketmode/socket_mode_managed_conn.go
seratch: >so ErrorRequestedDisconnect is fairly confusing as it can give users misconception that one might want to use this outside of this package.

Yes, actually I was thinking so too. Making it private should be the right direction 👍 

Also, I've verified the reconnecting logic works properly. Great work!

seratch (approved):
Although I am not a maintainer of this project, please allow me to give an approval. I am pretty sure that this Socket Mode client implementation is production-ready 👍

kanata2:
released! https://github.com/slack-go/slack/releases/tag/v0.8.0

prologic:
Hey 👋 Did the [slacktest](https://pkg.go.dev/github.com/slack-go/slack/slacktest) package ever get updated at all? Doesn't look like it did by this PP, and that means using it to unit test your bot's business logic doesn't currently work :/ e.g:

```
time="2023-12-11T14:14:58+10:00" level=debug msg="Starting SocketMode" file=socketmode/socket_mode_managed_conn.go line=293 logger=log2Logrus
2023/12/11 14:14:58 (127.0.0.1:57589 -) "POST /apps.connections.open HTTP/1.1" 404 19 41µs
time="2023-12-11T14:14:58+10:00" level=debug msg="HTTP/1.1 404 Not Found\r\nContent-Length: 19\r\nContent-Type: text/plain; charset=utf-8\r\nDate: Mon, 11 Dec 2023 04:14:58 GMT\r\nX-Content-Type-Options: nosniff\r\n\r\n404 page not found\n\n" file=misc.go line=266 logger=log2Logrus
```

PR #886, opened 2021-01-17
[Update] Update Interactions Message Method

sivchari:
This is my first contribution, so I apologize for any slip-ups.

##### PR preparation

- [x] Done.

This PR is updates interactions.go
The changes is adding method about validating verification tokens.
It can be easy to use for developers to compare verification tokens to sent token by slack interactive message.
I referred to ValidateToken method of slash.go

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -117,6 +117,16 @@ func (ic *InteractionCallback) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// ValidateToken validates verificationTokens
+func (ic InteractionCallback) ValidateToken(verificationTokens ...string) bool {
+	for _, token := range verificationTokens {
+		if ic.Token == token {
+			return true
+		}
+	}
+	return false
+}
+
 type Container struct {
 	Type         string      `json:"type"`
 	ViewID       string      `json:"view_id"`
--- a/interactions_test.go
+++ b/interactions_test.go
@@ -391,3 +391,76 @@ func TestInteractionCallback_InteractionTypeBlockActions_Unmarshal(t *testing.T)
 		cb.BlockActionState.Values["section_block_id"]["multi_convos"].SelectedConversations,
 		[]string{"G12345"})
 }
+
+func TestInteractionCallback_ValidateToken(t *testing.T) {
+	verificationTokens := []string{"valid", "valid2"}
+	tests := []struct {
+		name    string
+		args    InteractionCallback
+		wantErr bool
+	}{
+		{
+			name: "InteractionTypeDialogSubmission",
+			args: InteractionCallback{
+				Type:  InteractionTypeDialogSubmission,
+				Token: "valid",
+				User: User{
+					ID:   "testid",
+					Name: "testuser",
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "InteractionTypeViewClosed",
+			args: InteractionCallback{
+				Type:  InteractionTypeViewClosed,
+				Token: "valid",
+				User: User{
+					ID:   "testid",
+					Name: "testuser",
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "InteractionTypeViewSubmission",
+			args: InteractionCallback{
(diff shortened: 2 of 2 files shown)


interactions.go
```diff
@@ -117,6 +117,16 @@ func (ic *InteractionCallback) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// ValidateToken validates verificationTokens
+func (ic InteractionCallback) ValidateToken(verificationTokens ...string) bool {
```
kanata2, line 121: I can't decide if this interface is appropriate or not, so could you share actual examples?

interactions_test.go
```diff
@@ -391,3 +391,76 @@ func TestInteractionCallback_InteractionTypeBlockActions_Unmarshal(t *testing.T)
...
+			args: InteractionCallback{
+				Token: "invalid",
+			},
+			wantErr: false,
+		},
+	}
+	for i, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := tt.args.ValidateToken(verificationTokens...)
+			if got != tt.wantErr {
```
kanata2, line 461: I think `wantErr` is not appropriate naming because `(InteactionCallback).ValidateToken` just returns boolean 😄

interactions.go
kanata2, line 121: In this case, token means verification token, but Slack has two tokens.(the other is signing token.)
So we want to use more specific name in order to prevent users from making a mistake.

kanata2: FYI:
Verification Token is deprecated.
https://api.slack.com/authentication/verifying-requests-from-slack#verifying-requests-from-slack-using-signing-secrets__verification-token-deprecation

interactions_test.go
```diff
@@ -391,3 +391,76 @@ func TestInteractionCallback_InteractionTypeBlockActions_Unmarshal(t *testing.T)
 		cb.BlockActionState.Values["section_block_id"]["multi_convos"].SelectedConversations,
 		[]string{"G12345"})
 }
+
+func TestInteractionCallback_ValidateToken(t *testing.T) {
```
kanata2, line 395: `(InteractionCallback).ValidateToken` returns true if both `verificationTokens` and `(InteractionCallback).Token` are an empty.
Is this the intended behavior?

kanata2 (commented):
Thanks for your PR!

sivchari:
Thanks for some reviews !
I will modify it based on the review I received.
I will change the verification token from the deprecated one to the following method using sigining secret.
https://api.slack.com/authentication/verifying-requests-from-slack
We will also verify the test more strictly.
May I ask you to be a bit more specific as I do not have a clear understanding of the meaning of the review you gave in lines 120 and 121?


Translated with www.DeepL.com/Translator (free version)

kanata2:
Ah, sorry.

> but Slack has two tokens.(the other is signing token.)

This is my misunderstanding.(not signing token but sigining **secret**).
So please ignore this comment.

By the way, if you use signing secret, we already have `slack.SecretsVerifier`.

sivchari:
Okay, thank you very much!
I submitted this PR because the slash command had a verification token method, and I thought it would be necessary here.
It's deprecated, so it's not necessary. I'm sorry.
If it is deprecated in the future, I will submit a PR to remove the verification token method from the slash command.
What do you think about this?

kanata2:
> I submitted this PR because the slash command had a verification token method, and I thought it would be necessary here.
It's deprecated, so it's not necessary. I'm sorry.

Don't worry.
If you want to use verification token again for some reasons, please re-open this PR.

> If it is deprecated in the future, I will submit a PR to remove the verification token method from the slash command.
> What do you think about this?

Cool.
Slack marks verification token as deprecated, but we can still use this for now.
So if you find any announce about this from Slack, please open a PR!
Thank you for your suggestion!

sivchari:
Sorry to delay replying message.
I gonna use slack.SecretsVerifier method for future updates.

OK!  When Slack announce it, I'll open a PR!
Thanks for some conversations!

PR #767, opened 2020-07-30
Add Header block

cjdenio:
This PR adds Slack's shiny new Header block! (https://api.slack.com/reference/messaging/blocks#header)

PR #887, opened 2021-01-18
Convert Header block correctly

takanabe:
# Background

Header block was added https://github.com/slack-go/slack/pull/767 but not converted correctly and treated as `slack.UnknownBlock`. This commit convert blocks to the Header block if they have 'header' type by block converter.

# Goal

* Fix a bug and support **Header** block

# Related information

* https://github.com/slack-go/slack/pull/767


kanata2 (changes requested):
Thank you :)

kanata2 (approved):

PR #888, opened 2021-01-21
add thread_ts to container

ianfhall:
When you press a button inside of a thread, Slack sends the thread timestamp inside of the container.  This is not mentioned in the [block actions documentation](https://api.slack.com/reference/interaction-payloads/block-actions), but the field is there and it is named `thread_ts`.

Screen Shot 2021-01-21 at 09 55 51

This commit adds the missing field to Container.  The field is omitted when empty (when you are not posting in a thread).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,8 @@ require (
 	github.com/gorilla/websocket v1.4.2
 	github.com/pkg/errors v0.8.0
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/stretchr/testify v1.2.2
+	github.com/stretchr/objx v0.3.0 // indirect
+	github.com/stretchr/testify v1.3.0
 )
 
 go 1.13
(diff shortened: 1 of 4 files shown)


brainexe:
+1 to merge this PR - I run in exactly the same issue: I get a channel, timestamp, but no thread-ts to make a proper reply for the given Button

kanata2:
Umm,  it seems that java-slack-sdk's container doesn't have `thread_ts` as well as slack-go/slack...🤔 
https://github.com/slackapi/java-slack-sdk/blob/main/slack-app-backend/src/main/java/com/slack/api/app_backend/interactive_components/payload/BlockActionPayload.java#L69-L80

@seratch 
I'm sorry for the abrupt message. Do you know anything about this?

drew-richardson:
@kanata2 thank you for looking at this PR and #945. This description of the scenario is correct (mine was incorrect on #945). Is there any additional information I can provide to help advance this PR?

kanata2 (approved):
Thanks!
I've confirmed in this example. https://github.com/kanata2/slack-go-sandbox/commit/a860818e684c9ed8ead03efeef5b4e346400f279

PR #890, opened 2021-01-26
Add support for emoji_changed event

ToshihitoKon:
added support EventAPI `emoji_changed` event

https://api.slack.com/events/emoji_changed

### DIFF
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -186,7 +186,6 @@ type TokensRevokedEvent struct {
 type EmojiChangedEvent struct {
 	Type           string      `json:"type"`
 	Subtype        string      `json:"subtype"`
-	Value          string      `json:"value"`
 	EventTimeStamp json.Number `json:"event_ts"`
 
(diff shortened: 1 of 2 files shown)


slackevents/inner_events.go
```diff
@@ -182,6 +182,24 @@ type TokensRevokedEvent struct {
 	Tokens tokens `json:"tokens"`
 }
 
+// EmojiChangedEvent is the event of custom emoji has been added or changed
+type EmojiChangedEvent struct {
+	Type           string      `json:"type"`
+	Subtype        string      `json:"subtype"`
+	Value          string      `json:"value"`
```
kanata2: This field is an empty when removed emoji

kanata2: Does Slack send us to different errors for SocketMode and RTM??

ToshihitoKon: thanks for review and fixed it!

kanata2 (approved):
👍

PR #893, opened 2021-02-09
Add `dispatch_action` to input blocks

evnsio:
This PR adds the `dispatch_action` field on InputBlock as described [here](https://api.slack.com/reference/block-kit/blocks#input). Setting this to true results in changes to the block dispatching a block_action payload to the server.


kanata2 (approved):
Thanks!

PR #894, opened 2021-02-10
Make UserID optional in GetUserProfile()

itkq:
`users.profile.get` API takes user parameter as optional.
https://api.slack.com/methods/users.profile.get

This change breaks GetUserProfile interface.
The current interface does not allow the bot to retrieve
its own user profile.

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/users.go
+++ b/users.go
@@ -632,9 +632,15 @@ func (api *Client) UnsetUserCustomStatusContext(ctx context.Context) error {
 	return api.SetUserCustomStatusContext(ctx, "", "", 0)
 }
 
+// GetUserProfileParameters are the parameters required to get user profile
+type GetUserProfileParameters struct {
+	UserID        string
+	IncludeLabels bool
+}
+
 // GetUserProfile retrieves a user's profile information.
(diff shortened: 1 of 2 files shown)


itkq:
I found [bots.info method](https://api.slack.com/methods/bots.info) which the bot can also retrieve its own user profile, but [the current interface](https://github.com/slack-go/slack/blob/f018c18733dddb42ae5c63fec75018591e8dd4f1/bots.go#L39) requires `bot` string as well...

kanata2 (approved):
Thank you!
As you mentioned, This changes break backward-compatibility.
So I plan to release this in the next minor version.

kanata2:
>  but the current interface requires bot string as well...

Slack API doc says the `bot` parameter optional, but when we tried this, it returned only `ok`...🤔 

```
$ curl -H'Authorization: Bearer <bot_token>' https://slack.com/api/bots.info
{"ok":true}
```

itkq:
Oh.. I got the same response, which means this method is not suitable for the use case above (I actually mentioned bots.info method without trying it out 🙄 ).

Issue #881, opened 2021-01-11
Cannot use emoji with mrkdwn section


JoseFMP:
I try to send a hello world message using Slack go API.
```

const token = "my token in here!"

func main(){
  api := slack.New(token)

  chanelId := `do some stuff here to find the channel ID, means the API client is kind of working as I can list the channels`
	
  textBlockObject := slack.NewTextBlockObject("mrkdwn", "hello world!", true, false)
  sectionBlock := slack.NewSectionBlock(textBlockObject, nil, nil)

  blocks := []slack.Block{sectionBlock}
  blockOptions := slack.MsgOptionBlocks(blocks...)
  _, _, errorSending := api.PostMessage(channelID, blockOptions)
}
```

And in Slack response, (HTTP 200) there is an upper layer error message there is just this: `invalid_blocks` and obviously the message is not posted.

Since the message I am sending is so extremely simple, what could be wrong!? 

### Versions
- Go: go1.15.6 linux/amd64
- slack-go/slack: v0.7.4

JoseFMP:
Seems is related to the `mrkdwn` type. If I set it to `plain_text` it works just fine. What is wrong? I can see in the Slack message builder, plenty of blocks text objects have the type `mrkdn`.
Any ideas?

JoseFMP:
I got to work around this issue by using mark down formatting and the option `false` to disable emojis. So this is really counter-intuitive and the error message from Slack API could be improved. Not even that, but the client itself could check this before sending the HTTP request.

So using emojis + mrkdown gives an error from backend.

kanata2:
Thanks for reporting.
It seems that Slack does not allow `emoji` property in section block now.

https://api.slack.com/reference/block-kit/blocks#section

Enviy:
In this example Slack does appear to support the use of `emoji` with `mrkdown` in `section` type blocks. However I still encounter errors when attempting to use `mrkdown`.
Using `mrkdown` returns this error while replacing it with `plain_text` successfully uses the formatted string and uses emoji.

[...]

PR #895, opened 2021-02-12
Create a validation function to check TextBlockObject values

sryoya:
##### Description
**What**
Created a new validation function that can be optionally used to validate the fields in TextBlockObject

**Why**
1. Updates #881 
If we want to catch this problem absolutely like adding the validation like when initializing TextBlockObject, we need to change the IO of the existing API. So, it's a suggestion of the workaround to catch this.

2. It's also useful to catch other problems in TextBlockObject values
For example, we can add a new validation that picks up an unexpected value is in `Type` field by typo.
Even without using this in the production code, the client can make it useful in their unit test codes.

##### API changes
It just created a new public function that doesn't mutate any data or states. Also, no change for the any existing API.

### DIFF
--- a/block_object.go
+++ b/block_object.go
@@ -2,6 +2,7 @@ package slack
 
 import (
 	"encoding/json"
+	"errors"
 )
 
 // Block Objects are also known as Composition Objects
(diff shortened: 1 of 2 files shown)


kanata2 (approved):
Thank you! Looks good to me :)

PR #898, opened 2021-02-20
Missing fields in Attachment struct

xNok:
When a link is added to a message, slack add an attachment containing specific fields. Those fields are missing from the current implementation of attachment

PR Content:
     * Add 4 fields to struct Attachment


kanata2 (approved):
Thanks! Looks good to me.

PR #900, opened 2021-02-25
Expose ReplyTo information in ack error events.

nolanlum:
This allows clients to better handle ack errors by linking them to the original outbound message. Clients can then implement their own retry/backoff logic using this information.


kanata2 (approved):
Thanks! LGTM.

PR #901, opened 2021-02-25
Make more examples directly runnable by taking values from the environment

parsley42:
I noticed several of the examples were already looking for values in the environment. I made these changes so I could run a few examples directly without having to copy and paste in my secrets, which I think is a good thing anyway because really, secrets almost never belong in code.


kanata2 (approved):
LGTM

PR #902, opened 2021-02-26
Add the new condition to the validation func for TextBlockObject

sryoya:
# What
- Add the new validation condition that checks whether the value in the `Type` field is valid to the validation func for the `TextBlockObject`

# Why
- There may be a case that a user puts invalid value to this field such as mistaking it for other Struct or having a simple typo. 

# Ref
 [official API doc](https://api.slack.com/reference/block-kit/composition-objects#text)


kanata2 (approved):

kanata2:
LGTM :)

Issue #899, opened 2021-02-20
Suggestion: use middlerware pattern for socketmode


xNok:
With the current implementation for socketmode we need to systematically create an event loop link in the example:
https://github.com/slack-go/slack/blob/master/examples/socketmode/socketmode.go. It is very verbose and requires duplicating that code in all our project. I was thinking of adding a `socketmodeHandler` that follows the middleware design pattern. 

This is a design choice so I wanted to discuss the idea before submitting a PR. As a matter of fact, I could also create a separate library, but why no including it directly in slack-go?

An example of how we would use such implementation
```
socketmodeHandler:= socketmode.NewsSocketmodeHandler()
socketmodeHandler.Handle(socketmode.EventTypeConnecting, middlewareOne(middlewareTwo(finalHandler)))
```

kanata2:
@xNok 
Sorry for slow response.
Currently, there are very few active maintainers of this library, so there are some concerns (for example expanding the area that needs to be maintained).
How about including this suggestion as an experimental feature?
(Personally, I think your suggestion is a good idea. I also think it would be nice to have a framework like Bolt provided by an official.)

xNok:
For sure, it would be nice if Slack were maintaining a similar library. I am ok with It to be an experimental feature, I am willing to maintain that feature and everything that is related to Socket Mode (use the  SocketMode tag or mention me if needed).

I am providing a tutorial repository on the side so you can be sure I will be looking into most of Slack features. https://github.com/xNok/slack-go-demo-socketmode. It will an extra layer of test for this library.

kanata2:
Thanks. I'll confirm your PR as soon as possible.

PR #904, opened 2021-02-27
Socketmode Middleware Design Pattern

xNok:
I recently suggested (#899) to implement a middleware design pattern to make the use of socketmode convenient. I believe this to be such an improvement to my own code that I implemented it right away.

The basic concept behind is that it Matches incoming Event against the list of registered Event and calls all handler that are expecting that specific event. Much as you declare your routes when writing an APIs.


kanata2:
Personally, I think it's a good idea. Let me consider whether this should be included in slack-go/slack.

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,197 @@
...
+	eventApiMap := make(map[slackevents.EventAPIType][]SocketmodeHandlerFunc)
+
+	return &SocketmodeHandler{
+		Client:                         client,
+		EventMap:                       eventMap,
+		EventApiMap:                    eventApiMap,
+		InteractionEventMap:            interactionEventMap,
+		InteractionBlockActionEventMap: interactionBlockActionEventMap,
+		Default: func(e *Event, c *Client) {
+			log.Printf("Unexpected event type received: %v\n", e.Type)
```
kanata2: Prefer to use client's logger.

```suggestion
			c.log.Printf("unexpected event type received: %v\n", e.Type)
```

examples/sockermode_handler/socketmode_handler.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/slack-go/slack/slackevents"
+	"github.com/slack-go/slack/socketmode"
+
+	"github.com/slack-go/slack"
+)
+
+func main() {
+	appToken := os.Getenv("SLACK_APP_TOKEN")
+	if appToken == "" {
+
```
kanata2: `os.Exit(1)` ?

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,197 @@
...
+	r.InteractionEventMap[et] = append(r.InteractionEventMap[et], f)
+}
+
+// Register a middleare function to use to handle an Interaction Block Action referenced by its ActionID
+func (r *SocketmodeHandler) HandleInteractionBlockAction(action_id string, f SocketmodeHandlerFunc) {
+	r.InteractionBlockActionEventMap[action_id] = append(r.InteractionBlockActionEventMap[action_id], f)
+}
+
+// Register a middleare function to use to handle an Event (from slackevents)
+func (r *SocketmodeHandler) HandleEventsAPI(et slackevents.EventAPIType, f SocketmodeHandlerFunc) {
```
kanata2: How about `HandleEvents`?

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,197 @@
...
+	r.EventMap[et] = append(r.EventMap[et], f)
+}
+
+// Register a middleare function to use to handle an Interaction
+func (r *SocketmodeHandler) HandleInteraction(et slack.InteractionType, f SocketmodeHandlerFunc) {
+	r.InteractionEventMap[et] = append(r.InteractionEventMap[et], f)
+}
+
+// Register a middleare function to use to handle an Interaction Block Action referenced by its ActionID
+func (r *SocketmodeHandler) HandleInteractionBlockAction(action_id string, f SocketmodeHandlerFunc) {
```
kanata2: actionID 😄

xNok:
@kanata2 I cannot reproduce the error on go 1.15. I triggered the actions on my side https://github.com/xNok/slack/actions. Maybe re-running the action may be enough

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,218 @@
...
+	Default SocketmodeHandlerFunc
+}
+
+// Handler have access to the event and socketmode client
+type SocketmodeHandlerFunc func(*Event, *Client)
+
+// Middleware accept SocketmodeHandlerFunc, and return SocketmodeHandlerFunc
+type SocketmodeMiddlewareFunc func(SocketmodeHandlerFunc) SocketmodeHandlerFunc
+
+func NewsSocketmodeHandler(client *Client) *SocketmodeHandler {
```
kanata2: `NewSocketmodeHandler` ? 😄

HunterL:
Hey xNok, love using a middleware pattern here. Noticed one quick thing while going over your PR
- The addition to examples seems to have a minor misspelling 
   `examples/sockermode_handler/socketmode_handler.go` should be 
   `examples/socketmode_handler/socketmode_handler.go`

Would be great to get this into `master`, thanks for your work

slackevents/inner_events.go
```diff
@@ -319,45 +319,47 @@ func (e MessageEvent) IsEdited() bool {
 		e.Message.Edited != nil
 }
 
+type EventAPIType string
```
kanata2: Would `EventsAPIType` be more appropriate?

socketmode/client.go
```diff
@@ -61,3 +61,8 @@ type Client struct {
 	debug bool
 	log   ilogger
 }
+
+// return a pointer to the slack API client
+func (c Client) GetApiClient() *slack.Client {
+	return &c.apiClient
```
kanata2: It seems good to change the type of `socketmode.apiClient` from a non-pointer to a pointer.

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,218 @@
...
+func (r *SocketmodeHandler) HandleInteractionBlockAction(action_id string, f SocketmodeHandlerFunc) {
+	r.InteractionBlockActionEventMap[action_id] = append(r.InteractionBlockActionEventMap[action_id], f)
+}
+
+// Register a middleare function to use to handle an Event (from slackevents)
+func (r *SocketmodeHandler) HandleEventsAPI(et slackevents.EventAPIType, f SocketmodeHandlerFunc) {
+	r.EventApiMap[et] = append(r.EventApiMap[et], f)
+}
+
+func (r *SocketmodeHandler) HandleSlashCommand(command string, f SocketmodeHandlerFunc) {
```
kanata2: Please add GoDoc 🙏

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,218 @@
...
+	r.SlashCommandMap[command] = append(r.SlashCommandMap[command], f)
+}
+
+// Register a middleare function to use as a last resort
+func (r *SocketmodeHandler) HandleDefault(f SocketmodeHandlerFunc) {
+	r.Default = f
+}
+
+// RunSlackEventLoop receives the event via the socket
+func (r *SocketmodeHandler) RunEventLoop() {
```
kanata2: How about returning an error? (just add `return r.Client.Run()` and returned type)

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,218 @@
+package socketmode
```
kanata2: Please rename socketmode_handler.go

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,218 @@
...
+	return false
+}
+
+// Dispatch interactions to the registered middleware
+func (r *SocketmodeHandler) interactionDispatcher(evt *Event) bool {
+	var ishandled bool = false
+
+	interaction, ok := evt.Data.(slack.InteractionCallback)
+	if !ok {
+		fmt.Printf("Ignored %+v\n", evt)
```
kanata2: Prefer to use socketmode client's logger.

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,218 @@
...
+		for _, f := range handlers {
+			go f(evt, r.Client)
+		}
+
+		ishandled = true
+	}
+
+	// Level 3 - interaction with actionID
+	block_actions := interaction.ActionCallback.BlockActions
+	// outmoded approach won`t be implemented
```
kanata2: 👌

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,218 @@
...
+	var ishandled bool = false
+
+	interaction, ok := evt.Data.(slack.InteractionCallback)
+	if !ok {
+		fmt.Printf("Ignored %+v\n", evt)
+		return false
+	}
+
+	// Level 1 - socketmode EventType
+	ishandled = r.socketmodeDispatcher(evt)
```
kanata2: How about returning `ishandled` here when `ishandled` is true?

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,218 @@
...
+		// If we registered an event
+		for _, f := range handlers {
+			go f(evt, r.Client)
+		}
+
+		ishandled = true
+	}
+
+	// Level 3 - interaction with actionID
+	block_actions := interaction.ActionCallback.BlockActions
```
kanata2: `blockActions`

socketmode/socketmode_eventhandler.go
```diff
@@ -0,0 +1,218 @@
...
+)
+
+type SocketmodeHandler struct {
+	Client *Client
+
+	EventMap                       map[EventType][]SocketmodeHandlerFunc
+	InteractionEventMap            map[slack.InteractionType][]SocketmodeHandlerFunc
+	InteractionBlockActionEventMap map[string][]SocketmodeHandlerFunc
+	EventApiMap                    map[slackevents.EventAPIType][]SocketmodeHandlerFunc
+	SlashCommandMap                map[string][]SocketmodeHandlerFunc
```
kanata2: [question]
Is it intended that it accepts multiple handlers for a single event, interaction, block action or slack command?

kanata2 (commented):
Sorry for slow response. I added some comments.

socketmode/socketmode_eventhandler.go
xNok: I want to be able to handle multiple levels: level 1 is the most generic and level 3 is the most user-friendly.

lvl 3 would be slash commands, shortcuts, workflow steps, etc.

I don't believe that I can create an exhaustive list handlers so that way if someone is stuck they can write middleware that deals with whatever comes in instead of opening an issue saying their use case is not possible.

socketmode/socketmode_eventhandler.go
xNok: Yes, as mentioned before I want to be as flexible as possible. Additionally, this is required to achieve separation of concerns. Controllers/Services are not aware of what other controllers registered.

xNok: I think it is worth mentioning that wrote a more in-depth tutorial here: https://github.com/xNok/slack-go-demo-socketmode. It may better illustrate the intent. Please feel free to comment on that as well.

socketmode/socketmode_eventhandler.go
xNok: There is several types of interaction that use actionID but it appears at different places in the CallbackInteraction. 

Our reference docs have a more detailed look at the payload structures for the different type sources. 

here is the list based Slack documentation:
* block_actions payloads are received when a user clicks a Block Kit interactive component.
* shortcut and message_actions payloads are received when global and message shortcuts are used.
* view_submission payloads are received when a modal is submitted.
* view_closed payloads are received when a modal is cancelled.

xNok: Ho, I think I got it wrong you meant:
`action_id string` -> `actionID string` 

I will be more careful about the camelCase

socketmode/socketmode_eventhandler.go
kanata2: Umm... 
For example,` action_id` is defined as unique, but if multiple handlers can be prepared for a single `action_id`, it may cause unintentional duplicate registrations and other disadvantages.

> Controllers/Services are not aware of what other controllers registered.

How about a way to return an error when registering duplicates, like the net/http package?

kanata2: https://github.com/golang/go/blob/master/src/net/http/server.go#L2429-L2455

xNok: I think this is the way forward it makes sense to me.

socketmode/client.go
xNok: I did not know if it would have an unexpected effect on top of that `socketmode.apiClient` is not exported so it cannot be used outside of the package

kanata2: Sorry for the confusion. The following is my suggestion:

```go
type Client struct {
    ...
    apiClient *slack.Client
}

func (c *Client) GetApiClient() *slack.Client {
  return c.apiClient
}
```

xNok:
Hi @kanata2, I've completed the refactoring. Let me know if anything else needs to be changed.

kanata2:
Ah, sorry for my slow response....
I'll confirm this weekend 🙏

socketmode/socketmode_eventhandler.go
kanata2: It seems that the same typo is still present in several places :weary:

socketmode/socketmode_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	if !ok {
+		r.Client.log.Printf("Ignored %+v\n", evt)
+		return false
+	}
+
+	// Level 1 - socketmode EventType
+	ishandled = r.socketmodeDispatcher(evt)
+
+	// Level 2 - SlackCommand by name
+	if handler, ok := r.SlashCommandMap[slashCommandEvent.Command]; ok {
```
kanata2, line 234: Since the slash command is tied to the actual command string (like `/command`), I think that it should not be handled by the event type. ~~(Or if you really want to handle them, you should process level 2 before level 1.)~~

kanata2: I think that If you want to apply middleware functioins to all slash commands, you might want to wrap each handler functions by middleware functions.

kanata2 (commented):
It looks generally good, but I made a few comments.

socketmode/socketmode_handler.go
xNok: The order doesn't matter since each dispatcher starts a goroutine for each matching event. One event can be handler both by level 1 and level 2 if they exist.

Ideally, the level 1 should not be used. But some interesting events like `shortcuts` that work the same way as slash commands are not yet implemented. Therefore, to filter shortcuts by name you can create a middleware function that patches the missing feature.

In a way, it is possible to write all the dispatcher logic in a recursive way, When you want to handle a slash command you register a middleware that captures that specific command. In this scenario, one would only have `socketmodeDispatcher`. The issue with this option is that if you have two commands then you have two middleware functions and both would be called when a `slash_command` event comes in. But only one will actually do something.

kanata2: > The order doesn't matter since each dispatcher starts a goroutine for each matching event. One event can be handler both by level 1 and level 2 if they exist.

Sorry, this is my misunderstanding. 

> But some interesting events like shortcuts that work the same way as slash commands are not yet implemented. Therefore, to filter shortcuts by name you can create a middleware function that patches the missing feature.

I'd rather avoid misuse than deal with unimplemented features such as shortcuts.

> In a way, it is possible to write all the dispatcher logic in a recursive way, When you want to handle a slash command you register a middleware that captures that specific command. In this scenario, one would only have socketmodeDispatcher. The issue with this option is that if you have two commands then you have two middleware functions and both would be called when a slash_command event comes in. But only one will actually do something.

Could you explain it in more detail?

xNok: Yes, instead of using Maps and dispatchers, like this:

```go
// Register a middleware or handler for an Event (from slackevents)
func (r *SocketmodeHandler) HandleEvents(et slackevents.EventsAPIType, f SocketmodeHandlerFunc) {
	r.EventApiMap[et] = append(r.EventApiMap[et], f)
}
```

You can achieve the same goal with middleware function

```go
// Register a middleware or handler for an Event (from slackevents)
func (r *SocketmodeHandler) HandleEvents(et slackevents.EventsAPIType, f SocketmodeHandlerFunc) {
	r.Handle(EventTypeInteractive, func(e *Event, c *Client){
             interaction, _ := evt.Data.(slack.InteractionCallback)
             if interaction.type == et {
                 f(e, c)
             }
        })
}
```

I thought that what you meant by

> I think that If you want to apply middleware functioins to all slash commands, you might want to wrap each handler functions by middleware functions.

LiorLieberman:
Any updates regarding the status of this PR?

kanata2:
Sorry for the delay. I'll check it again later 😢

notwedtm:
@kanata2 is there anything that can be done to assist you in getting this PR merged in? It's been open since April :(

socketmode/socketmode_handler.go
kanata2: @xNok 

Sorry for slow response.

As mentioned above, in most cases, Level 3 works as intended for the library user,
but in level 1 or 2, the possibility of unintended behavior is likely to increase.

The official library "bolt" also does not allow registering multiple handlers to for a single event(or action_id, slash command, ...),
so it can be confusing for people using other languages like Java, Python.

ref: https://github.com/slackapi/java-slack-sdk/blob/main/bolt/src/main/java/com/slack/api/bolt/App.java

I think it would be easier to use something closer to the bolt API, but... What do you think?

kanata2:
@xNok Could you please add a note in the README that this is an EXPERIMENTAL feature?

xNok:
@kanata2 I added the comment in the readme

As a side note `make pr-prep` doesn't lint recursively. It took me a while to understand why golangci-lint was not happy.

I'll do a separate PR to fix that.

socketmode/socketmode_handler.go
kanata2: @xNoK What do you think??

xNok: This is a good point I did not look into the bold library. 

I will refactor to make it closer to the bolt API

By any chance do you know if bolt library implement the same spec in JAVA and Python?

xNok: I have to admit that I like the JAVA implementation that separates the handler registration in `App` and `SocketModeApp` that is in charge of collecting the events. 

```
  public static void main(String[] args) throws Exception {
    String botToken = System.getenv("SLACK_BOT_TOKEN");
    App app = new App(AppConfig.builder().singleTeamBotToken(botToken).build());
    String appToken = System.getenv("SLACK_APP_TOKEN");
    SocketModeApp socketModeApp = new SocketModeApp(appToken, app);
    socketModeApp.start();
  }
```

kanata2: > By any chance do you know if bolt library implement the same spec in JAVA and Python?

I'm not very confident so I'll take a look at the implementation.

zotrix: maybe make sense merge this pr as experimental and allow use middleware. In parallel thinking about more perfect solution like bolt?

slackevents/inner_events.go
```diff
@@ -409,69 +409,71 @@ func (e MessageEvent) IsEdited() bool {
...
-	ChannelUnarchive = "channel_unarchive"
+	ChannelUnarchive = EventsAPIType("channel_unarchive")
 	// ChannelLeft is sent when a channel is left.
-	ChannelLeft = "channel_left"
+	ChannelLeft = EventsAPIType("channel_left")
 	// ChannelRename is sent when a channel is rename.
-	ChannelRename = "channel_rename"
+	ChannelRename = EventsAPIType("channel_rename")
 	// ChannelIDChanged is sent when a channel identifier is changed.
 	ChannelIDChanged = "channel_id_changed"
```
obalunenko, line 454: Missed EventsAPIType^^

socketmode/socketmode_handler.go
kanata2: @xNok Is there anything I can do to help?

mattgialelis:
Any updates on getting this merged in or anything i can help with to speed it up. @xNok @kanata2 Just would be great to be able to use the releases than tagged to a specific commit

arulk:
IS this PR going to be merged ?

zchee:
@xNok Hi, this is newer slack-go maintainer.
I so very interesting this approach. So I'll review soon.

@kanata2 When If I done of review (and fixed any pointed out by contributor if exist), Can I merge at my discretion?

kanata2:
@zchee 
I think the problem of this thread does not resolve yet.
So I had been considering preparing a new implementation based on this PR. (Sorry for delay...)
However, if you think it is ready to be merged, you are welcome to do so.

mattgialelis:
Just to find out whats the future of this PR if you're planning to implement something yourself @kanata2 ? And id be curious when would the new implantation be ready to use ?  

As having the functionality of this or another middleware makes it a lot simpler to interact with socketmode without the need for alot of boiler plate code just to structure an event handler ?

zchee:
@kanata2 have you any local develop branch, or something?
Also,

> I think the problem of this thread does not resolve yet

Where is you said "problem"? (just for confirming

kanata2:
I'm plan to merge this and release v0.11.0 in a few days.

zchee:
@kanata2 What's the status of merge this?

kanata2:
@zchee I believe it is good to merge. What do you think?

zchee:
@kanata2 I think so, too.

zchee (approved):

kanata2 (approved):

kanata2:
I'll fix linter's error in another PR.

PR #905, opened 2021-03-05
Add client callback for API warning messages

ghost:
**Problem:** There's no way to surface Slack's API warnings in application code from the library.

This PR exposes a new client `Option` that registers a callback for every response that comes back with warning information. This suits the use case for logging/tracking warnings as they appear, without affecting the request/response interface of each endpoint. The nodejs sdk seems to take [a similar approach](https://github.com/slackapi/node-slack-sdk/blob/0456eb09985832cec5cbf10f150dced9a10fa92c/packages/web-api/src/WebClient.ts#L172-L175).

```go
onWarning := func(warning *slack.Warning, path string, values url.Values) {
  // Log the warning here.
}
client := slack.New(token, slack.OptionOnWarning(onWarning))
```

An easy way to test this and see warnings is to update [outgoing headers](https://github.com/slack-go/slack/blob/bf8fbd57a52941ee9eb35653eca7bd39a9877fe6/misc.go#L107) to have `; charset=utf-8` at the end.

**Relevant docs/reading:**
- https://api.slack.com/web#slack-web-api__evaluating-responses
- https://api.slack.com/changelog/2020-01-deprecating-antecedents-to-the-conversations-api#warnings
- https://medium.com/slack-developer-blog/well-wishes-and-warnings-in-the-web-api-489456837220

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -26,6 +26,7 @@ import (
 type SlackResponse struct {
 	Ok               bool             `json:"ok"`
 	Error            string           `json:"error"`
+	Warning          string           `json:"warning"`
 	ResponseMetadata ResponseMetadata `json:"response_metadata"`
 }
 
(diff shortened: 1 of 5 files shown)


ghost:
Closing to keep downstream changes from appearing in this PR.

PR #907, opened 2021-03-10
Mior wording change

aubelsb2:
Just a string change on an exmple


kanata2 (changes requested):
This example deletes after uploading. So in my opinion, this message is inappropriate.

aubelsb2:
Ah whoops you're right. I didn't read it properly. Sorry was rushing.

Issue #903, opened 2021-02-27
AuthTest Returns Invalid_Auth for Bot User Token


moss-n:
### What happend
I attempted to use the UploadFile function to upload a file to a workspace. When calling UploadFile, the function returns an error "invalid_auth". I confirmed that the file:write scope was provided to the bot user token, and I was able to successfully use the token when calling GetUserInfo. I noticed that GetUserInfo does not first call AuthTest before making the request, unlike FileUpload

### Expected behavior
Expected that when FileUpload was called with the given params, it would upload the text file I provided in the params argument 

### Steps to reproduce
This is my program 

```
import (
	"fmt"
	"github.com/slack-go/slack"
)

func main() {
	api := slack.New("TOKEN")

	params := slack.FileUploadParameters{
		File: "test.txt",
	}
	file, err := api.UploadFile(params)

	if err != nil {
		fmt.Printf("%s\n", err)
		return
	}

	fmt.Printf("Name: %s, URL: %s\n", file.Name, file.URL)
}
```

### Versions
- Go: go version go1.15.8 
- slack-go/slack: latest

aubelsb2:
I'm also getting the same issue with both an user token and a bot token.. My permission are as follows:
Bot scopes: (xoxb-)
channels:history channels:join channels:read chat:write files:read files:write users:read
User scopes: (xoxp-)
channels:history channels:read chat:write files:read files:write users:read

The code is as follows:
```
	sf, err := api.UploadFile(slack.FileUploadParameters{
		Reader:   bytes.NewReader(bimg.Bytes()),
		Filename: fmt.Sprintf("%s.png", rightnow.Format("2006-01-02T15")),
		Channels: []string{wfo_wfh.ChannelId},
		InitialComment: "Auto generated",
	})

```

The token is being used for other things successfully.

aubelsb2:
Ok a little digging, uploads only work if you use the "Content" method right now. I tested my token against with CURL and it worked as expected so it's a library issue.

aubelsb2:
```
 curl -F file=@slack_api_logo_vogue.png -H "Authorization: Bearer xoxp-" -F channels=ASDF https://slack.com/api/file
s.upload
```

aubelsb2:
Okay as far as I can tell it's because the authorization token is missing
https://github.com/slack-go/slack/pull/908

aubelsb2:
Ok I think I solved the issue:
https://github.com/slack-go/slack/pull/908

[...]

Issue #857, opened 2020-11-30
Deprecating usage of token as a query string parameter in Web API requests


eliot-stripe:
Slack announcement (2020-11-30): https://api.slack.com/changelog/2020-11-no-more-tokens-in-querystrings-for-newly-created-apps

Based on my initial splunking, it looks like `getResource` passes the the request parameters as a URL parameters: https://github.com/slack-go/slack/blob/853f55b35241219807941e42914bfc7905afb4c8/misc.go#L259-L268

The callsites of this include `token` as one of those URL parameters: https://github.com/slack-go/slack/blob/master/chat.go#L696-L714

It's possible that other http methods do as well, but I haven't done an exhaustive search of the codebase.

kanata2:
Fixed in v0.8.3

PR #908, opened 2021-03-10
Seems the token was missing from the request.

aubelsb2:
As far as I can tell it's because of a token is missing.

### DIFF
--- a/files.go
+++ b/files.go
@@ -320,12 +320,12 @@ func (api *Client) UploadFileContext(ctx context.Context, params FileUploadParam
 		values.Add("content", params.Content)
 		err = api.postMethod(ctx, "files.upload", values, response)
 	} else if params.File != "" {
-		err = postLocalWithMultipartResponse(ctx, api.httpclient, api.endpoint+"files.upload", params.File, "file", values, response, api)
+		err = postLocalWithMultipartResponse(ctx, api.httpclient, api.endpoint+"files.upload", params.File, "file", api.token, values, response, api)
 	} else if params.Reader != nil {
 		if params.Filename == "" {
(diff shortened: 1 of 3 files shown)


aubelsb2:
Probably handled in the http client.

aubelsb2:
Verified to fix my issue. Seems they are checking for unnecessary parameters or at least token being in the wrong location.

kanata2:
@aubelsb2

I commented related issue. https://github.com/slack-go/slack/issues/903#issuecomment-800505141
Could you fix other codes where same problems occur too?(if exists)
If it is difficult, please let me know so I can find the time to do it.

bailantaotao:
Is there any update for this?

aubelsb2:
When I get a weekend spare I can work on it.

raylas:
Would also love to see this merged ☝🏼

joshm91:
Looks like #921 fixes this issue properly. This library is pretty much broken to any newcomers at the moment so it'd be great to get the fix merged ASAP. Thanks!

kanata2 (approved):

kanata2:
Relased in v0.8.3

arran4:
Following up from personal account. Does effort still need to be done by me? I feel like the change to remove query param token from the library will require significant change.

kanata2:
In my recognition, #857 was resolved in v0.8.3.

However it may be better to use application/json as Slack's recommendation(the followings).

> Some methods, like chat.postMessage and dialog.open feature arguments that accept an associative JSON array. These methods can be difficult to properly construct when using a application/x-www-form-urlencoded Content-type, and we strongly recommend using JSON-encoded bodies instead.

PR #914, opened 2021-03-19
Update ExcludeArchived type to bool & match others

mxygem:
Small PR to update `ExcludeArchived`'s type in `GetConversationsParameters` in `conversation.go` to bool so that it matches all other instances of `ExcludeArchived` in the package. It changes the signature of `GetConversationsParameters` but makes it more communicative as to what the value should be.


kanata2:
Thanks. This PR is same as #832.

mxygem:
Moving the comment here instead

@kanata2 Thanks for the reply! Happy to help out with other stuff as needed. I asked in Gitter about it, too.

PR #916, opened 2021-03-26
slacktest: Save thread_ts to message

StupidScience:
**Problem**: There is no way to check if message was sent to thread.

I just save `thread_ts` to message struct so we'll be able to check it later with `GetSeenOutboundMessages()`


kanata2 (approved):
Thanks!

PR #918, opened 2021-04-01
Add timepicker to UnmarshalJSON

tapih:
`timepicker` does not exist under the `switch` in `UnmarshalJSON`, and the socket mode client failed to parse incoming messages saying `unsupported block element type timepicker`.


kanata2 (approved):
Looks good. Thanks.

PR #919, opened 2021-04-04
Add code samples for interactive mode.

yasutakatou:
I was trying to add an interactive message funciton in socket mode to my bot, but the sample didn't show how to do it.
After some search, I figured out how to do it and added it. I think to add example for same case developers.

- [x ] is it a convenience method? (no new functionality, streamlines some use case)


kanata2:
Thanks.
It seems that this is not socketmode specific process.
Instead, if there is something missing, could you please add an example as interactive components? (see also: examples/blocks/blocks.go)

yasutakatou:
I'm sorry.
I'm sorry for sending you such a strange pull request.

kanata2:
Please don't worry about it.  Thank you!

PR #920, opened 2021-04-05
Add team integration logs API

naormatania:
Add support for https://api.slack.com/methods/team.integrationLogs 
Add unittest to check it works as expected
Also ran `make pr-prep` to check it's ready for review

### DIFF
--- a/team.go
+++ b/team.go
@@ -9,6 +9,8 @@ import (
 const (
 	DEFAULT_LOGINS_COUNT = 100
 	DEFAULT_LOGINS_PAGE  = 1
+	DEFAULT_LOGS_COUNT   = 100
+	DEFAULT_LOGS_PAGE    = 1
 )
 
 type TeamResponse struct {
@@ -43,6 +45,25 @@ type Login struct {
 	Region    string `json:"region"`
 }
 
+type LogResponse struct {
+	Logs   []Log `json:"logs"`
+	Paging `json:"paging"`
+	SlackResponse
+}
+
+type Log struct {
+	ServiceID   string `json:"service_id"`
+	ServiceType string `json:"service_type"`
+	AppID       string `json:"app_id"`
+	AppType     string `json:"app_type"`
+	UserID      string `json:"user_id"`
+	UserName    string `json:"user_name"`
+	Date        string `json:"date"`
+	ChangeType  string `json:"change_type"`
+	Reason      string `json:"reason"`
+	Scope       string `json:"scope"`
+}
+
 type BillableInfoResponse struct {
 	BillableInfo map[string]BillingActive `json:"billable_info"`
 	SlackResponse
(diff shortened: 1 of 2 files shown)


team.go
```diff
@@ -43,6 +45,25 @@ type Login struct {
 	Region    string `json:"region"`
 }
 
+type LogResponse struct {
+	Logs   []Log `json:"logs"`
+	Paging `json:"paging"`
+	SlackResponse
+}
+
+type Log struct {
```
kanata2, line 54: Please add `Channel` field.

kanata2 (commented):
Looks good except for the parts I pointed out!

team.go
naormatania: Thanks for the feedback
done

kanata2:
@naormatania 
When I tried it, I came across strange behavior.
Slack documentation says that the `service_id` field is string, but actual response's one is not string but int... So slack-go/slack could not decode.
I'll confirm Slack support.

naorsec:
@kanata2 has slack support answered about this issue ?

kanata2:
Sorry for delay.
According to support, this is documentation's mistake.

naormatania:
@kanata2 it seems like slack updated their documentation that `service_id` can be both int or string.
I changed PR to handle both cases. let me know what you think. thanks !

kanata2:
@naormatania 
I confirmed support again.
According to support's response, the type of `service_id` is only integer,
so we do not seem to need to consider the string type. :sob:

naormatania:
Thanks for the help @kanata2 ! I changed it to `int` now

kanata2:
@naormatania

thanks!
I found a new undocumented behavior...
It seems that the `user_id` can be 0 (= a number) like following example.

```json
        {
            "user_id": 0,
            "user_name": "slack",
            "date": "1595673313",
            "change_type": "disabled",
            "reason": "channel_archive",
            "service_id": 123456789,
            "service_type": "Incoming WebHooks",
            "scope": "incoming-webhook"
        }
```

team.go
```diff
@@ -43,6 +45,26 @@ type Login struct {
...
+	ServiceID   int    `json:"service_id"`
+	ServiceType string `json:"service_type"`
+	AppID       string `json:"app_id"`
+	AppType     string `json:"app_type"`
+	UserID      string `json:"user_id"`
+	UserName    string `json:"user_name"`
+	Channel     string `json:"channel"`
+	Date        string `json:"date"`
```
dorsec, line 62:
```go
DateTime   JSONTime `json:"date,string"`
```

team.go
```diff
@@ -43,6 +45,26 @@ type Login struct {
...
+	Paging `json:"paging"`
+	SlackResponse
+}
+
+type Log struct {
+	ServiceID   int    `json:"service_id"`
+	ServiceType string `json:"service_type"`
+	AppID       string `json:"app_id"`
+	AppType     string `json:"app_type"`
+	UserID      string `json:"user_id"`
```
dorsec, line 59: stringOrInt

dorsec (changes requested):

PR #938, opened 2021-05-07
Add missing token values when uploading a file

### LINKED ISSUE #936: Uploading a file with Content param as bot user fails with not_authed error
### What happend

Uploading a file with `Content` param (not `File` or `Reader`)  as bot user fails with `not_authed` error.
A valid bot user token is provided and the token has proper permission for uploading file (which is `files:read`).

I dug into this issue and found out that token is not being sent as expected.
If you look at the following `UploadFileContext` function in `files.go`..
```
// UploadFileContext uploads a file and setting a custom context
func (api *Client) UploadFileContext(ctx context.Context, params FileUploadParameters) (file *File, err error) {
	// Test if user token is valid. This helps because client.Do doesn't like this for some reason. XXX: More
	// investigation needed, but for now this will do.
	_, err = api.AuthTest()
	if err != nil {
		return nil, err
	}
	response := &fileResponseFull{}
	values := url.Values{}
	if params.Filetype != "" {
		values.Add("filetype", params.Filetype)
	}
	if params.Filename != "" {
		values.Add("filename", params.Filename)
	}
	if params.Title != "" {
		values.Add("title", params.Title)
	}
	if params.InitialComment != "" {
		values.Add("initial_comment", params.InitialComment)
	}
	if params.ThreadTimestamp != "" {
		values.Add("thread_ts", params.ThreadTimestamp)
	}
	if len(params.Channels) != 0 {
		values.Add("channels", strings.Join(params.Channels, ","))
	}
	if params.Content != "" {
		values.Add("content", params.Content)
		err = api.postMethod(ctx, "files.upload", values, response)
	} else if params.File != "" {
		err = postLocalWithMultipartResponse(ctx, api.httpclient, api.endpoint+"files.upload", params.File, "file", api.token, values, response, api)
	} else if params.Reader != nil {
		if params.Filename == "" {
			return nil, fmt.Errorf("files.upload: FileUploadParameters.Filename is mandatory when using FileUploadParameters.Reader")
		}
		err = postWithMultipartResponse(ctx, api.httpclient, api.endpoint+"files.upload", params.Filename, "file", api.token, values, params.Reader, response, api)
	}

	if err != nil {
		return nil, err
	}

	return &response.File, response.Err()
}
```
`api.token` is being passed onto requests when handling `File` and `Reader` uploads but not when uploading using `Content`. I followed through the `api.postMethod` and subsequent function calls  and could not find any use of `api.token` down the line.
The solution is simple. Adding `values.Add("token", api.token)` right before `api.postMethod(..)` should work.
If there's something I'm missing, please tell me. Otherwise, I can draw up a PR to fix this.

### Expected behavior

The file should be uploaded without auth failure.

### Steps to reproduce

Run following code. (replace `<TOKEN>` and `<CHANNEL>` to proper value)
```
package main

import (
	"fmt"
	"os"

	"github.com/slack-go/slack"
)

func main() {
	api := slack.New("<TOKEN>", slack.OptionDebug(true))
	params := slack.FileUploadParameters{
		Title:    "Batman Example",
		Filetype: "txt",
		Filename: "example.txt",
		Content:  "Nan Nan Nan Nan Nan Nan Nan Nan Batman",
		Channels: []string{"<CHANNEL>"},
	}
	file, err := api.UploadFile(params)
	if err != nil {
		fmt.Printf("%s\n", err)
		return
	}
	fmt.Printf("Name: %s, URL: %s\n", file.Name, file.URL)
}
```
And it fails like this.
```
slack-go/slack2021/05/04 19:02:53 slack.go:125: Challenging auth...
not_authed
```

### Versions
- Go: 1.16
- slack-go/slack: 0.9.0

lch88:
### Summary

Fix #936 by adding `token` query parameter when uploading a file using `Content`.

### DIFF
--- a/files.go
+++ b/files.go
@@ -316,6 +316,7 @@ func (api *Client) UploadFileContext(ctx context.Context, params FileUploadParam
 	}
 	if params.Content != "" {
 		values.Add("content", params.Content)
+		values.Add("token", api.token)
 		err = api.postMethod(ctx, "files.upload", values, response)
 	} else if params.File != "" {
 		err = postLocalWithMultipartResponse(ctx, api.httpclient, api.endpoint+"files.upload", params.File, "file", api.token, values, response, api)


kanata2 (approved):
LGTM. Thanks!

PR #942, opened 2021-06-01
slacktest: Fix broken json if newline in text

StupidScience:
*Problem:* If there is newline in message's text, slacktest will generate broken json and then will fall with error on any `SendMessage...` method due to json syntax error.

So I introduce anonymous structure  and encode it then in test handler instead of `fmt.Sprintf(...)`


StupidScience:
Tests unexpectedly failed at absolutely not related to changes place.

StupidScience:
Hello @kanata2,

Sorry for bothering would you be able to give this a review?

kanata2 (approved):
Sorry for delay... Thanks!

PR #952, opened 2021-07-14
fix button sample

### LINKED ISSUE #937: Buttons example not showing selected Value field
### What happend
Copy paste buttons example returns empty value.
```
2021/05/06 16:05:25 Message with buttons sucessfully sent to channel xxx at xxxx
2021/05/06 16:05:35 Message button pressed by user xxxx with value     [EXPECTED STRING HERE]
```

### Expected behavior
Payload is parsed with the expected configured values

### Steps to reproduce
Copy paste 
https://github.com/slack-go/slack/blob/master/examples/buttons/buttons.go
Go build and publish to be able to get the interaction handler.

### Versions
- Go: 1.16
- slack-go/slack: v0.9.0

sivchari:
related #937 

I've fixed it.

### DIFF
--- a/examples/buttons/buttons.go
+++ b/examples/buttons/buttons.go
@@ -61,5 +61,5 @@ func actionHandler(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		fmt.Printf("Could not parse action response JSON: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


kanata2 (approved):
Cool. Thanks 👍

PR #954, opened 2021-07-16
Audit log fixes 

unkempthenry:
Fixes a couple things wrt to audit logs:

1) Fix "Limit" option. The Number of results returned is specified by the `limit` URL param, previously this was getting switched to `count` so the API would always return the default ( https://api.slack.com/admins/audit-logs ).

2) Stop adding token to request URL params. This was passing a token in both the request Headers and as a URL param.  Based on other issues here, it sounds like Slack deprecated including tokens in URLs, so here it was being included and probably ignored.  Requests would still work though as the token was also being included in the header. 

##### PR preparation

✅ 

N/A. Just a couple fixes

### DIFF
--- a/audit.go
+++ b/audit.go
@@ -115,7 +115,7 @@ func (api *Client) GetAuditLogsContext(ctx context.Context, params AuditLogParam
 		"token": {api.token},
 	}
 	if params.Limit != 0 {
-		values.Add("count", strconv.Itoa(params.Limit))
+		values.Add("limit", strconv.Itoa(params.Limit))
 	}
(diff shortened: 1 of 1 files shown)


unkempthenry:
@kanata2 , wondering if you'll have time to review this soon (sorry to bug you)

kevinpark1217:
Hmm looks like it's taking them a while. Maybe you need to create an issue and link it to this to alert them?

kanata2:
Sorry for slow response. I'll take a look this weekend :pray:

kanata2 (approved):
LGTM!

PR #955, opened 2021-07-17
Add context aware UnfurlMessage

ptxmac:
The MessageTimeStamp was incorrectly marked as a number. It should be
a string as any other timestamp, and it required for Unfurl to work

### DIFF
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -165,7 +165,7 @@ type LinkSharedEvent struct {
 	User             string        `json:"user"`
 	TimeStamp        string        `json:"ts"`
 	Channel          string        `json:"channel"`
-	MessageTimeStamp string        `json:"message_ts"`
+	MessageTimeStamp json.Number   `json:"message_ts"`
 	ThreadTimeStamp  string        `json:"thread_ts"`
 	Links            []sharedLinks `json:"links"`
 }
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -165,7 +165,7 @@ type LinkSharedEvent struct {
 	User             string        `json:"user"`
 	TimeStamp        string        `json:"ts"`
 	Channel          string        `json:"channel"`
-	MessageTimeStamp json.Number   `json:"message_ts"`
+	MessageTimeStamp string        `json:"message_ts"`
 	ThreadTimeStamp  string        `json:"thread_ts"`
 	Links            []sharedLinks `json:"links"`
 }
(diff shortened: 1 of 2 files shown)


slackevents/inner_events.go
```diff
@@ -81,7 +81,7 @@ type LinkSharedEvent struct {
 	User             string        `json:"user"`
 	TimeStamp        string        `json:"ts"`
 	Channel          string        `json:"channel"`
-	MessageTimeStamp json.Number   `json:"message_ts"`
+	MessageTimeStamp string        `json:"message_ts"`
```
kanata2: json.Number, string, and int are currently mixed in this library.
 It would be best if it could be aligned to one of them, but...

Number is probably more realistic.

kanata2: Is there any reason why it has to be a string and not json.Number?

ptxmac: the api returns a string, and golang's json encoding does not recognize a quoted number (`{"message_ts":"123"}`) as a number - so the current one just gives an error

kanata2: The `json.Number` type can handle quoted number correctly.
This is an example. https://play.golang.org/p/0px4NuQ3nKE

So, it is possible that the problem is elsewhere. 🤔

ptxmac: You're right... it's been two months since I made the PR, so I misremembered. It's because the "timestamp" is not really a number, it's two numbers with a dot, like `"message_ts":"1631912836.000600"`. It should never be treaded as a number

kanata2: Ah, I see.
If you want to treat this field as string in your code, you can use `(json.Number).String()`.

https://pkg.go.dev/encoding/json#Number.String

ptxmac: true, but that seems a very silly way to do it - slack provides no guarantee that `ts` is a number - so they're free to change to so any other string as they like - I believe it's a mistake that number was every introduced, and all `*_ts` should be converted to strings

kanata2: Your point makes sense.
I suspect that the introduction of the string and number (and json.Number) types probably has something to do with the behavior of the Slack API historically.
Since changing types breaks compatibility for users, I think it would be ideal to do it all together somewhere.

kanata2: How about discussing this in issues or discussions?

ptxmac: Agreed.

When I made the PR the old code didn't work - I think maybe the API returned non-number for the `LinkSharedEvent` struct. But it's been 2 months, so I don't recall exactly.

I suspect I'm the only one using this part of the API since no one else seems to hit the problem

ptxmac: I've reverted this part for now

kanata2: Thanks 🙇 

I'll check later whether this method works correctly or not.

PR #968, opened 2021-08-27
[CI] Add test setup for Go1.17 environment

sryoya:
# What
Updated the CI to execute the tests in the Go1.17 environment as well.

# Why
I'm not sure about the policy of this repository for Go1.17, but I think we should ensure that the build and tests work fine even in this version.


kanata2 (approved):
Thanks!

Issue #963, opened 2021-08-16
Isn't it better for compatibility and simplicity that socketmode.Client.apiClient is an embedded struct?


hnw:
### Description

In `socketmode/client.go`, `socketmode.Client` has the member `apiClient` which type is `slack.Client`, but it's not an embedded struct. (contrary to the comment in source code )

```go
type Client struct {
	// Client is the main API, embedded
	apiClient slack.Client
(snip)
}
```

On the other hand, many structs in this project has `slack.Client` as embedded struct. For example, `slack.RTM` is such one.

```go
type RTM struct {
        // Client is the main API, embedded
        Client
(snip)
}
```

Because of the embedded struct, We can use all methods of `slack.Client` from `slack.RTM`. For Instance, `rtm.Debugf()` 
is used 13 times in `slack/websocket_managed_conn.go`.

For another example, we can call not only `api.PostMessage()` but also `rtm.PostMessage()`. It is very convenient for the 
 users of this library.

In addition, `socketmode.Debugf()` is very similar to `slack.Debugf()` and I think it's better to be merged in terms of simplicity. This can be easily achieved by modifying apiClient to an embedded structure.

### Suggestion

In `socketmode/client.go`, I think it's better to modifying apiClient to an embedded structure.

```go
type Client struct {
	// Client is the main API, embedded
	slack.Client
(snip)
}
```

This change exposes some member of `socketmode.Client`. But it's intended change and similar to other structs in this project, so I think no problem.

kanata2:
Thank you for the suggestion and PR. I think it's good.

kanata2:
merged

PR #971, opened 2021-09-07
Change the regular field `apiClient` in `socketmode.Client` to an embedded struct.

hnw:
Hi! As explained in #963, it seems natural for socketmode client struct to provide main API via embedded struct `slack.Client` like an RTM struct.

This pull request changes the regular field `apiClient` in `socketmode.Client` to an embedded struct.


kanata2 (approved):
Cool. Thanks!

Issue #974, opened 2021-09-24
Unable to produce an input block with the ability to select users


karl-johan-grahn:
### What happened
I am unable to reproduce this modal payload from the Block Kit Builder:
```
{
	"type": "modal",
	"title": {
		"type": "plain_text",
		"text": "My App",
		"emoji": true
	},
	"submit": {
		"type": "plain_text",
		"text": "Submit",
		"emoji": true
	},
	"close": {
		"type": "plain_text",
		"text": "Cancel",
		"emoji": true
	},
	"blocks": [
		{
			"type": "input",
			"element": {
				"type": "users_select",
				"placeholder": {
					"type": "plain_text",
					"text": "Select a user",
					"emoji": true
				}
			},
			"label": {
				"type": "plain_text",
				"text": "Choose user",
				"emoji": true
			}
		}
	]
}
```

Tried this:
```
label := slack.NewTextBlockObject(slack.PlainTextType, "Choose user", false, false)
blockElement := slack.NewOptionsSelectBlockElement(slack.OptTypeUser, "Select a user", "user", nil)
block := slack.NewInputBlock("user", label, blockElement)
```
but getting error "invalid_arguments", message "Error opening view".

### Expected behavior
Expecting to be able to produce an input block containing a user select dropdown list.

### Versions
- Go: 1.17.1
- slack-go/slack: v0.6.6

kanata2:
Could you try with the latest version(= v0.9.5)?
If you encounter the same problem, it would be helpful if you could share all the reproducible code with us.

karl-johan-grahn:
I tried `v0.9.5` and same problem.

You can reproduce the problem by updating the example https://github.com/slack-go/slack/blob/master/examples/modal_users/users.go - replace `NewSectionBlock` with the lines above.

In https://github.com/slack-go/slack/blob/master/block_element.go it says `NewOptionsSelectBlockElement returns a new instance of SelectBlockElement for use with the Options object only` - what does the `Options object` refer to, what kind of block is that?

karl-johan-grahn:
Is a `NewUsersSelectBlockElement` missing?

karl-johan-grahn:
Some progress, after reading this blog https://sho-memo01.blogspot.com/2020/09/slack-go.html I could see the `NewOptionsSelectBlockElement` should be specified like:
```
blockElement := slack.NewOptionsSelectBlockElement(slack.OptTypeUser, "Select a user", "user")
```
[...]

PR #976, opened 2021-09-26
Update users example

karl-johan-grahn:
For https://github.com/slack-go/slack/issues/974


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+### v0.9.6 - 26 Sept 2021
+Update users example with input examples
+
```
kanata2: CHANGELOG is not updated now... 😢

kanata2 (commented):
Thanks!

CHANGELOG.md
karl-johan-grahn: I'll update it with my change at least

kanata2: If we update it halfway, users may be confused, so let us only use the release notes for now.

kanata2: related: #987

kanata2 (approved):
Thanks.

Issue #723, opened 2020-05-09
unify slackevents package Event types with main package


theckman:
Hello,

Because Slack has recommended folks no longer use the RTM API for creating bots, and to instead use their Events API, I'm in the process of rewriting the Go Slack Workspace's bot. In the process I ran in to some peculiarities with this package, and wonder if it makes sense to unify the types and leave it up to the consumer to know what fields should be filled-in and why.

Specifically, it's unusual to me that there's a `slackevents.MessageEvent` type, but no `slackevents.TeamJoinEvent` type. So I end up needing to mix imports to deal with these incoming messages from Slack.

Alternatively, I can raise a PR to duplicate these types in to that package so that they are entirely isolated for consumers.

What are your thoughts?

Cheers!

PR #978, opened 2021-09-29
Add TeamJoinEvent

karlkeefer:
This PR adds an `innerEvent` type that corresponds to the `team_join` event from Slack.
Related to #723


kanata2 (approved):
Thanks!

Issue #980, opened 2021-10-01
Extend GetTeamInfo API to accept team argument


ed-wp:
https://api.slack.com/methods/team.info

```
team | string | to get info on, if omitted, will return information about the current  team. Will only return team that the authenticated token is allowed to see through external shared channels
```

The library omits this by default, but does not provide a way to specify another team.

**Current:**
```
func (api *Client) GetTeamInfo() (*TeamInfo, error) 
```

**Proposed:**
```
func (api *Client) GetTeamInfo(team string) (*TeamInfo, error) 
```

**Expected behaviors:**
```
If `team` == "" omit `team` from url.values
If `team` != "" add `team` to url.values
```

PR #981, opened 2021-10-01
Add optional team parameter to GetTeamInfo function calls

ed-wp:
Solves: https://github.com/slack-go/slack/issues/980

##### Pull Request Guidelines

##### PR preparation

##### API changes

- [no tests, if you're adding to the API include at least a single test of the happy case.

### DIFF
--- a/team.go
+++ b/team.go
@@ -96,14 +96,21 @@ func (api *Client) accessLogsRequest(ctx context.Context, path string, values ur
 }
 
 // GetTeamInfo gets the Team Information of the user
-func (api *Client) GetTeamInfo() (*TeamInfo, error) {
-	return api.GetTeamInfoContext(context.Background())
+// if the team is empty (""), will return information about the current team.
+// https://api.slack.com/methods/team.info
+func (api *Client) GetTeamInfo(team string) (*TeamInfo, error) {
+	return api.GetTeamInfoContext(context.Background(), team)
 }
 
 // GetTeamInfoContext gets the Team Information of the user with a custom context
-func (api *Client) GetTeamInfoContext(ctx context.Context) (*TeamInfo, error) {
+// if the team is empty (""), will return information about the current team.
+// https://api.slack.com/methods/team.info
+func (api *Client) GetTeamInfoContext(ctx context.Context, team string) (*TeamInfo, error) {
 	values := url.Values{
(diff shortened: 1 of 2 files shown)


team.go
```diff
@@ -96,15 +96,22 @@ func (api *Client) accessLogsRequest(ctx context.Context, path string, values ur
 }
 
 // GetTeamInfo gets the Team Information of the user
-func (api *Client) GetTeamInfo() (*TeamInfo, error) {
-	return api.GetTeamInfoContext(context.Background())
+// if the team is empty (""), will return information about the current team.
+// https://api.slack.com/methods/team.info
+func (api *Client) GetTeamInfo(team string) (*TeamInfo, error) {
```
ShamanR, line 101: Backward compatibility issue.

ShamanR (commented):
sould we test  emty team case?

team.go
kanata2: I'm having trouble with only one optional parameter...
I've come up with a few ideas, which one do you think is better?

### other idea 1
How about adding `GetAnotherTeamInfo(team string)` method .

```go
// Get your own team information.
info, err := api.GetTeamInfo()

// Get other team information.
info, err := api.GetOtherTeamInfo(teamID)
```

### other idea 2
Use functional option pattern.

```go
// Get your own team information.
info, err := api.GetTeamInfo()

// Get other team information.
info, err := api.GetTeamInfo(slack.GetTeamInfoOptionTeamID(teamID))
```

ed-wp: I feel like even though this is not backwards compatible, it follows how the slack API actually operates and makes it less confusing. 

From [Slack API Docs](https://api.slack.com/methods/team.info):
```
Team to get info on, if omitted, will return information about the current team. Will only return team that the authenticated token is allowed to see through external shared channels
```
I think that other option 2 is closer to the APIs definition, but not nearly as close as my proposal.

Thoughts?

kanata2: > I think that other option 2 is closer to the APIs definition, but not nearly as close as my proposal.

The parameter `team` is an optional in Slack while it is required in our `GetTeamInfo` API with your proposal, so I think it is not so good design.
The difference in idea 2 is that it keeps `team` parameter optional.

kanata2:
#1118 is merged.

PR #982, opened 2021-10-01
Fix socket mode example

rafaelcpalmeida:
I noticed that there was a just a blank line in the validation wether the app token was empty. This fixes it 😁


kanata2 (approved):
Thanks 😆

PR #986, opened 2021-10-08
add message subType constants from https://api.slack.com/events/message

ShamanR:
Add Message SubTypes from https://api.slack.com/events/message as constants


kanata2 (approved):
Thanks :)

Issue #939, opened 2021-05-19
Return response metadata with chat errors


StupidScience:
### Description

Currently with methods like `SendMessageContext` there is no way to get info about error like it is possible for `OpenViewContext` as example. So it leads that you just get errors like `invalid_blocks` without any additional information while in real response can be something like
```json
{
    "ok": false,
    "error": "invalid_blocks",
    "errors": [
        "failed to match all allowed schemas [json-pointer:/blocks/1/elements/0/text]",
        "invalid additional property: verbatim [json-pointer:/blocks/1/elements/0/text]"
    ],
    "response_metadata": {
        "messages": [
            "[ERROR] failed to match all allowed schemas [json-pointer:/blocks/1/elements/0/text]",
            "[ERROR] invalid additional property: verbatim [json-pointer:/blocks/1/elements/0/text]"
        ]
    }
}
```
I've tested it with invocation like this:

```
s.SendMessageContext(
    context.TODO(),
    channelID,
    slack.MsgOptionText("", false),
    slack.MsgOptionBlocks(
	  slack.NewSectionBlock(
		  slack.NewTextBlockObject(slack.MarkdownType, "test", true, true),
		  nil,
		  nil,
	  ),
  ),
)
```

So it is really annoying to debug this issues in semi-manual way to understand what is going on in real.

May be there is some option to create some kind of error wrapper type like the following to avoid breaking the compatibility:
```
type Error struct {
	Err             string
	Messages []string
}
```
Or return response from specified function so it would be possible to get this information?

### (Optional) Slack's documentation

https://api.slack.com/changelog/2016-09-28-response-metadata-is-on-the-way

StupidScience:
@kanata2 sorry for bothering. what do you think about this?

kanata2:
Sorry for my slow response :weary:
I'll confirm later.

abustany:
one could imagine a `SlackErrorResponse`, which would basically be a `SlackResponse` with `Ok == false`. That struct would implement the `error` interface returning its `Error` field as the current code does, but could also be casted using `errors.As` to get access to `ResponseMetadata`. Something like

```go
// SlackErrorResponse is a SlackResponse with an error set
type SlackErrorResponse struct {
  Error string
  ResponseMetadata ResponseMetadata
}

func (r SlackErrorResponse) Error() string { return r.Error } // keeps the same behavior as before

// we'd change this method
func (t SlackResponse) Err() {
  if t.Ok { return nil } // unchanged
[...]

PR #990, opened 2021-10-29
Allow wrapping of error metadata.

abourget:
This PR enables wrapping (`errors.As`) of slack responses which contain rich error messages, e.g., indicating which blocks of position in the submitted JSON the error was found.

Solves #939

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -41,8 +41,16 @@ func (t SlackResponse) Err() error {
 		return nil
 	}
 
-	return errors.New(t.Error)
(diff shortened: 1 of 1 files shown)


abourget:
Then users can do something like:
```golang
func enrichError(err error) error {
	var fullErr slack.SlackErrorResponse
	if errors.As(err, &fullErr) {
		err = fmt.Errorf("%s: %s", err.Error(), strings.Join(fullErr.ResponseMetadata.Messages, ", "))
	}
	return err
}
```
and simply decorate their `err`.

kanata2 (approved):
LGTM. Thanks :)

Issue #992, opened 2021-11-03
Setting user profile picture


jasonbronson:
### What happened
When trying to set a profile pic the api returns with invalid_auth. I thought it might of been related to the api token but attempting to use it in postman works fine. 

See postman working here https://jmp.sh/IrLuiCH

### Expected behavior
To change the profile picture

### Steps to reproduce

```
api := slack.New(token, slack.OptionDebug(true))
// pp := slack.GetUserProfileParameters{
	// 	UserID: "UUL4NKM",
	// }
	// d, err := api.GetUserProfile(&pp)
	// if err != nil {
	// 	log.Println(err)
	// }
	// log.Println(d)

	userSetPhotoParams := slack.UserSetPhotoParams{}
	filename := getPics(path)
	err := api.SetUserPhoto(filename, userSetPhotoParams)
	if err != nil {
		log.Print(err)
	} else {
		fmt.Println("Changed pic")
	}
```

API response was
```
[...]

PR #994, opened 2021-11-05
remove token from url parameters

jasonbronson:
This PR resolves the url which is sent in as parameters for several slack calls. This is a deprecated method as of 2021 from Slack documentation and will result in a failed attempt to call there API. Even if the API is sent in URL params and the token is sent in headers the reply will still fail.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -9,7 +9,7 @@ import (
 
 func (api *Client) adminRequest(ctx context.Context, method string, teamName string, values url.Values) error {
 	resp := &SlackResponse{}
-	err := parseAdminResponse(ctx, api.httpclient, method, teamName, values, resp, api)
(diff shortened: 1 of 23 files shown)


kanata2:
This PR doesn't seem to resolve #992.
It would also be helpful if you could write a description of the PR.

jasonbronson:
/Users/jason/Downloads/profileapp/pics/gopher40.png
2021/11/06 10:45:09 original library
2021/11/06 10:45:09 map[crop_w:[0] crop_x:[0] crop_y:[0]]
Changed pic

It does resolve it.

kanata2:
Umm...this PR doesn't resolve original issue as I mentioned above.(I tried again ,but got same result)
However in the master branch, the original issue should be fixed now.

PR #999, opened 2021-11-10
link_shared `message_ts` field isn't always a number

### LINKED ISSUE #998: The value contained `message_ts` in LinkShared event is not only a timestamp
we're seeing a lot of failures to unmarshal events from slack to their respective event structs, so I went looking to see if there was something amiss.  I found a number of places in `inner_events.go` where `event_ts` is defined as a `json.Number`:

```
	EventTimeStamp json.Number `json:"event_ts"`
```

but i can't find any mention  of `event_ts` in the spec (https://github.com/slackapi/slack-api-specs/blob/760ee9843ec8045ce3f62edd451751b0353acead/events-api/slack_events_api_async_v1.json) where the type is anything other than `string`.

I haven't tracked the errors down to event_ts specifically, but the failures are universally attempts to unmarshal uuids string as `json.Number`, which fails.

toshok:
This is a one-liner, but a breaking change.

As documented in https://api.slack.com/events/link_shared, the `message_ts` field will not be a number if the event is generated from an unfurl in the compose message area (the feature was announced here: https://api.slack.com/changelog/2021-08-changes-to-unfurls).

Includes a new test, which is the example payload from the `link_shared` url above.

Fixes #998

### DIFF
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -161,10 +161,13 @@ type GridMigrationStartedEvent struct {
 
 // LinkSharedEvent A message was posted containing one or more links relevant to your application
 type LinkSharedEvent struct {
-	Type             string        `json:"type"`
-	User             string        `json:"user"`
-	TimeStamp        string        `json:"ts"`
-	Channel          string        `json:"channel"`
+	Type      string `json:"type"`
+	User      string `json:"user"`
+	TimeStamp string `json:"ts"`
+	Channel   string `json:"channel"`
+	// MessageTimeStamp can be both a numeric timestamp if the LinkSharedEvent corresponds to a sent
+	// message and (contrary to the field name) a uuid if the LinkSharedEvent is generated in the
+	// compose text area.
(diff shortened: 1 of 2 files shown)


kanata2:
Thanks! I'll release as soon as possible after merging this.

kanata2 (approved):

slackevents/inner_events.go
```diff
@@ -165,7 +165,7 @@ type LinkSharedEvent struct {
 	User             string        `json:"user"`
 	TimeStamp        string        `json:"ts"`
 	Channel          string        `json:"channel"`
-	MessageTimeStamp json.Number   `json:"message_ts"`
+	MessageTimeStamp string        `json:"message_ts"`
```
kanata2, line 171: Could you please add the reason for using `string` type(= `message_ts` sends more than just timestamps, contrary to the variable name)?

kanata2 (approved):

slackevents/inner_events.go
toshok: absolutely

PR #1012, opened 2021-12-22
new profile update method

PavelVershinin:
Added a method for updating the user profile

### DIFF
--- a/profile.go
+++ b/profile.go
@@ -1,0 +1,38 @@
+package slack
+
+import (
+	"context"
+	"encoding/json"
+	"net/url"
+)
+
+// SetUserProfile will set a profile for the provided user
+func (api *Client) SetUserProfile(user string, profile *UserProfile) error {
+	return api.SetUserProfileContext(context.Background(), user, profile)
+}
+
+// SetUserProfileContext will set a profile for the provided user with a custom context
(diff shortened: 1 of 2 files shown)


kanata2:
Please add a description for this PR.

PavelVershinin:
> Please add a description for this PR.

Done

profile.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// SetUserProfile will set a profile for the provided user
+func (api *Client) SetUserProfile(user string, profile *UserProfile) error {
+	return api.SetUserProfileContext(context.Background(), user, profile)
+}
+
+// SetUserProfileContext will set a profile for the provided user with a custom context
+func (api *Client) SetUserProfileContext(ctx context.Context, user string, profile *UserProfile) error {
+	jsonProfile, err := json.Marshal(profile)
```
kanata2, line 16: Each field in UserProfile is not omitempty, so when it is converted to json, it will all have initial values. So it may occur unintended updates.

kanata2 (changes requested):

profile.go
batazor: How do you think we can correct this behavior?

PR #1020, opened 2022-01-19
add latest_reply property to Msg struct

ameliagapin:
When retrieving a message using [`conversations.History`](https://api.slack.com/methods/conversations.history/test), the Slack API returns a field called `latest_reply` on each message in a thread with the timestamp value in it. This property is not part of the `Msg` struct so I am adding it with this PR.


graphtheoryiscool:
is there an update on this? It would be useful for a project I'm working on 🥺

kanata2 (approved):
Thanks!

PR #1029, opened 2022-02-09
Add support for admin.emoji.list

ryho:
This PR adds functions to call admin.emoji.list endpoint and it adds a test for it. This code is modeled after the code for the emoji.list (non-admin) endpoint.
I have run `make pr-prep`, and it changed the files `webhooks_go112.go` and `webhooks_go113.go` which I have opted not include.

### DIFF
--- a/admin_emoji.go
+++ b/admin_emoji.go
@@ -1,0 +1,41 @@
+package slack
+
+import (
+	"context"
+	"net/url"
+)
+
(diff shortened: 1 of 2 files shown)


kanata2 (changes requested):
Thanks.
Please add support for `cursor` and `limit` arguments 🙏

PR #1032, opened 2022-02-13
all: add new //go:build lines

zchee:
Add new `//go:build` lines.

```console
$ go version
go version go1.17.7 darwin/amd64

$ make fmt
# have diffs
```


kanata2 (approved):
Thanks.
(Actually, we no longer support v1.12, so we should delete webhooks_go112.go 😆 )

PR #1033, opened 2022-02-13
all: remove github.com/pkg/errors dependency

zchee:
Remove `github.com/pkg/errors` dependency.

The https://github.com/pkg/errors package has been deprecated (archived). Use `fmt.Errorf("...:  %w", err)` instead of.

Note that,
- `webhooks_go112.go` for Go version 1.12. This means can't use the `%w` verb for wrapped error
	- Just use `%v` verb
- `fmt.Errorf` haven't `Cause` interface and stacktrace. If you really want to them, we should use `golang.org/x/xerrors` but which is another dependency too
	- IMO, no need for backward compatibility to the `Cause` interface because `slack-go/slack` didn't use `github.com/pkg/errors` much


kanata2 (approved):
LGTM. Thanks!

kanata2 (changes requested):
Ah, sorry.
This package currently activates vendoring.
So please run `go mod vendor` to remove pkg/errors.

zchee:
@kanata2 good catch, PTAL.

kanata2 (approved):
👍

PR #1035, opened 2022-02-14
Optimize NewSafeID using atomic

zchee:
Optimize NewSafeID using atomic instead of a mutex lock.

- [x] is it performance related

---

```
name                  old time/op    new time/op    delta
NewSafeID-20            14.0ns ± 1%     5.9ns ± 1%  -57.76%  (p=0.008 n=5+5)
NewSafeIDParallel-20    23.8ns ± 1%    21.0ns ± 2%  -12.08%  (p=0.008 n=5+5)

name                  old alloc/op   new alloc/op   delta
NewSafeID-20             0.00B          0.00B          ~     (all equal)
NewSafeIDParallel-20     8.00B ± 0%     8.00B ± 0%     ~     (all equal)

name                  old allocs/op  new allocs/op  delta
NewSafeID-20              0.00           0.00          ~     (all equal)
NewSafeIDParallel-20      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```

### DIFF
--- a/messageID.go
+++ b/messageID.go
@@ -1,6 +1,6 @@
 package slack
 
-import "sync"
+import "sync/atomic"
 
 // IDGenerator provides an interface for generating integer ID values.
 type IDGenerator interface {
@@ -11,20 +11,17 @@ type IDGenerator interface {
 // concurrent use by multiple goroutines.
 func NewSafeID(startID int) IDGenerator {
 	return &safeID{
-		nextID: startID,
-		mutex:  &sync.Mutex{},
+		nextID: int64(startID),
 	}
 }
 
 type safeID struct {
-	nextID int
-	mutex  *sync.Mutex
+	nextID int64
 }
 
-func (s *safeID) Next() int {
-	s.mutex.Lock()
-	defer s.mutex.Unlock()
-	id := s.nextID
-	s.nextID++
+func (s *safeID) Next() (id int) {
+	id = int(atomic.LoadInt64(&s.nextID))
+	atomic.AddInt64(&s.nextID, 1)
+
 	return id
 }
--- a/messageID.go
+++ b/messageID.go
@@ -19,9 +19,8 @@ type safeID struct {
 	nextID int64
 }
 
-func (s *safeID) Next() (id int) {
-	id = int(atomic.LoadInt64(&s.nextID))
-	atomic.AddInt64(&s.nextID, 1)
+func (s *safeID) Next() int {
+	id := atomic.AddInt64(&s.nextID, 1)
 
-	return id
+	return int(id)
 }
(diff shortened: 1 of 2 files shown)


zchee:
@kanata2 ping

messageID.go
```diff
@@ -11,20 +11,17 @@ type IDGenerator interface {
...
 }
 
-func (s *safeID) Next() int {
-	s.mutex.Lock()
-	defer s.mutex.Unlock()
-	id := s.nextID
-	s.nextID++
+func (s *safeID) Next() (id int) {
+	id = int(atomic.LoadInt64(&s.nextID))
+	atomic.AddInt64(&s.nextID, 1)
```
brainexe: for me it looks like a race condition here. Multiple goroutines can load the int...this call is atomic, sure, but could return the same integer if it happens before the first goroutine reached the "AddInt64".

As a prove you can add a test calling the Next() concurrent and process the result via `sort | uniq -c` -> I get some rare (but unwanted) duplicates. ~100 cases in ~1mio Next() calls (inside RunParallel)

brainexe: this should at least be safe and faster, as it's only one operation:  
```suggestion
func (s *safeID) Next() int {
        id := atomic.AddInt64(&s.nextID, 1)

        return int(id)
}
```

zchee: @brainexe will check. thanks for advice.

zchee: good catch. will fix.

zchee:
@brainexe cc: @kanata2 Fixed. PTAL.

Note that @brainexe suggestions will increase more performance. Thanks!

Below is before and after the [`daa4329` (#1035)](https://github.com/slack-go/slack/pull/1035/commits/daa43297127ca2b361edd79e75f33031b53088d0) commit.
This pull request's actual optimized benchmark result is I'd edited pull request comment.

```
name                  old time/op    new time/op    delta
NewSafeID-20            7.60ns ± 1%    5.93ns ± 1%  -21.97%  (p=0.008 n=5+5)
NewSafeIDParallel-20    21.0ns ± 1%    21.0ns ± 2%     ~     (p=0.952 n=5+5)

name                  old alloc/op   new alloc/op   delta
NewSafeID-20             0.00B          0.00B          ~     (all equal)
NewSafeIDParallel-20     8.00B ± 0%     8.00B ± 0%     ~     (all equal)

name                  old allocs/op  new allocs/op  delta
NewSafeID-20              0.00           0.00          ~     (all equal)
NewSafeIDParallel-20      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```

kanata2 (approved):
LGTM.

@brainexe Could you take another look?

brainexe:
for me it looks good now :+1:

brainexe (approved):

zchee:
@brainexe Thanks for polite review :D

@kanata2 Can I merge this pull request?

kanata2:
Yeah, go ahead!

PR #1037, opened 2022-02-22
Implement auth.teams.list api method

DavyJohnes:
##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/team.go
+++ b/team.go
@@ -24,6 +24,16 @@ type TeamInfo struct {
 	Icon        map[string]interface{} `json:"icon"`
 }
 
+type TeamInfoShort struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+type AuthTeamsResponse struct {
+	Teams []TeamInfoShort `json:"teams"`
+	SlackResponse
+}
+
 type LoginResponse struct {
 	Logins []Login `json:"logins"`
 	Paging `json:"paging"`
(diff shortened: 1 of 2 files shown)


zchee:
@DavyJohnes Sorry for late, seem to LGTM. will review soon.

users.go
```diff
@@ -292,6 +292,12 @@ func GetUsersOptionPresence(n bool) GetUsersOption {
 	}
 }
 
+func GetUsersOptionTeamID(t string) GetUsersOption {
+	return func(p *UserPagination) {
+		p.teamID = t
+	}
+}
+
```
kanata2, line 300: unused?

team.go
```diff
@@ -165,3 +185,21 @@ func (api *Client) GetBillableInfoForTeamContext(ctx context.Context) (map[strin
 
 	return api.billableInfoRequest(ctx, "team.billableInfo", values)
 }
+
+func (api *Client) GetAuthTeams() ([]TeamInfoShort, error) {
+	return api.GetAuthTeamsContext(context.Background())
+}
```
kanata2, line 191: `ListAuthTeams` looks better 😄

team.go
```diff
@@ -95,6 +105,16 @@ func (api *Client) accessLogsRequest(ctx context.Context, path string, values ur
 	return response, response.Err()
 }
 
+func (api *Client) authTeamsRequest(ctx context.Context, path string, values url.Values) (*AuthTeamsResponse, error) {
```
kanata2, line 108: This method can be added when `auth.teams` specific processing is needed.

team.go
```diff
@@ -165,3 +185,21 @@ func (api *Client) GetBillableInfoForTeamContext(ctx context.Context) (map[strin
 
 	return api.billableInfoRequest(ctx, "team.billableInfo", values)
 }
+
+func (api *Client) GetAuthTeams() ([]TeamInfoShort, error) {
+	return api.GetAuthTeamsContext(context.Background())
+}
+
+func (api *Client) GetAuthTeamsContext(ctx context.Context) ([]TeamInfoShort, error) {
```
kanata2, line 193: Could you support `include_icon` param?

https://api.slack.com/methods/auth.teams.list

kanata2 (changes requested):
Thanks!
I've commented. Please confirm 🙏

PR #1041, opened 2022-03-09
optimize slackutilsx.EscapeMessage function

brainexe:
This change is it performance related....but quite small TBH. Just found if while checking some code. 
The underlaying string.Replacer is optimized to buildup the lookup tables only once (but only if really needed!). 

```
➜  slackutilsx $ go test -bench=. --benchmem               
goos: linux
goarch: amd64
pkg: github.com/slack-go/slack/slackutilsx
cpu: Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz
BenchmarkEscapeMessage-8      	  991929	      1230 ns/op	    6792 B/op	       9 allocs/op
BenchmarkEscapeMessageNew-8   	22310755	        58.94 ns/op	      32 B/op	       2 allocs/op
```
-> ~20 times faster, ~200 times less allocated bytes

### DIFF
--- a/slackutilsx/slackutilsx.go
+++ b/slackutilsx/slackutilsx.go
@@ -50,10 +50,12 @@ func DetectChannelType(channelID string) ChannelType {
 	}
 }
 
+// initialize replacer only once (if needed)
+var escapeReplacer = strings.NewReplacer("&", "&amp;", "<", "&lt;", ">", "&gt;")
+
 // EscapeMessage text
 func EscapeMessage(message string) string {
-	replacer := strings.NewReplacer("&", "&amp;", "<", "&lt;", ">", "&gt;")
-	return replacer.Replace(message)
+	return escapeReplacer.Replace(message)
 }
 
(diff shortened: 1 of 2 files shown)


zchee:
@brainexe Thanks contribute and your change very optimized text replacer :D

Actually, I had to compare using `benchstat`, below is the result:
```
name              old time/op    new time/op    delta
EscapeMessage-20    1.54µs ± 1%    0.05µs ± 4%  -96.70%  (p=0.008 n=5+5)

name              old alloc/op   new alloc/op   delta
EscapeMessage-20    6.79kB ± 0%    0.03kB ± 0%  -99.53%  (p=0.008 n=5+5)

name              old allocs/op  new allocs/op  delta
EscapeMessage-20      9.00 ± 0%      2.00 ± 0%  -77.78%  (p=0.008 n=5+5)
```

That's would be great. I'll merge soon.

zchee (approved):

PR #1046, opened 2022-03-22
Return ScheduledMessageID everywhere

### LINKED ISSUE #757: missing scheduled_message_id
Hello,
based on https://api.slack.com/messaging/scheduling, I should receive a `scheduled_message_id` when using
`POST https://slack.com/api/chat.scheduleMessage` this is in the API:
`ScheduleMessage(channelID, postAt string, options ...MsgOption) (string, string, error)`

I would like to use this to delete my message using the API
`DeleteScheduledMessage(params *DeleteScheduledMessageParameters) (bool, error)`

A workaround is to list the ScheduleMessage, but it would be nicer to get it in the return of `ScheduleMessage()`.

My piece of code
```golang
channelID := "xxx"
token:="xxx"

api := slack.New(token)
now := time.Now().UnixNano() / 1000000000
timestampPlanned := now + 10 //in 10 seconds
fmt.Printf("Planning message at %d ; now=%d\n", timestampPlanned, now)
channelID, timestampMess, err := api.ScheduleMessage(channelID, fmt.Sprintf("%d", timestampPlanned), slack.MsgOptionAttachments(slack.Attachment{Pretext: "test in future"}))
if err != nil {
	panic("Could not send message: " + err.Error())
}
fmt.Printf("Message with Schedule sucessfully sent to channel %s at %s\n", channelID, timestampMess)

_, err = api.DeleteScheduledMessage(&slack.DeleteScheduledMessageParameters{Channel: channelID, ScheduledMessageID: "missing scheduled_message_id"})
if err != nil {
	panic("Could not send message: " + err.Error())
}
fmt.Printf("Message with Schedule sucessfully deleted \n")
```

mfine:

##### API changes

This extends `SendMessage*` to return `scheduled_message_id` everywhere. It would be great if the impact of this change could be smaller, but the common `SendMessage*` functionality is used everywhere so needed to be extended everywhere it was used. If there's a better way to expose `scheduled_message_id` in a more limited fashion, that would be great.

Fixes #757

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -91,7 +91,7 @@ func (api *Client) DeleteMessage(channel, messageTimestamp string) (string, stri
 
 // DeleteMessageContext deletes a message in a channel with a custom context
 func (api *Client) DeleteMessageContext(ctx context.Context, channel, messageTimestamp string) (string, string, error) {
-	respChannel, respTimestamp, _, err := api.SendMessageContext(
+	respChannel, respTimestamp, _, _, err := api.SendMessageContext(
 		ctx,
 		channel,
 		MsgOptionDelete(messageTimestamp),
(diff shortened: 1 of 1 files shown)


kanata2 (commented):
More than 4 or 5 return values for a public function/method is too many,
 so we would like to consider some other approach.

kanata2:
@zchee What do you think?

pankona:
@kanata2 @zchee 
Hello, I'm a one of waiting for release of this patch. I'm afraid to say that I'm happy if you proceed this pull request. Thanks!

hussachai:
Isn't scheduled message ID applicable to SendScheduledMessage API only? And messageTs is applicable to only SendMessage API. IMO, this change makes it more confusing, and I personally don't like multi-values-return that returns more than 2 values because it's hard to know which one is what without a name. I'd replicate `SendMessageContext` for scheduled API maybe name it `ScheduleMessageContext`  and returns scheduled message ID in the position of message TS. The better approach would be refactoring the API to reduce the duplication in code which is not much to reduce anyway.  Just my 2 cents from the user who is waiting for this. 
PS. I have my own a fork version which supports this. My function returns only 2 values instead of 3 like SendMessageContext because I don't know why we need to get a channel ID back when the channel ID is a already known since it is one of function arguments. I don't mind keeping it consistent, but either cases, this will be a breaking change unless we have a new function (the hard part will be naming it :)) and deprecate the old one.

hussachai:
I proposed the fix for this in the least intrusive way here https://github.com/slack-go/slack/pull/1153

Issue #1038, opened 2022-02-23
Question: Enterprise Grid supports


lycano:
Hello slack-go developers,

i would like to ask the current scope of this project before deciding to create a pull request. Im currently working on extending and implementing official admin Web-API functionality so that these can be used via slack bot. In my use-case i need a way to manage SSO exclusion list on enterprise grid. For that i would like to use slack-go/slack.

For now im working on specific methods like admin.auth.policy.* but would also like to implement other admin functions. I see that there is an admin.go file currently present but it uses unofficial API which let me believe this project may not be scoped to enterprise usage?

I would love to contribute.

For a PoC i just added adminEE.go and used existing schema to implement admin functionality and would like to separate each sub section of admin API further like it is done with block.

Example adminEE.go (new files - please ignore method names for now this is just an example)
```
package slack

import (
	"context"
	"fmt"
	"net/url"
)

type adminEEResponseFull struct {
	AdminEETeams       []AdminEETeam        `json:"teams,omitempty"`
	AuthPolicyEntities []AuthPolicyEntities `json:"entities,omitempty"`
	SlackResponse
}

func (api *Client) adminEERequest(ctx context.Context, path string, values url.Values) (*adminEEResponseFull, error) {
	response := &adminEEResponseFull{}
	err := api.postMethod(ctx, fmt.Sprintf("admin.%s", path), values, response)
	if err != nil {
		return nil, err
	}

	if err := response.Err(); err != nil {
		return nil, err
	}

	return response, nil
}
```

adminEE_teams.go
```
package slack

import (
	"context"
	"net/url"
	"strconv"
)

type AdminEETeamOwner struct {
	UserId string `json:"id"`
	Email  string `json:"email"`
}

type AdminEETeam struct {
	Id              string           `json:"id"`
	Name            string           `json:"name"`
	Discoverability string           `json:"discoverability"`
	PrimaryOwner    AdminEETeamOwner `json:"primary_owner"`
	TeamUrl         string           `json:"team_url"`
}

// AdminTeamList List all teams on an Enterprise organization.
func (api *Client) AdminTeamList() (*[]AdminEETeam, error) {
	return api.AdminTeamListContext(context.Background(), "", 100)
}

// AdminTeamListContext List all teams on an Enterprise organization with a custom context
func (api *Client) AdminTeamListContext(ctx context.Context, cursor string, limit int) (*[]AdminEETeam, error) {
	values := url.Values{
		"token": {api.token},
[...]

PR #1047, opened 2022-03-22
Added Details field to Audit event entries 

JustinJudd:
Added the optional details field for the audit events. As there aren't details what all of the sub-fields are, adding different fields that we've seen in our logs.

### DIFF
--- a/audit.go
+++ b/audit.go
@@ -40,19 +40,19 @@ type AuditEntry struct {
 		IPAddress string `json:"ip_address"`
 	} `json:"context"`
 	Details struct {
-		NewValue interface{} `json:"new_value"`
-		PrevValue interface{} `json:"previous_value"`
-		PolicyName string `json:"policy_name"`
-		AuthenticationMode string `json:"authentication_mode"`
-		EntityIds string `json:"entity_ids"`
-		MobileOnly bool `json:"mobile_only"`
-		WebOnly bool `json:"web_only"`
-		NonSSOOnly bool `json:"non_sso_only"`
-		SucceededUsers string `json:"succeeded_users"`
-		FailedUsers string `json:"failed_users"`
-		ExportType string `json:"export_type"`
-		ExportStart string `json:"export_start_ts"`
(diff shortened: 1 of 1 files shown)


zchee:
@JustinJudd Could you run `make fmt`?

JustinJudd:
Sorry about that. Formatted.

audit.go
```diff
@@ -39,6 +39,21 @@ type AuditEntry struct {
 		UA        string `json:"ua"`
 		IPAddress string `json:"ip_address"`
 	} `json:"context"`
+	Details struct {
```
kanata2, line 42: [NOTE]
The official Java SDK's implementations are here.

https://github.com/slackapi/java-slack-sdk/blob/v1.21.2/slack-api-client/src/main/java/com/slack/api/audit/response/LogsResponse.java#L165

audit.go
```diff
@@ -39,6 +39,21 @@ type AuditEntry struct {
...
+		NewValue           interface{} `json:"new_value"`
+		PrevValue          interface{} `json:"previous_value"`
```
kanata2:
```suggestion
		PreviousValue          interface{} `json:"previous_value"`
```

kanata2 (commented):
I couldn't confirm some fields in Java SDK's implementation.

- policy_name
- authentication_mode
- entity_ids
- succeeded_users
- failed_users

As I mentioned in #1038, we don't have a way to verify Enterprise Grid API's behavior.
If these fields are not essential to your use case, we can merge them as soon as you remove them. How about it?

audit.go
JustinJudd: Thanks for the review!

Those are actual fields we have seen, but we can get by without them for now. I've removed them for parity with the Java and Python SDKs.

kanata2: Thanks!

> Those are actual fields we have seen, but we can get by without them for now

Umm, maybe there has been a change at some point... 🤔

kanata2 (approved):
👍

Issue #1051, opened 2022-04-08
Add event_ts field on all "inner events"


abustany:
### Description
Currently (as of 93035c946040f00b2eb69dc9a954f83a9803e931) the `EventTimestamp` field is only present on some inner events types. According to [the OpenAPI specs](https://github.com/slackapi/slack-api-specs/blob/master/events-api/slack_events_api_async_v1.json#L30), the `event_ts` field should be present along with `type` on all inner events.

### (Optional) Slack's documentation
https://github.com/slackapi/slack-api-specs/blob/master/events-api/slack_events_api_async_v1.json#L30

The web docs don't mention that `event_ts` should be present, but I can verify that it actually is.

kanata2:
The slack-api-specs repository is not maintained precisely as far as I can see.
So if you want to know the latest specifications, you can look at the official SDKs for Node.js and Java.

abustany:
ah good thanks for the pointer to the Java SDK. I updated https://github.com/slack-go/slack/pull/1052 to not add `event_ts` on the grid migration events, since those don't seem to have that field according to the Java SDK.

I wonder if one could hack a quick&dirty parser to generate the Go models from the Java ones :shrug:

PR #1052, opened 2022-04-08
Uniformize event_ts support on inner events

abustany:
This PR adds uniform support for the `event_ts` field on all inner events, fixing #1051 . It includes two breaking changes, but I consider those optional and would be happy with only merging the first commit.

The two breaking changes are the following
1. Rename `WorkflowStepExecuteEvent.EventTS` to `EventTimestamp`. I did a "brutal" rename, but if backwards compatibility is an issue we could also investigate keeping both fields and have custom `MarshalJSON`/`UnmarshalJSON` functions, though this also opens weird questions like "what do we do when serializing if both are set?".
2. Change all `EventTimestamp` fields to be strings, and not `json.Number`. We're not worried about losing precision here since `json.Number` is actually a string, but it seems (cf. #998) that this field had to be switched to a string for couple of events already, and having an inconsistent type here feels a bit weird.

### DIFF
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -3,8 +3,6 @@
 package slackevents
 
 import (
-	"encoding/json"
-
 	"github.com/slack-go/slack"
 )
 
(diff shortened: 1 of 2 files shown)


kanata2 (approved):
LGTM. Thanks!

PR #1054, opened 2022-04-12
`WithStyle` for `ConfirmationBlockObject`s should be fluent

elldritch:
This PR fixes `WithStyle` for `ConfirmationBlockObject`s to be fluent, much like the `WithStyle` for `ButtonBlockElement`s.


kanata2 (approved):
👍

PR #1055, opened 2022-04-12
Add fluent `WithConfirm` for `ButtonBlockElement`s

elldritch:
This PR adds a `WithConfirm` fluent method helper that behaves much the same way as `WithStyle` does. This allows buttons to be fully constructed as expressions more often.


kanata2 (commented):
LGTM. Thanks!

kanata2 (approved):

PR #1059, opened 2022-04-22
Goodbye vendor dir

kanata2:
Remove vendor directory


obalunenko:
Prefer to keep venor directory committed to have reproducible builds and be save from disappearing of libraries from GitHub.
https://blog.boot.dev/golang/should-you-commit-the-vendor-folder-in-go/

obalunenko (changes requested):
Prefer to keep vednor directory committed to have reproducible builds and be save from disappearing of libraries from GitHub.
https://blog.boot.dev/golang/should-you-commit-the-vendor-folder-in-go/

kanata2:
My understanding is that Go modules should remain cached in the module proxy even if a repository is deleted.

obalunenko:
@kanata2 Unfortunately proxy servers that're not self hosted cannot provide any guarantee that they will not lose module.
Anyway this topic was always controversial. Just interesting article to read https://www.ardanlabs.com/blog/2020/04/modules-06-vendoring.html

zchee (approved):
I think self hosted proxy server haven't good SLO than hosted by Google Go team. so LGTM.

PR #1060, opened 2022-05-06
add: workflows.stepCompleted and workflwows.stepFailed support

nakatanakatana:
add support for the [workflows.stepCompleted](https://api.slack.com/methods/workflows.stepCompleted) and [workflows.stepFailed](https://api.slack.com/methods/workflows.stepFailed) 

using this API to indicate Slack that the workflow step is complete

### DIFF
--- a/workflow_step_execute.go
+++ b/workflow_step_execute.go
@@ -19,14 +19,25 @@ type (
 	}
 )
 
+type WorkflowStepCompletedRequestOption func(opt WorkflowStepCompletedRequest) error
+
+func WorkflowStepCompletedRequestOptionOutput(outputs map[string]string) WorkflowStepCompletedRequestOption {
+	return func(opt WorkflowStepCompletedRequest) error {
+		if len(outputs) > 0 {
+			opt.Outputs = &outputs
+		}
+		return nil
+	}
+}
+
 // WorkflowStepCompleted indicates step is completed
-func (api *Client) WorkflowStepCompleted(workflowStepExecuteID string, outputs *map[string]string) error {
+func (api *Client) WorkflowStepCompleted(workflowStepExecuteID string, options ...WorkflowStepCompletedRequestOption) error {
 	// More information: https://api.slack.com/methods/workflows.stepCompleted
 	r := WorkflowStepCompletedRequest{
 		WorkflowStepExecuteID: workflowStepExecuteID,
 	}
-	if outputs != nil {
-		r.Outputs = outputs
+	for _, option := range options {
+		option(r)
 	}
 
 	endpoint := api.endpoint + "workflows.stepCompleted"
--- a/workflow_step_execute.go
+++ b/workflow_step_execute.go
@@ -1,0 +1,74 @@
+package slack
+
+import (
+	"context"
+	"encoding/json"
+)
+
+type (
+	WorkflowStepCompletedRequest struct {
+		WorkflowStepExecuteID string             `json:"workflow_step_execute_id"`
+		Outputs               *map[string]string `json:"outputs"`
+	}
+
+	WorkflowStepFailedRequest struct {
+		WorkflowStepExecuteID string `json:"workflow_step_execute_id"`
+		Error                 struct {
+			Message string `json:"message"`
+		} `json:"error"`
+	}
+)
+
+// WorkflowStepCompleted indicates step is completed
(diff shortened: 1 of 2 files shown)


kanata2:
Please add description for your changes.

workflow_step_execute.go
```diff
@@ -0,0 +1,74 @@
...
+	WorkflowStepFailedRequest struct {
+		WorkflowStepExecuteID string `json:"workflow_step_execute_id"`
+		Error                 struct {
+			Message string `json:"message"`
+		} `json:"error"`
+	}
+)
+
+// WorkflowStepCompleted indicates step is completed
+func (api *Client) WorkflowStepCompleted(workflowStepExecuteID string, outputs *map[string]string) error {
```
kanata2: [question]
What is the reason of using a pointer of map?

kanata2: How about receiving optional arguments like the following way?
https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html

nakatanakatana: > Why use a pointer to a map?

`outputs` parameter is optional argment and it key can be any.

https://api.slack.com/methods/workflows.stepCompleted#arg_outputs

I used pointer to represent empty because `outputs` is optional
Also, I used  map[string]string to use any key/value pairs.


> How about receiving optional arguments like the following way?

Does this mean creating a function to set `outputs` and using to argument of WorkflowStepCompleted function?
like below

```diff
diff --git a/workflow_step_execute.go b/workflow_step_execute.go
index de31b19..d1d77cb 100644
--- a/workflow_step_execute.go
+++ b/workflow_step_execute.go
@@ -19,14 +19,25 @@ type (
 	}
 )
 
+type WorkflowStepCompletedRequestOption func(opt WorkflowStepCompletedRequest) error
+
+func WorkflowStepCompletedRequestOptionOutput(outputs map[string]string) WorkflowStepCompletedRequestOption {
+	return func(opt WorkflowStepCompletedRequest) error {
+		if len(outputs) > 0 {
+			opt.Outputs = &outputs
+		}
+		return nil
+	}
+}
+
 // WorkflowStepCompleted indicates step is completed
-func (api *Client) WorkflowStepCompleted(workflowStepExecuteID string, outputs *map[string]string) error {
+func (api *Client) WorkflowStepCompleted(workflowStepExecuteID string, options ...WorkflowStepCompletedRequestOption) error {
 	// More information: https://api.slack.com/methods/workflows.stepCompleted
 	r := WorkflowStepCompletedRequest{
 		WorkflowStepExecuteID: workflowStepExecuteID,
 	}
-	if outputs != nil {
-		r.Outputs = outputs
+	for _, option := range options {
+		option(r)
 	}
 
 	endpoint := api.endpoint + "workflows.stepCompleted"

```

kanata2: > I used pointer to represent empty because outputs is optional

FYI: Maybe, you can pass nil without pointer.

> Does this mean creating a function to set outputs and using to argument of WorkflowStepCompleted function?

Yes, that's right.

nakatanakatana: Thank you for reviewing.
I pushed new commit.

workflow_step_execute.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"context"
+	"encoding/json"
+)
+
+type (
+	WorkflowStepCompletedRequest struct {
+		WorkflowStepExecuteID string             `json:"workflow_step_execute_id"`
+		Outputs               *map[string]string `json:"outputs"`
```
kanata2: Please use `map[string]string` instead of `*map[string]string`.

Same as https://github.com/slack-go/slack/pull/1060#discussion_r896800825

kanata2 (commented):
Sorry for delay. I commented on one point.

workflow_step_execute.go
nakatanakatana: fixed https://github.com/slack-go/slack/pull/1060/commits/9013791a078cb1d461d50e6646bacfe1978c9c81

kanata2 (commented):
Thanks!

kanata2 (approved):

kanata2:
Released! https://github.com/slack-go/slack/releases/tag/v0.11.2

PR #1064, opened 2022-05-17
Fix field types in `SubteamMembersChangedEvent`

youdofoo:
`AddedUsersCount` and `RemovedUsersCount` fields in `SubteamMembersChangedEvent` struct are defined as a string type, but in fact these fields in received data from [`subteam_members_changed`](https://api.slack.com/events/subteam_members_changed) have a number type.
So when you use `slackevents.ParseEvent` function to this event, following parse error occurs:
```
EventsAPI Error parsing inner event: unmarshalling_error, json: cannot unmarshal number into Go struct field SubteamMembersChangedEvent.added_users_count of type string: wrapError
```

To fix this, I changed the type of `AddedUsersCount` and `RemovedUsersCount` fields from string to int.

By the way, also in [API document](
https://api.slack.com/events/subteam_members_changed), `"added_users_count"` and `"removed_users_count"` are defined as a string.
I already reported this to https://slack.com/help.

### DIFF
--- a/websocket_subteam.go
+++ b/websocket_subteam.go
@@ -14,9 +14,9 @@ type SubteamMembersChangedEvent struct {
 	DatePreviousUpdate JSONTime `json:"date_previous_update"`
 	DateUpdate         JSONTime `json:"date_update"`
 	AddedUsers         []string `json:"added_users"`
-	AddedUsersCount    string   `json:"added_users_count"`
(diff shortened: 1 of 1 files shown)


kanata2 (approved):
LGTM.
This PR breaks backward compatibility, so I'm planning to release in v0.11.0.

PR #1068, opened 2022-05-19
Support "reply_broadcast" parameter in webhook API

jjaxu:
Hello! 

Slack's [Webhook API](https://api.slack.com/messaging/webhooks) actually supports the `"reply_broadcast"` parameter when sending replies, just like its [`chat.postMessage` API](https://api.slack.com/messaging/webhooks) although it's not documented explicitly in its webhook API docs. (I discovered it while working on a Slack bot for work).

The webhook SDK currently doesn't support this (it's only supported in `PostMessageParameters`), so I would like to add this to the `WebhookMessage` struct as well so it can be used for webhooks.

Note:
The default value of this parameter is `false` when not specified, so there won't be any breaking changes.

Hopefully, this can get accepted so I can start using the SDK over raw HTTP requests.
Thanks!


kanata2 (approved):
LGTM!

PR #1069, opened 2022-05-27
Add `Permalink` to Msg struct

jreinstra:
At least in the Slack list-stars method if not others, a permalink is included with each item returned. This permalink is not available via slack-go because it is not a field on the struct used to parse that response, so this change adds that missing field.

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -131,7 +131,6 @@ type Msg struct {
 
 	// Block type Message
 	Blocks Blocks `json:"blocks,omitempty"`
-	
 	// permalink
 	Permalink string `json:"permalink,omitempty"`
 }
--- a/messages.go
+++ b/messages.go
@@ -131,6 +131,9 @@ type Msg struct {
 
 	// Block type Message
 	Blocks Blocks `json:"blocks,omitempty"`
+	
+	// permalink
+	Permalink string `json:"permalink,omitempty"
 }
 
 const (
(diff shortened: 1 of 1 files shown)


messages.go
```diff
@@ -131,6 +131,9 @@ type Msg struct {
 
 	// Block type Message
 	Blocks Blocks `json:"blocks,omitempty"`
+	
+	// permalink
+	Permalink string `json:"permalink,omitempty"
```
kanata2:
```suggestion
	Permalink string `json:"permalink,omitempty"`
```

kanata2 (changes requested):
Thanks for your contribution! I commented, please confirm.

jreinstra:
> Thanks for your contribution! I commented, please confirm.

Oh whoops, accepted fix, thanks!

messages.go
```diff
@@ -131,6 +131,9 @@ type Msg struct {
 
 	// Block type Message
 	Blocks Blocks `json:"blocks,omitempty"`
+
```
kanata2: 😄 

```suggestion

```

kanata2 (changes requested):
I found one minor style-related fix, so please confirm 🙏

jreinstra:
> I found one minor style-related fix, so please confirm 🙏

whoops, accepted, thanks!

kanata2 (approved):
LGTM! Thanks.

PR #1071, opened 2022-06-01
Expose StatusCodeError

### LINKED ISSUE #878: statusCodeError struct should be accessible outside package

photown:
This change addresses issue #878

This allows library clients to perform checks on the HTTP status code.

##### Pull Request Guidelines

##### PR preparation

- [X] exposes a previously private type, const, method, etc.

##### API changes

Resolve: #878

### DIFF
--- a/status_code_error.go
+++ b/status_code_error.go
@@ -1,4 +1,4 @@
-package misc
+package slack
 
 import (
 	"fmt"
--- a/misc.go
+++ b/misc.go
@@ -18,8 +18,6 @@ import (
 	"strconv"
 	"strings"
 	"time"
-
-	"github.com/slack-go/slack/internal/misc"
 )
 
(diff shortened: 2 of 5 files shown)


zchee:
Looks good to me.

@kanata2 WDYT?

status_code_error.go
```diff
@@ -1,4 +1,4 @@
-package misc
+package slack
```
kanata2, line 1: [note]
Maybe it is time to refactor error handling 🤔

kanata2 (approved):
LGTM. Thanks!

PR #1073, opened 2022-06-04
interactions: parse AppInstalledTeamID

7ail:
Slack includes a field called `app_installed_team_id` within the view object when the following interactions are triggered

* `view_submission`
* `view_closed`

This PR aims to expose the data to consumers of this library when unmarshalling into a InteractionCallback struct

This field helps us differentiate between different instances of a Slack App that we are interacting with within a Slack Connect channel.

### DIFF
--- a/interactions_test.go
+++ b/interactions_test.go
@@ -68,7 +68,8 @@ const (
 					"text": "Select a date"
 				  }
 				}
-			}]
+			}],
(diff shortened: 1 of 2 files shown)


yoshdog (approved):

kanata2 (approved):
Good catch! Thanks.

PR #1075, opened 2022-06-05
Add missing panic in socket mode handler example

puskunalis:
Adds a missing panic statement when the app token environment variable is empty inside an empty if block.


kanata2 (approved):
👍

PR #1077, opened 2022-06-08
Added GetUsersOptionTeamID so that a team ID can be passed in for use 

antonylucisano:
- GetUsers and GetUsersContext now takes Options
- Added team ID as an additional option as this is required for Organization level API calls used in GRID


users.go
```diff
@@ -292,6 +292,13 @@ func GetUsersOptionPresence(n bool) GetUsersOption {
 	}
 }
 
+// GetUsersOptionTeamID include team_id
+func GetUsersOptionTeamID(team_id string) GetUsersOption {
```
kanata2: Prefer to use camel case in Go projects.

kanata2 (commented):
Thanks!

users.go
antonylucisano: Cool, have updated to camelCase

kanata2 (approved):

PR #1080, opened 2022-06-25
sharedLinks struct should be exported

### LINKED ISSUE #1079: slackevents.sharedLinks struct is unexported
### What happened

I'm writing a test for my shared link unfurler, and the `LinkSharedEvent` struct has a field `Links` which is an array of `sharedLinks`.

https://github.com/slack-go/slack/blob/9de73cee5f30be540fd051e94c3753c6d89ea905/slackevents/inner_events.go#L185-L188

### Expected behavior

The `sharedLinks` type should be exported.

#### reproducible code

Try to create a `LinkSharedEvent` with a list of `Links`:
```
	e := slackevents.LinkSharedEvent{
		Links:            []slackevents.sharedLinks{},
	}
```

### Versions
- Go: 1.18
- slack-go/slack: v0.10.2

pmenglund:
Fixes #1079 - breaks API compatibility as a struct is exported, but I assume it was not intended to be unexported.

### DIFF
--- a/slackevents/inner_events_test.go
+++ b/slackevents/inner_events_test.go
@@ -1,7 +1,8 @@
-package slackevents
+package slackevents_test
 
 import (
 	"encoding/json"
+	"github.com/slack-go/slack/slackevents"
 	"testing"
 )
 
@@ -20,7 +21,7 @@ func TestAppMention(t *testing.T) {
 				"blah": "test"
 		}
 	`)
-	err := json.Unmarshal(rawE, &AppMentionEvent{})
+	err := json.Unmarshal(rawE, &slackevents.AppMentionEvent{})
 	if err != nil {
 		t.Error(err)
 	}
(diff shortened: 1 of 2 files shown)


slackevents/inner_events_test.go
```diff
@@ -1,7 +1,8 @@
-package slackevents
+package slackevents_test
```
kanata2: Why did you change package name?

kanata2 (commented):
Thanks for your PR!

slackevents/inner_events_test.go
pmenglund: To ensure it is tested using the exported API

kanata2: I see.
I think we don't need to change the package name for these tests, so let's leave it as it is 🙏

pmenglund: reverted

kanata2: Thanks.

kanata2 (approved):
LGTM! Thanks.

Actually, I think this changes don't break user's applications by updating slack-go, so maybe okay to release in the next patch version.

PR #1086, opened 2022-07-04
Add option to detect unmapped error

mszostok:
**Description**

This pull request adds:
- dedicated type error for `UnmappedError` 
- unit tests coverage

**Reason**

The library uses the `UnmarshallingErrorEvent` error in two cases:
1. when there was a real `json.Unmarshal`
2. or when there was no corresponding Go struct for unmarshal process

We want to be able to check what kind of error we got. As a result, we will be able to do the assertion whether we care about this unmarshal error or not. Example snippet:

```go
// handlingEvents holds event names for which we have corresponding business logic.
var handlingEvents = sets.NewString("message", "connected")

// ...trimmed...

switch ev := msg.Data.(type) {
case *slack.UnmarshallingErrorEvent:
  err, ok := ev.ErrorObj.(slack.UnmappedError)
  if ok && !handlingEvents.Has(err.EventType()) { // it's not an err for us, as we don't care about this event type.
    b.log.Debugf("Slack unmarshalling error: %+v", ev.Error())
    continue
  }

  b.log.Errorf("Slack unmarshalling error: %+v", ev.Error())
}
```

Currently, we have in logs:
```
ERRO[2022-06-15T11:30:02Z] Slack unmarshalling error: RTM Error: Received unmapped event "user_status_changed"
```

We could do the error string assertion but it's not a good practice, so we want to avoid that.

**Implementation**
I decided to leave the `UnmarshallingErrorEvent` as it is, but instead change 
```go
err := fmt.Errorf("RTM Error: Received unmapped event %q", typeStr)
```
to typed error. As a result, is **not a breaking change.**

### DIFF
--- a/websocket_managed_conn.go
+++ b/websocket_managed_conn.go
@@ -19,33 +19,26 @@ import (
 // UnmappedError represents error occurred when there is no mapping between given event name
 // and corresponding Go struct.
 type UnmappedError struct {
-	eventType string
-	rawEvent  json.RawMessage
-	ctxMsg    string
+	// EventType returns event type name.
+	EventType string
+	// RawEvent returns raw event body.
+	RawEvent json.RawMessage
+
+	ctxMsg string
 }
 
 // NewUnmappedError returns new UnmappedError instance.
 func NewUnmappedError(ctxMsg, eventType string, raw json.RawMessage) *UnmappedError {
 	return &UnmappedError{
 		ctxMsg:    ctxMsg,
-		eventType: eventType,
-		rawEvent:  raw,
+		EventType: eventType,
+		RawEvent:  raw,
 	}
 }
 
 // Error returns human-readable error message.
 func (u UnmappedError) Error() string {
-	return fmt.Sprintf("%s: Received unmapped event %q", u.ctxMsg, u.eventType)
-}
-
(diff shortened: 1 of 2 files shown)


websocket_managed_conn.go
```diff
@@ -16,6 +16,38 @@ import (
...
+
+// EventType returns event type name.
+func (u UnmappedError) EventType() string {
+	return u.eventType
+}
+
+// RawEvent returns raw event body.
+func (u UnmappedError) RawEvent() json.RawMessage {
+	return u.rawEvent
+}
```
kanata2: It seems to me that these fields could be exported. What do you think?

kanata2 (approved):
I commented on one point, but LGTM.

websocket_managed_conn.go
mszostok: Yep, in this case it makes. I automatically went with the method style to have more control in the future. For example:
```go
type Event struct {
	eType *string
}

func (e *Event) EventType() string {
	if e.eType == nil {
		return "unknown"
	}
	return *e.eType
}
```

but it's not the case here, so I changed as suggested 👍

kanata2 (approved):
👍

PR #989, opened 2021-10-21
Add include_num_members support to conversation.info request

Yohan460:
Add [include_num_members](https://api.slack.com/methods/conversations.info#arg_include_num_members) support to [conversation.info](https://api.slack.com/methods/conversations.info) request

PR #1090, opened 2022-07-14
Add include_num_members option to GetConversationInfo

nowakca:
##### API changes

This changes GetConversationInfo to add a second boolean to support the include_num_members field. This could be resolved similarly by creating a new method like GetConversationInfoWithMemberInfo, but otherwise resolving this missing field more generally would require a more structural change (a way of supporting arbitrary fields, which would require a more significant change to the API generally)

Since the existing tests did no specific testing for include_locale, I didn't add one for include_num_members, though I did update the test to compile/run successfully.

### DIFF
--- a/conversation.go
+++ b/conversation.go
@@ -358,16 +358,17 @@ func (api *Client) CreateConversationContext(ctx context.Context, channelName st
 }
 
 // GetConversationInfo retrieves information about a conversation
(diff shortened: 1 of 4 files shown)


kanata2 (commented):
Do yadmin_calls.go and yadmin_generic.go need this PR?

kanata2:
Also, this PR is dup of #989, so close this now.

PR #1091, opened 2022-07-15
add debug logging for rate limited errors

3choBoomer:
In order to have some visibility into continued rate limit errors, adding call for debug logging.

### DIFF
--- a/socketmode/socket_mode_managed_conn.go
+++ b/socketmode/socket_mode_managed_conn.go
@@ -219,6 +219,7 @@ func (smc *Client) connect(ctx context.Context, connectionCount int, additionalP
 				return nil, nil, err
 			}
 		case *slack.RateLimitedError:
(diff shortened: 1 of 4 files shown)


socketmode/socket_mode_managed_conn.go
```diff
@@ -219,6 +219,7 @@ func (smc *Client) connect(ctx context.Context, connectionCount int, additionalP
 				return nil, nil, err
 			}
 		case *slack.RateLimitedError:
+			smc.Debugf("socket mode open and dial: %s", err.(*slack.RateLimitedError).Error())
```
kanata2, line 222: How about using `actual` instead of `err.(*slack.RateLimitedError)`?

```suggestion
			smc.Debugf("socket mode open and dial: %s", actual.Error())
```

kanata2 (commented):
Thanks.
I commented one minor point.

Issue #996, opened 2021-11-08
Add support for `cookie` HTTP header


pbar1:
### Description

Slack has multiple different types of API tokens, each identified by a specific prefix. For example, bot users have a token prefixed with `xoxb`.

Using browser devtools, one can inspect HTTP requests with Slack and see that the UI is communicating with the API using a special token prefixed with `xoxc`. This token functions similarly to a standard Slack token, with the caveat that it only works with the HTTP header `cookie` also set.

Support for setting the `cookie` header would enable using `xoxc` tokens with this library. An example use case for them would be developing locally, as having to make a bot user is extra overhead when trying out new ideas. 

### (Optional) Slack's documentation

This is undocumented behavior. However, here is some unofficial discussion on the subject:
- https://github.com/adsr/irslackd/issues/84
- https://github.com/adsr/irslackd/wiki/IRC-Client-Config#xoxc-tokens

rusq:
@pbar1 you can check [github.com/rusq/slack](https://github.com/rusq/slack) (a fork with cookie support)
```
slack.New(token, slack.OptionCookie(cookie))
```

nolanlum:
@rusq Is there a reason you haven't opened a PR to get cookie support merged into upstream?

rusq:
@nolanlum I'd be happy to, would save me from merging the upstream updates every now and then.  Saw there's already a PR with "thinking" state, and this very issue with no attention, and thought there's no interest in this.  Can submit a PR, no problem.

rusq:
I left the same comment in the PR, but will copy here for convenience:

Whoever's going to be searching for this, the solution is already there, no need to build anything custom:
- I was able to work around that by using a custom `http.Client` with a cookie jar and passing `OptionHTTPClient` to slack.New. [Example](https://github.com/rusq/slackdump/blob/85358171b3fd3c44e78f82e4dbd684d84eb4be74/slackdump.go#L116)

You can grab the chttp implementation from slackdump or BYO.

Thanks, you're welcome. Closing this.

rusq:
Can't self-close the issue.

PR #1092, opened 2022-07-19
Add support for "Client token"/Cookie auth

rusq:
Encouraged by @nolanlum in #996, opening a PR, hoping that this would be useful to community.

##### PR preparation

##### What has changed
This PR adds support for slack client token/cookie (xoxc, xoxd) pair authentication by adding two options that may be passed on instantiation of the Client:

- `OptionCookie(name,value)` - convenience function that allows  adding a "slack.com" cookie that will be added to each request.
- `OptionCookieRAW(cookies ...*http.Cookie)` - gives more flexibility and allows adding any custom cookie.  Main intent of this function is to allow the caller to pass the cookies from Mozilla "cookies.txt".

This change affects `downloadFile`, and `postForm` functions and the calling function.  All others remain untouched (i.e. `doPost` and its callers).  Reason for that - the application that uses this function ([slackdump](https://github.com/rusq/slackdump))  is read-only, and I simply wouldn't be able to test that in practical application, but would be open if someone stands up to help me updating those, if required.

Either way - it might be a good starting point to add the support for this.


rusq:
The Go 1.17 failure does not seem related to the change?

rusq:
Hey @nolanlum is there a reason that this hasn't been given attention yet?

Is there anything needing to be addressed?

nolanlum:
> Hey @nolanlum is there a reason that this hasn't been given attention yet?

I don't actually have write access! So my review wouldn't mean very much. But it does look good to me.

nolanlum (approved):

rusq:
> > Hey @nolanlum is there a reason that this hasn't been given attention yet?
> 
> 
> 
> I don't actually have write access! So my review wouldn't mean very much. But it does look good to me.

Thank you anyway :) now it is not so lonely in this PR!

robrecord:
I an unable to authenticate with any other method for some reason - would be great to see this PR reviewed by someone with write access :-)

rusq:
Happy New Year!

rusq:
Whoever's going to be searching for this, the solution is already there, no need to build anything custom:
- I was able to work around that by using a custom `http.Client` with a cookie jar and passing `OptionHTTPClient` to slack.New. [Example](https://github.com/rusq/slackdump/blob/85358171b3fd3c44e78f82e4dbd684d84eb4be74/slackdump.go#L116)

You can grab the chttp implementation from slackdump or BYO.

Thanks, you're welcome. Closing this.

PR #57, opened 2016-02-06
reaction item only contains IDs

MDrollette:
The `reaction_added` and `reaction_removed` events [do not include](https://api.slack.com/events/reaction_added) the embedded items the same as when retrieving the reaction list.

PR #1095, opened 2022-08-10
Expose reactionItem and reactionEvent

grimmy:
Per the pr guidelines this should apparently been an issue, but seeing as nothing else in the library uses this and it was public until https://github.com/slack-go/slack/pull/57 was merged that we could forgo the issue.

### DIFF
--- a/websocket_reactions.go
+++ b/websocket_reactions.go
@@ -1,7 +1,7 @@
 package slack
 
-// reactionItem is a lighter-weight item than is returned by the reactions list.
-type reactionItem struct {
+// ReactionItem is a lighter-weight item than is returned by the reactions list.
+type ReactionItem struct {
 	Type        string `json:"type"`
 	Channel     string `json:"channel,omitempty"`
 	File        string `json:"file,omitempty"`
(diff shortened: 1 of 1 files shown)


grimmy:
Since the lint is failing on unrelated code, I assume I need to rebase this?

kanata2 (approved):
Sorry for delay. LGTM. Thanks!

kjelle:
Hello. Is there a reason for not expose `slackevents.reactionEvent` aswell?
https://github.com/slack-go/slack/blob/master/slackevents/inner_events.go#L315

kjelle:
Hello. Is there a reason for not to expose `slackevents.reactionEvent` aswell? Almost (didnt check all) all other events in that file are public. Reason for this change is the need to use `slackevents.reactionEvent` outside, e.g. in mock/tests.

https://github.com/slack-go/slack/blob/master/slackevents/inner_events.go#L315

I see there is a PR for this issue before: https://github.com/matthew-graves/slack/commit/a293ed46be13cdf1972630d22de6c44bfd991eb8

PR #1097, opened 2022-08-15
Support team.profile.get method

MattDavisRV:
This PR is a duplicate of #474 that was closed due to no activity on my part.  Reopening to have it merged as I believe others could benefit from it.

This PR adds support to retrieve custom fields defined on the Team, which can be populated via the `users.profile.set` endpoint.  Profile field variables are custom to the team, so being able to retrieve them is important when setting new values.  These values can also be found within the profile section of the Admin Console as noted in the comment of the function.

This PR also adds support for updating the custom fields on a user using these variables.  The new function accepts a map of the existing struct `UserProfileCustomField`.  The key of the map is the custom Team Variable to define the custom field on the user.   

Note that any existing field set, but not passed to function to be updated will be cleared based on Slack's native functionality.  

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/users.go
+++ b/users.go
@@ -556,6 +556,55 @@ func (api *Client) SetUserRealNameContextWithUser(ctx context.Context, user, rea
 	return response.Err()
 }
 
+// SetUserCustomFields sets Custom Profile fields on the provided users account.  Due to the non-repeating elements
+// within the request, a map fields is required.  The key in the map signifies the field that will be updated.
+//
+// Note: You may need to change the way the custom field is populated within the Profile section of the Admin Console from
+// SCIM or User Entered to API.
+//
+// See GetTeamProfile for information to retrieve possible fields for your account.
+func (api *Client) SetUserCustomFields(userID string, customFields map[string]UserProfileCustomField) error {
+	return api.SetUserCustomFieldsContext(context.Background(), userID, customFields)
+}
+
+// SetUserCustomFieldsContext will set a users custom profile field with context.
+//
+// For more information see SetUserCustomFields
+func (api *Client) SetUserCustomFieldsContext(ctx context.Context, userID string, customFields map[string]UserProfileCustomField) error {
+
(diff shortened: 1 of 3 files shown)


MattDavisRV:
I'm not sure why the lint test is failing, I did run `make pr-prep` prior to putting this in and everything passed.

kanata2:
The failure of linter check is caused by golangci-lint's upgrade, so don't worry about this 🙏

users.go
```diff
@@ -556,6 +556,55 @@ func (api *Client) SetUserRealNameContextWithUser(ctx context.Context, user, rea
...
+	updateFields.SetMap(customFields)
+
+	// This anonymous struct is needed to set the fields level of the request data.  The base struct for
+	// UserProfileCustomFields has an unexported variable named fields that does not contain a struct tag,
+	// which has resulted in this configuration.
+	profile, err := json.Marshal(&struct {
+		Fields UserProfileCustomFields `json:"fields"`
+	}{
+		Fields: updateFields,
+	})
```
kanata2, line 587: According to the following document, It seems that you don't need to wrap by "fields" field 🤔 

https://api.slack.com/methods/users.profile.set

kanata2 (changes requested):

users.go
MattDavisRV: Under the section on that page titled "Custom profile fields", the following example request is displayed, which does have the fields section within Profile.  I've tested this code against my production slack workspace and it works as expected as well.  

```{
    "profile": {
        "fields": {
            "Xf0111111": {
                "value": "Barista",
                "alt": ""
            },
            "Xf0222222": {
                "value": "2022-04-11",
                "alt": ""
            },
            "Xf0333333": {
                "value": "https://example.com",
                "alt": ""
            }
        }
    }
}```

kanata2: Sorry, I was looking at another part of the document...

kanata2 (approved):

Issue #844, opened 2020-11-09
OAuth Slack Token Logged In Clear


christophercutajar:
**Describe the bug**

While using Argo Event which leverage `slack-go/slack` for Slack triggers, it was noticed that the OAuth Slack token is logged in the clear. Specifically, https://github.com/slack-go/slack/blob/686c209f9525a78313cfe54cbc07cd86bd677384/chat.go#L217 is logging the whole API request resulting in also logging a clear version of the OAuth slack token in the log files. The following is an example:

```
sensor-qrlrb-58d497d595-q9dmh main 2020-11-09T10:11:50.147597554Z slack-go/slack2020/11/09 10:11:50 chat.go:217: Sending request: channel=random&text=webhook+triggered%21&token=xoxb-xxxx-xxx-xxx
```

**Expected behavior**

OAuth Slack token not never be logged in the clear in the log files.

_**Argo Events related issue:**_ https://github.com/argoproj/argo-events/issues/944

kanata2:
Thanks for reporting.
As you said in argoproj/argo-events#944, `api.Debugf` writes only if debug mode is enabled.
So I think it's not a bug 🤔
What do you think?

christophercutajar:
For the Argo part, I agree with you, the `api.Debugf` should be false. 

In general, my personal opinion is that I don't think such token should be logged in the log files even in `debug` mode.

kanata2:
Okay, it seems good to me. Please send a PR?

christophercutajar:
@kanata2 will be able to submit a PR in the coming 2 days.

kanata2:
Thanks!

kanata2:
@christophercutajar I'll close it once, but feel free to re-open it or send us a pull request if you want!

briemarie:
@kanata2 We're seeing the same behavior of the token being printed out into our logs. We've updated to the latest version as of today v0.11.2 but still getting the printout. We have these debug logs being stored in dashboards so this is a major problem. Has it been fixed? It is never ideal to have tokens printed in any logs, debug or dev or whatever.

christophercutajar:
@kanata2 my apologies I wasn't able to create a PR for this. Can you re-open the ticket please as not I'm able to work on this.

@briemarie this wasn't solved from my end :(

briemarie:
[...]

PR #1102, opened 2022-08-24
Issue 844: Redact api token in chat message

briemarie:
Debug logs are printing api tokens as part of the request message.
Instead of printing the token, the token is transformed into a redacted
string if it is not empty. If the token is empty, which is a useful
piece of information that should be surfaced in the request, the empty
string is transformed into a nil string to make it more obvious.

##### Pull Request Guidelines

##### PR preparation
[PASSED] Run `make pr-prep` from the root of the repository to run formatting, linting and tests.

##### Should this be an issue instead - YES IT IS #844 
- [X] exposes a previously private type, const, method, etc.

##### API changes

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -243,6 +243,17 @@ func UnsafeApplyMsgOptions(token, channel, apiurl string, options ...MsgOption) 
 }
 
 func applyMsgOptions(token, channel, apiurl string, options ...MsgOption) (sendConfig, error) {
+	// It maybe be useful to know if the token supplied was empty or a token,
+	// however we do not want tokens printed out in messages. If the token is
+	// empty, change the token to a nil string to make it obvious in the message.
+	// If the token is not empty, change the token to a redacted string.
+	switch token {
+	case "":
+		token = "nil"
+	default:
+		token = "<REDACTED>"
+	}
+
 	config := sendConfig{
 		apiurl:   apiurl,
 		endpoint: apiurl + string(chatPostMessage),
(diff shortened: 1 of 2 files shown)


christophercutajar (approved):
LGTM

briemarie:
@christophercutajar I see you approved this. Will it be getting merged or was there another release that fixed the issue of the token being exposed in logs?

christophercutajar:
@kanata2 when you have time can you please review it yourself and merge it!

chat.go
```diff
@@ -243,6 +243,17 @@ func UnsafeApplyMsgOptions(token, channel, apiurl string, options ...MsgOption)
...
+	// It maybe be useful to know if the token supplied was empty or a token,
+	// however we do not want tokens printed out in messages. If the token is
+	// empty, change the token to a nil string to make it obvious in the message.
+	// If the token is not empty, change the token to a redacted string.
+	switch token {
+	case "":
+		token = "nil"
+	default:
+		token = "<REDACTED>"
+	}
```
kanata2, line 255: If the token is rewritten, will it not be able to authenticate correctly?

kanata2: This changes affect other than debug logs.

briemarie:
[kanata2](https://github.com/kanata2) I realize that my proposed solution wasn't the proper one, but has this been solved internally by your team in another PR?

kanata2:
No, we have not worked on it yet.
If you think the labeling by GitHub Actions is inappropriate, please re-open it and reply to my comment.

briemarie:
@kanata2 Its ok that this PR is closed since it wasn't the right solution. Looks like the original issue https://github.com/slack-go/slack/issues/844 is still open so hopefully that gets addressed.

PR #1105, opened 2022-08-30
adds postephemeral handler to slacktest to audit outgoing messages

mtintes:
##### The Changes
PostEphemeral was not exposing messages in the outgoing messages on slacktest. This is mostly a copy-paste from how postMessage handles that with the [addition](https://github.com/slack-go/slack/compare/master...mtintes:slack:master?expand=1#diff-9b81311406fe82eca68598c5090a2b687e6ab02444abe0ec25cd3bc83c39d2ceR251) that we use the userId that gets handed into the method.

### DIFF
--- a/slacktest/handlers.go
+++ b/slacktest/handlers.go
@@ -211,6 +211,98 @@ func (sts *Server) postMessageHandler(w http.ResponseWriter, r *http.Request) {
 	_ = json.NewEncoder(w).Encode(resp)
 }
 
+// handle chat.postEphemeralMessage
+func (sts *Server) postEphemeralHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


mtintes:
Just wanted to check in on this to see if there were any questions or concerns surrounding it that I could help clear up. I kind of just threw this in because I needed it to write some tests and most of our sends are ephemeral messages.

mtintes:
Hey just wanted to circle back and see if you had any thoughts on this? Happy to do the work if you have alternative ideas.

mtintes:
hey @kanata2 I know it's been a while, but just want to follow up on this. Happy to update it and whatnot. Or if you have another idea I would happily work through it.

PR #1109, opened 2022-09-03
fix: UploadFileContext not calling api.AuthTestContext

kevin-faliam:
`UploadFileContext` was calling `api.AuthTest` instead of `api.AuthTestContext`


kanata2:
@kevin-faliam Please add description for your changes.

kanata2 (approved):
Thanks!

PR #1113, opened 2022-09-28
Add hidden subtypes to MessageEvent

hinshun:
When handling events through Socket Mode, we end up getting a `MessageEvent` from the Events API. In order to distinguish hidden subtypes, there are fields that isn't being marshaled into the struct like `Hidden bool` which will be nice to have.

### DIFF
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -218,6 +218,11 @@ type MessageEvent struct {
 	// Message Subtypes
 	SubType string `json:"subtype,omitempty"`
 
+	// Hidden Subtypes
+	Hidden           bool   `json:"hidden,omitempty"`     // message_changed, message_deleted, unpinned_item
+	DeletedTimestamp string `json:"deleted_ts,omitempty"` // message_deleted
+	EventTimestamp   string `json:"event_ts,omitempty"`
+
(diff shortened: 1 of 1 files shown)


slackevents/inner_events.go
```diff
@@ -218,6 +218,11 @@ type MessageEvent struct {
 	// Message Subtypes
 	SubType string `json:"subtype,omitempty"`
 
+	// Hidden Subtypes
+	Hidden           bool   `json:"hidden,omitempty"`     // message_changed, message_deleted, unpinned_item
+	DeletedTimestamp string `json:"deleted_ts,omitempty"` // message_deleted
+	EventTimestamp   string `json:"event_ts,omitempty"`
```
kanata2, line 224: ` EventTimestamp` has already been added in L207.

slackevents/inner_events.go
```diff
@@ -218,6 +218,11 @@ type MessageEvent struct {
 	// Message Subtypes
 	SubType string `json:"subtype,omitempty"`
 
+	// Hidden Subtypes
```
kanata2, line 221:
```suggestion
	// fields for Hidden Subtypes
```

kanata2 (changes requested):
Thanks!

PR #1115, opened 2022-10-12
Support slack events about file

andyhaskell:
This PR adds support for deserializing file_change, file_deleted, file_shared, and file_unshared events from the Slack Events API

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -182,6 +182,14 @@ type FileSharedEvent struct {
 	EventTimestamp string        `json:"event_ts"`
 }
 
+// FileUnsharedEvent represents the information associated with the File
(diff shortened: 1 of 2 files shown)


kanata2 (approved):
Thanks!

andyhaskell:
@kanata2 No problem! Btw it looks like in the 1.17 tests, `TestGetRTMInfo` is failing. I haven't used this CI system before, so I was wondering, is there a way to restart that build? At least the wording there makes it sound like a transient issue

kanata2:
Actually, this test is so fraky...
I understood test failure is not related to this changes, so I'll merge.

PR #1117, opened 2022-10-23
fix: create multipart form when multipart request

EkeMinusYou:
This PR fixs form parameter format when multipart request.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -63,13 +63,12 @@ func (e *RateLimitedError) Retryable() bool {
 	return true
 }
 
-func fileUploadReq(ctx context.Context, path string, values url.Values, r io.Reader) (*http.Request, error) {
+func fileUploadReq(ctx context.Context, path string, r io.Reader) (*http.Request, error) {
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, path, r)
 	if err != nil {
 		return nil, err
 	}
 
-	req.URL.RawQuery = values.Encode()
 	return req, nil
 }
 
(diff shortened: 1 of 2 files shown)


kanata2:
@EkeMinusYou
Could you add an explanation of what the problem is with the current implementation, or a link to the issue?

EkeMinusYou:
It is not clearly described in Slack's API documentation and is difficult to explain.

https://api.slack.com/methods/files.remote.add

However, when I checked the actual behavior, it seems that in the case of mutlipart request, the parameters must be put in the body instead of the query.  Otherwise the token cannot be passed and authentication will fail.

EkeMinusYou:
@kanata2 
It is not clearly described in Slack's API documentation and is difficult to explain.

https://api.slack.com/methods/files.remote.add

However, when I checked the actual behavior, it seems that in the case of mutlipart request, the parameters must be put in the body instead of the query.  Otherwise the token cannot be passed and authentication will fail.

samarthkathal:
Hi @EkeMinusYou @kanata2 ,
any updates on this?

context:

This worked fine for me when using `files.remote.add` using postman with same params but failed with slack-go

I was receiving `invalid_auth` when using `AddRemoteFileContext`. I had not investigated why this was happening but I did check my bot token and all other attributes.

Seems like this PR should fix that issue.

lorenzoaiello:
@samarthkathal this PR may be a valid fix for #1295 but I've having some trouble reproducing the issue to understand whether this would fix it. Would you be able to help me reproduce the issue?

jjjjeeffff:
> @samarthkathal this PR may be a valid fix for #1295 but I've having some trouble reproducing the issue to understand whether this would fix it. Would you be able to help me reproduce the issue?

@lorenzoaiello see my comment here to repro: https://github.com/slack-go/slack/issues/1295#issuecomment-2234141986

lorenzoaiello (approved):
Thank you to @EkeMinusYou for the original PR as well as @jjjjeeffff for helping to shepherd the PR to this point.

I've validated that this both a backwards compatible change as well as addresses an edge case where providing a Preview Image resulted in an `invalid_auth` response. More details are in the issue thread #1295 .

Apologies for the delay, and thank you for your patience 🙇

PR #1119, opened 2022-10-26
Add ListTeams and TeamAccess{Granted,Revoked}Events and fix missing enterprise ID

stephenwan-opal:
This PR makes a few changes useful for building Slack Enterprise Grid integrations.
- Adds support for [`team_access_granted`](https://api.slack.com/events/team_access_granted) and [`team_access_revoked`](https://api.slack.com/events/team_access_revoked) events
- Adds `ListTeams` to call [`auth.teams.list`](https://api.slack.com/methods/auth.teams.list)
- Fixes an issue where `EnterpriseID` is not filled in the events api for callback events
- Lets callers pass `TeamID` to `CreateConversation`. This is a **breaking change**, but hopefully one that prevents future breaking changes.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -38,3 +38,37 @@ func (api *Client) SendAuthRevokeContext(ctx context.Context, token string) (*Au
 
 	return api.authRequest(ctx, "auth.revoke", values)
 }
+
+type listTeamsResponse struct {
+	Teams []Team `json:"teams"`
+	SlackResponse
+}
+
+type ListTeamsParameters struct {
+	Limit  int
+	Cursor string
+}
+
+// ListTeams returns all workspaces a token can access.
+// More info: https://api.slack.com/methods/admin.teams.list
+func (api *Client) ListTeams(params ListTeamsParameters) ([]Team, string, error) {
+	return api.ListTeamsContext(context.Background(), params)
+}
+
+// ListTeams returns all workspaces a token can access with a custom context.
+func (api *Client) ListTeamsContext(ctx context.Context, params ListTeamsParameters) ([]Team, string, error) {
+	values := url.Values{
+		"token": {api.token},
+	}
+	if params.Cursor != "" {
+		values.Add("cursor", params.Cursor)
+	}
+
+	response := &listTeamsResponse{}
+	err := api.postMethod(ctx, "auth.teams.list", values, response)
+	if err != nil {
+		return nil, "", err
+	}
+
+	return response.Teams, response.ResponseMetadata.Cursor, response.Err()
+}
--- a/conversation.go
+++ b/conversation.go
@@ -342,17 +342,26 @@ func (api *Client) CloseConversationContext(ctx context.Context, channelID strin
 	return response.NoOp, response.AlreadyClosed, response.Err()
 }
 
+type CreateConversationParams struct {
+	ChannelName string
+	IsPrivate   bool
+	TeamID      string
+}
+
 // CreateConversation initiates a public or private channel-based conversation
-func (api *Client) CreateConversation(channelName string, isPrivate bool) (*Channel, error) {
-	return api.CreateConversationContext(context.Background(), channelName, isPrivate)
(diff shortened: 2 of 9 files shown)


conversation.go
```diff
@@ -337,17 +337,24 @@ func (api *Client) CloseConversationContext(ctx context.Context, channelID strin
...
 // CreateConversationContext initiates a public or private channel-based conversation with a custom context
-func (api *Client) CreateConversationContext(ctx context.Context, channelName string, isPrivate bool) (*Channel, error) {
+func (api *Client) CreateConversationContext(ctx context.Context, params CreateConversationParams) (*Channel, error) {
 	values := url.Values{
 		"token":      {api.token},
-		"name":       {channelName},
-		"is_private": {strconv.FormatBool(isPrivate)},
+		"name":       {params.ChannelName},
+		"is_private": {strconv.FormatBool(params.IsPrivate)},
+		"team_id":    {params.TeamID},
```
kanata2: I'm not sure I understand the behavior when passing an empty parameter... it is better to skip if these fields are empty.

auth.go
```diff
@@ -38,3 +38,38 @@ func (api *Client) SendAuthRevokeContext(ctx context.Context, token string) (*Au
...
+	Teams []Team `json:"teams"`
+	SlackResponse
+}
+
+type ListTeamsParameters struct {
+	Limit  int
+	Cursor string
+}
+
+// ListTeams returns all workspaces a token can access.
```
kanata2, line 52: Please add link to this API's doc in comment.
https://api.slack.com/methods/admin.teams.list

auth.go
```diff
@@ -38,3 +38,38 @@ func (api *Client) SendAuthRevokeContext(ctx context.Context, token string) (*Au
...
+func (api *Client) ListTeams(params ListTeamsParameters) ([]Team, *ListTeamsParameters, error) {
+	return api.ListTeamsContext(context.Background(), params)
+}
+
+// ListTeams returns all workspaces a token can access with a custom context.
+func (api *Client) ListTeamsContext(ctx context.Context, params ListTeamsParameters) ([]Team, *ListTeamsParameters, error) {
```
kanata2: [IMO]
I think the `ListTeamsParameter` is a struct for API calls, so I prefer to return next cursor instead of `ListTeamsParameter`. What do you think?

kanata2 (changes requested):
Thank you for submitting PR!

auth.go
stephenwan-opal: do you mean returning `([]Team, string, error)`? I am okay with that too - I originally did this as I was following the example from https://github.com/stephenwan-opal/slack/blob/master/files.go#L288 and https://github.com/stephenwan-opal/slack/blob/master/remotefiles.go#L174

Let me know what you prefer

auth.go
stephenwan-opal: done - thanks!

conversation.go
stephenwan-opal: done, thanks!

stephenwan-opal:
@kanata2 thank you for reviewing - i addressed/responded to the PR comments

stephenwan-opal:
@kanata2 thanks again for reviewing - could you take a look at the new changes here?

kanata2:
Sorry for delay. I'll confirm within a few days.

auth.go
hermanschaaf: Sorry for chiming in here; I'm neither the author nor maintainer, but I would very much like to use the work in this PR. 😄  I think that yes, returning `([]Team, string, error)` is what @kanata2 meant. I guess there are some inconsistencies in the SDK, but e.g. the `GetConversationsContext` method returns a string cursor: https://github.com/slack-go/slack/blob/657f4e78d9216ca7ccefeaf841b4b9623c32926b/conversation.go#L481

stephenwan-opal: i've updated the second commit to return the cursor instead of the entire params struct

kanata2 (approved):
Thanks!
I'll release this in next minor version.

PR #1120, opened 2022-10-31
Add support for status_emoji_display_info

paultyng:
Add missing field to `users.profile.get`, `status_emoji_display_info` which is a list of all the emoji's present in the status emoji or status text and details about them.

Here is an example response:

```
{
    "ok": true,
    "profile": {
        ...
        "status_text": "Construction!",
        "status_emoji": ":construction:",
        "status_emoji_display_info": [
            {
                "emoji_name": "construction",
                "unicode": "1f6a7",
                "display_url": "https://a.slack-edge.com/production-standard-emoji-assets/14.0/apple-large/1f6a7.png"
            }
        ],
        "status_expiration": 0,
        ...
       }
}
```

### DIFF
--- a/users.go
+++ b/users.go
@@ -17,30 +17,38 @@ const (
 
 // UserProfile contains all the information details of a given user
 type UserProfile struct {
-	FirstName             string                  `json:"first_name"`
-	LastName              string                  `json:"last_name"`
-	RealName              string                  `json:"real_name"`
-	RealNameNormalized    string                  `json:"real_name_normalized"`
-	DisplayName           string                  `json:"display_name"`
-	DisplayNameNormalized string                  `json:"display_name_normalized"`
-	Email                 string                  `json:"email"`
-	Skype                 string                  `json:"skype"`
-	Phone                 string                  `json:"phone"`
-	Image24               string                  `json:"image_24"`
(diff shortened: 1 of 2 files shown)


users.go
```diff
@@ -17,30 +17,38 @@ const (
...
+	Team                   string                              `json:"team"`
+	Fields                 UserProfileCustomFields             `json:"fields"`
+}
+
+type UserProfileStatusEmojiDisplayInfo struct {
+	EmojiName    string `json:"emoji_name"`
+	DisplayAlias string `json:"display_alias,omitempty"`
+	DisplayURL   string `json:"display_url,omitempty"`
+	Unicode      string `json:"unicode,omitempty"`
 }
```
kanata2, line 52: [review memo]
java sdk's definition: https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/User.java#L118-L122

kanata2 (approved):
Thanks!
I'll merge this after confirming actual response on my end.

users.go
paultyng: `unicode` is new, I just added it (I'm an eng at Slack)

Issue #1063, opened 2022-05-16
Omitempty prevents setting replace_original to false


wojtkiewicz:
### What happened

At the moment it's impossible to send `replace_original: false` in WebhookMessage because of `omitempty`.
This would be fine if `replace_original` was set to `false` by default. Unfortunately it's not.

### Expected behavior

It should be possible to set `replace_original: false` from slack-go.

### Steps to reproduce

```
import (
	"encoding/json"
	"fmt"
)

type WebhookMessage struct {
	ReplaceOriginal bool `json:"replace_original,omitempty"`
}

func main() {
        msg := &WebhookMessage{
		ReplaceOriginal: false,
	}
	out, _ := json.Marshal(msg)
	fmt.Println(string(out))
}

// produces {}

```

### Versions
- Go: any
- slack-go/slack: 0.10.3

kanata2:
Resolved by #1122

PR #1122, opened 2022-11-03
fix: remove omitempty from replace_original and delete_orginal

yokishava:
## why
I saw the issue on https://github.com/slack-go/slack/issues/1063 and fixed it.

## what
- I removed omitempty in replaceOriginal and deleteOriginal in webhookmessage.
- I made a few changes to the webhookmessage test.


kanata2 (approved):
Thanks!

PR #1128, opened 2022-11-15
fix: panic when ParseEvent if event data not provide

sword-jin:
##### API changes

No API changes

##### Bug fixes

```go
noInnerEvent := `{"type":"event_callback"}`
slackevents.ParseEvent(json.RawMessage(noInnerEvent))
```

When json don't contains `event` field, `ParseEvent` will panic.


kanata2 (commented):
Thanks.
What is the case when the `event` field is not sent?
It would be helpful if you could give me the steps to reproduce this.

sword-jin:
> s the case when the `event` field is not sent?
> It would be helpful if you could

In my own service, integrated slack webhook, fuzz testing!

Issue #1108, opened 2022-09-02
Large File upload causes timeout


DestroyerAlpha:
### What happened
I was experimenting different file size uploads to my slack workspace. The client timed out with a file size of about 50MB. The timeout happens after 20secs consistently.
### Expected behavior
Send the file
### Steps to reproduce
slackClient.UploadFile as well as slackClient.UploadFileContext with 1 minute context timeout, both fail with:
```
slack-go/slack2022/09/02 14:00:08 misc.go:272: HTTP/2.0 408 Request Timeout
Content-Length: 14
Content-Type: text/plain
Date: Fri, 02 Sep 2022 08:30:08 GMT
Server: envoy
Via: envoy-edge-bom-qrmq
X-Edge-Backend: envoy-www
X-Slack-Edge-Shared-Secret-Outcome: no-match

stream timeout

```
### Versions
- Go: 1.18
- slack-go/slack: 0.10.2

pierrick-martin-dev:
It seems to be a serverside timeout.
The default timeout for the `http.Client{}` is 0. which means `no limit`.
The python lib for Slack got the same issue: https://github.com/slackapi/python-slack-sdk/issues/1165

As I can understand, they implemented an other function after exposing an other way in the Slack backend to upload File. They explained that here: https://github.com/slackapi/python-slack-sdk/releases/tag/v3.19.0

DestroyerAlpha:
Any plans for similar support on slack-go

sarthak-kothari:
Any plans on updating the API? Seeing similar issues

kanata2:
@flowHater Thanks for your investigation!

kanata2:
I'm reviewing @sarthak-kothari's PR (#1130) now.

kanata2:
Released v0.12.0 which includes #1130.

ghost:
Hey @kanata2 no #1130 in https://github.com/slack-go/slack/releases/tag/v0.12.0

kanata2:
Oh...sorry.

pierrick-martin-dev:
@kanata2
Can we expect a release soon with this patch ? Or are you waiting more commits to bundle the whole in a larger release ?

pierrick-martin-dev:
What I saw, is that there is still pending comments on the MR, so It's not merged and so cannot be released.

DestroyerAlpha:
[...]

PR #1130, opened 2022-11-21
Added new FileUploadV2 function to avoid server side file timeouts

sarthakkothari:
### Linking issue 
This PR solves for - https://github.com/slack-go/slack/issues/1108
Implements the methods suggested in - https://github.com/slackapi/python-slack-sdk/releases/tag/v3.19.0

Need to account for nil interface because slack doesn't give `SlackResponse` or any json response for that matter when uploading file via post to [GetUploadURLExternal](https://api.slack.com/methods/files.getUploadURLExternal)'s UploadURL

### DIFF
--- a/files.go
+++ b/files.go
@@ -524,7 +524,7 @@ func (api *Client) completeUploadExternal(ctx context.Context, fileID string, pa
 	return response, nil
 }
 
-// UploadFileV2 uploads file to a given slack channel using 3 steps - 
+// UploadFileV2 uploads file to a given slack channel using 3 steps -
 // 	1. Get an upload URL using files.getUploadURLExternal API
 //  2. Send the file as a post to the URL provided by slack
 //  3. Complete the upload and share it to the specified channel using files.completeUploadExternal
@@ -532,7 +532,7 @@ func (api *Client) UploadFileV2(params FileUploadV2Parameters) (*FileSummary, er
 	return api.UploadFileV2Context(context.Background(), params)
 }
 
-// UploadFileV2 uploads file to a given slack channel using 3 steps with a custom context - 
+// UploadFileV2 uploads file to a given slack channel using 3 steps with a custom context -
 // 	1. Get an upload URL using files.getUploadURLExternal API
 //  2. Send the file as a post to the URL provided by slack
 //  3. Complete the upload and share it to the specified channel using files.completeUploadExternal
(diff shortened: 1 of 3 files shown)


sarthakkothari:
@kanata2 making a PR for the first time, how do I request your review/add label?

kanata2:
@sarthakkothari Sorry for delay. Will review within a few days. 🙏

files.go
```diff
@@ -416,3 +455,119 @@ func (api *Client) ShareFilePublicURLContext(ctx context.Context, fileID string)
...
+		return nil, err
+	}
+	if response.Err() != nil {
+		return nil, response.Err()
+	}
+	return response, nil
+}
+
+// UploadFileV2 uploads file to a given slack channel using 3 steps -
+// 	1. Get an upload URL using files.getUploadURLExternal API
```
kanata2, line 527:
```suggestion
//  1. Get an upload URL using files.getUploadURLExternal API
```

files.go
```diff
@@ -145,6 +146,44 @@ type ListFilesParameters struct {
 	Cursor  string
 }
 
+type FileUploadV2Parameters struct {
```
kanata2, line 149: [IMO] Prefer `UploadFileV2Parameters`. It would be easier to use if aligned with the method name.

files.go
```diff
@@ -145,6 +146,44 @@ type ListFilesParameters struct {
...
+	SnippetText     string
+}
+
+type getUploadURLExternalResponse struct {
+	UploadURL string `json:"upload_url"`
+	FileID    string `json:"file_id"`
+	SlackResponse
+}
+
+type uploadToExternalParams struct {
```
kanata2, line 169: `uploadToURLParameters` ?

files.go
```diff
@@ -416,3 +455,119 @@ func (api *Client) ShareFilePublicURLContext(ctx context.Context, fileID string)
 	}
 	return &response.File, response.Comments, &response.Paging, nil
 }
+
+// getUploadURLExternal gets a URL and fileID from slack which can later be used to upload a file
+func (api *Client) getUploadURLExternal(ctx context.Context, fileSize int, fileName, altText, snippetText string) (*getUploadURLExternalResponse, error) {
```
kanata2, line 460: How about defining struct since there are many parameters?

files.go
```diff
@@ -416,3 +455,119 @@ func (api *Client) ShareFilePublicURLContext(ctx context.Context, fileID string)
...
+	} else if params.File != "" {
+		err = postLocalWithMultipartResponse(ctx, api.httpclient, params.UploadURL, params.File, "file", api.token, values, nil, api)
+	} else if params.Reader != nil {
+		err = postWithMultipartResponse(ctx, api.httpclient, params.UploadURL, params.Filename, "file", api.token, values, params.Reader, nil, api)
+	}
+	return err
+}
+
+// completeUploadExternal once files are uploaded, this completes the upload and shares it to the specified channel
+func (api *Client) completeUploadExternal(ctx context.Context, fileID string, params FileUploadV2Parameters) (file *completeUploadExternalResponse, err error) {
```
kanata2, line 497: It is preferable to define struct instead of using `FileUploadV2Prameters`.

PR #1132, opened 2022-11-22
feat: add number input block element

sedyn:
I added `NumberInputBlockElement` type. It's necessary to make input blocks which accepts only numbers.

### DIFF
--- a/block_element.go
+++ b/block_element.go
@@ -11,6 +11,7 @@ const (
 	METTimepicker     MessageElementType = "timepicker"
 	METPlainTextInput MessageElementType = "plain_text_input"
 	METRadioButtons   MessageElementType = "radio_buttons"
+	METNumber         MessageElementType = "number_input"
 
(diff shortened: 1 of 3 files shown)


block_element.go
```diff
@@ -475,3 +476,34 @@ func NewRadioButtonsBlockElement(actionID string, options ...*OptionBlockObject)
 		Options:  options,
 	}
 }
+
+// NumberInputBlockElement creates a field where a user can enter number
+// data.
+// Number input elements are currently only available in modals.
+//
+// More Information: https://api.slack.com/reference/block-kit/block-elements#number
+type NumberInputBlockElement struct {
```
kanata2, line 485: [note] focus_on_load field is missing, but I think it is OK to submit another PR for this one after merging this PR.

kanata2 (approved):

kanata2:
Thanks. LGTM!

PR #1134, opened 2022-11-24
Add team_id as a parameter to the users.conversations API method

ctennis:
This API method takes an optional team_id parameter, which is required if you're using an app/bot with organization access. This just exposes this in golang. There's no existing test for this method, so nothing for me to add on to here.


kanata2 (approved):
LGTM. Thanks!

PR #1083, opened 2022-06-29
Metadata Support

H3rby7:
## Disclaimer

I am somewhat new to this library AND to writing Go Code, feel free to correct me or suggest changes I am happily learning!

## API changes

Add support for Metadata, as requested by: 
https://github.com/slack-go/slack/issues/1057

https://api.slack.com/metadata/publishing

Changes:
* added `MsgOptionMetadata` in `chat.go`, so we can add metadata to postMessage
* added `includeAllMetadata` to conversationHistory to retrieve existing metadata.

PR #1139, opened 2022-11-30
add include_all_metadata to conversations.replies API

ebkn:
# WHAT
add `include_all_metadata` param to `conversations.replies` API
doc: https://api.slack.com/methods/conversations.replies#arg_include_all_metadata

# WHY
Metadata is supported (Thanks https://github.com/slack-go/slack/pull/1083), but does not support `conversations.replies` API.  `include_all_metadata` param is needed to get metadata.


kanata2 (approved):
Thanks!

PR #1140, opened 2022-11-30
add block elements for email_text_input and url_text_input

ameliagapin:
Adds [`email_text_input`](https://api.slack.com/reference/block-kit/block-elements#email) and [`url_text_input`](https://api.slack.com/reference/block-kit/block-elements#url) block elements, which were recently added to BlockKit, to the SDK.

I tested these by using them in a modal:
Screenshot 2022-11-30 at 9 51 11 AM

And the values in the request from Slack:
Screenshot 2022-11-30 at 9 52 14 AM

### DIFF
--- a/block_element.go
+++ b/block_element.go
@@ -402,6 +402,7 @@ type EmailTextInputBlockElement struct {
 	Placeholder          *TextBlockObject      `json:"placeholder,omitempty"`
 	InitialValue         string                `json:"initial_value,omitempty"`
 	DispatchActionConfig *DispatchActionConfig `json:"dispatch_action_config,omitempty"`
+	FocusOnLoad          bool                  `json:"focus_on_load,omitempty"`
 }
 
 // ElementType returns the type of the Element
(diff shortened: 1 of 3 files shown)


block_element.go
```diff
@@ -389,6 +391,62 @@ func NewTimePickerBlockElement(actionID string) *TimePickerBlockElement {
...
+		Placeholder: placeholder,
+	}
+}
+
+// URLTextInputBlockElement creates a field where a user can enter url data.
+//
+// url-text-input elements are currently only available in modals.
+//
+// More Information: https://api.slack.com/reference/block-kit/block-elements#url
+type URLTextInputBlockElement struct {
```
kanata2, line 429: [note] `focus_on_load` field is missing, but I think it is OK to submit another PR for this one after merging this PR.

block_element_test.go
```diff
@@ -142,6 +142,20 @@ func TestNewPlainTextInputBlockElement(t *testing.T) {
 
 }
 
+func TestNewEmailTextInputBlockElement(t *testing.T) {
+	emailTextInputElement := NewEmailTextInputBlockElement(nil, "test@test.com")
```
kanata2: [IMO]
Prefer to use `example.com` for testing data.
(Existing test data should be corrected as well...)

kanata2 (changes requested):
Sorry for delay. I commented on a couple of minor points.

block_element.go
ameliagapin: I originally left it off because it doesn't look like `focus_on_load` is included on the other elements that support it. I've added it now,  though.

block_element_test.go
ameliagapin: Done

kanata2 (approved):
Thanks!

PR #1142, opened 2022-12-13
Add an example of updating modal

KouWakai:
Hi!! I added updating modal example

since I wanted to know how could I handle updating modal in examples.

### DIFF
--- a/examples/modal/modal.go
+++ b/examples/modal/modal.go
@@ -60,6 +60,30 @@ func generateModalRequest() slack.ModalViewRequest {
 	return modalRequest
 }
 
+func updateModal() slack.ModalViewRequest {
+	// Create a ModalViewRequest with a header and two inputs
+	titleText := slack.NewTextBlockObject("plain_text", "My App", false, false)
+	closeText := slack.NewTextBlockObject("plain_text", "Close", false, false)
+	submitText := slack.NewTextBlockObject("plain_text", "Submit", false, false)
+
+	headerText := slack.NewTextBlockObject("mrkdwn", "Modal updated!", false, false)
+	headerSection := slack.NewSectionBlock(headerText, nil, nil)
+
+	blocks := slack.Blocks{
+		BlockSet: []slack.Block{
+			headerSection,
+		},
+	}
+
+	var modalRequest slack.ModalViewRequest
+	modalRequest.Type = slack.ViewType("modal")
+	modalRequest.Title = titleText
+	modalRequest.Close = closeText
+	modalRequest.Submit = submitText
+	modalRequest.Blocks = blocks
+	return modalRequest
+}
+
 // This was taken from the slash example
 // https://github.com/slack-go/slack/blob/master/examples/slash/slash.go
 func verifySigningSecret(r *http.Request) error {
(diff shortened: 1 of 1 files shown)


examples/modal/modal.go
```diff
@@ -149,6 +173,22 @@ func handleModal(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusUnauthorized)
 		return
 	}
+
+	// update modal sample
+	switch i.Type {
+	//update when interaction type is view_submission
+	case slack.InteractionTypeViewSubmission:
+		api := slack.New("YOUR_TOKEN_HERE")
```
kanata2: Already defined in L167.

examples/modal/modal.go
```diff
@@ -149,6 +173,22 @@ func handleModal(w http.ResponseWriter, r *http.Request) {
...
+		//you can use any modal you want to show to users just like creating modal.
+		updateModal := updateModal()
+		//you can use hash and ViewID from payload and externalID if exists
```
kanata2:
```suggestion
		// You must set one of external_id or view_id and you can use hash for avoiding race condition.
		// More details: https://api.slack.com/surfaces/modals/using#updating_apis
```

examples/modal/modal.go
```diff
@@ -149,6 +173,22 @@ func handleModal(w http.ResponseWriter, r *http.Request) {
...
+		_, err := api.UpdateView(updateModal, "", i.View.Hash, i.View.ID)
+		if err != nil {
+			fmt.Printf("Error updating view: %s", err)
+			w.WriteHeader(http.StatusUnauthorized)
```
kanata2:
```suggestion
			w.WriteHeader(http.StatusInternalServerError)
```

KouWakai:
Thank you for feedback:pray:
I would modify codes to the suggestion

examples/modal/modal.go
KouWakai: Sure,I would delete L181.

examples/modal/modal.go
KouWakai: I see,this should be StatusInternalServerError.

examples/modal/modal.go
```diff
@@ -149,6 +173,22 @@ func handleModal(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusUnauthorized)
 		return
 	}
+
+	// update modal sample
+	switch i.Type {
+	//update when interaction type is view_submission
+	case slack.InteractionTypeViewSubmission:
```
charlesoconor, line 172: I couldn't get `api.UpdateView` to work under this event. I had to use `slack.InteractionTypeBlockActions` e.g `block_action`. Is this working for you?

KouWakai: Thank you for the feedback,I would check that.

KouWakai: It's working but modal closes right after update.

I think this is caused by slack server.
after update modal,app respond http status 200 to slack server and the server see it as over

I'v been running this code on local and connect via ngrok and this happen

KouWakai: by adding `time.Sleep(time.Second * 2)` after api.UpdateView you can prevent sending http 200 and can see updated modal.

```go
	case slack.InteractionTypeViewSubmission:
		//you can use any modal you want to show to users just like creating modal.
		updateModal := updateModal()
		// You must set one of external_id or view_id and you can use hash for avoiding race condition.
		// More details: https://api.slack.com/surfaces/modals/using#updating_apis
		_, err := api.UpdateView(updateModal, "", i.View.Hash, i.View.ID)
		time.Sleep(time.Second * 2)
		if err != nil {
			fmt.Printf("Error updating view: %s", err)
			w.WriteHeader(http.StatusInternalServerError)
			return
		}
```

KouWakai: @kanata2 @charlesoconor 
sorry for kept you waiting,could you review this code?

parsley42 (approved):
Even if this example isn't perfect, it's an expansion of what we had before.

PR #1145, opened 2022-12-19
add datetimepicker component

hussachai:
This is a new component that Slack supports `Datetimepicker` where date and time components are in the same row.
This new component takes number which I think it should be int64 for golang because Go's epoch API returns int64 data type. 
Since `datetimepicker` takes two columns, it's not supported in Accessory.  So, I don't include it as part of Section serialization/deserialization. 

PS. Sorry I deleted and restored a couple times. First didn't realize that I pushed the upstream, then I changed my mind because I tested it and it's working as expected. 

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/block_conv.go
+++ b/block_conv.go
@@ -110,6 +110,8 @@ func (b *InputBlock) UnmarshalJSON(data []byte) error {
 		e = &DatePickerBlockElement{}
 	case "timepicker":
 		e = &TimePickerBlockElement{}
+	case "datetimepicker":
+		e = &DateTimePickerBlockElement{}
 	case "plain_text_input":
 		e = &PlainTextInputBlockElement{}
 	case "email_text_input":
@@ -190,6 +192,8 @@ func (b *BlockElements) UnmarshalJSON(data []byte) error {
 			blockElement = &DatePickerBlockElement{}
 		case "timepicker":
 			blockElement = &TimePickerBlockElement{}
+		case "datetimepicker":
+			blockElement = &DateTimePickerBlockElement{}
 		case "plain_text_input":
 			blockElement = &PlainTextInputBlockElement{}
 		case "email_text_input":
@@ -233,6 +237,7 @@ func (a *Accessory) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON implements the Unmarshaller interface for Accessory, so that any JSON
 // unmarshalling is delegated and proper type determination can be made before unmarshal
+// Note: datetimepicker is not supported in Accessory
 func (a *Accessory) UnmarshalJSON(data []byte) error {
 	var r json.RawMessage
 
--- a/block_element.go
+++ b/block_element.go
@@ -9,6 +9,7 @@ const (
 	METOverflow       MessageElementType = "overflow"
 	METDatepicker     MessageElementType = "datepicker"
 	METTimepicker     MessageElementType = "timepicker"
+	METDatetimepicker MessageElementType = "datetimepicker"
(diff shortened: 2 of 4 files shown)


block_element.go
```diff
@@ -392,6 +393,30 @@ func NewTimePickerBlockElement(actionID string) *TimePickerBlockElement {
 	}
 }
 
+// DateTimePickerBlockElement defines an element that allows the selection of both
+// a date and a time of day formatted as a UNIX timestamp.
+// More Information: https://api.slack.com/reference/messaging/block-elements#datetimepicker
+type DateTimePickerBlockElement struct {
+	Type            MessageElementType       `json:"type"`
+	ActionID        string                   `json:"action_id,omitempty"`
+	Placeholder     *TextBlockObject         `json:"placeholder,omitempty"`
```
kanata2: According to document and Java SDK, it seems not to exist this field.

- https://api.slack.com/reference/block-kit/block-elements#datetimepicker
- https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/block/element/DatetimePickerElement.java

hussachai: My bad. I realized this and forgot to remove it🤦‍♂️. Now it's fixed. Thank you for your feedback.

Houndie:
Thank you for doing this!  Saved me some time for doing it myself <3

block_conv.go
```diff
@@ -233,6 +237,7 @@ func (a *Accessory) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON implements the Unmarshaller interface for Accessory, so that any JSON
 // unmarshalling is delegated and proper type determination can be made before unmarshal
+// Note: datetimepicker is not supported in Accessory
```
kanata2, line 240: [Question] Is it correct to say that this is because datetimepicker does not work with Section?

hussachai: I believe so, and the builder complains about this https://app.slack.com/block-kit-builder/T3E1TGDM3#%7B%22blocks%22:%5B%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22Pick%20a%20date%20for%20the%20deadline.%22%7D,%22accessory%22:%7B%22type%22:%22datetimepicker%22,%22action_id%22:%22datepicker-action%22%7D%7D%5D%7D

I assume that the space is not permitted for the layout. I hope Slack supports grid layout like Teams in future. Currently, everything seems to be limited to 2 columns at most.

hussachai: I might be wrong on this. Let me check one more time. Looks like it is supported
https://github.com/slackapi/java-slack-sdk/commit/4176b912895c57e99de5d3a8ee691807801159f2#diff-212bb028a409f158a94e9af3431b61e080b8bec0e72711ba99dc294e02fd5084L2866

hussachai: Working on it. Should be done soon. Thanks for asking this. The assumption needs to be tested.

hussachai: Hmm.  I tested it out, and my assumption is right.
```
0 = {string} "[ERROR] failed to match all allowed schemas [json-pointer:/view]"
1 = {string} "[ERROR] unsupported type: datetimepicker [json-pointer:/view/blocks/8/accessory/type]"
```
datetimepicker won't work in Section.  That kinda makes sense to me because I don't see any room to put text on it. It will go overflow.

kanata2 (approved):
LGTM!

PR #1147, opened 2022-12-20
fix: change period to comma

yokishava:
### Why
The character before omitempty in the RichTextSectionElement struct tag was period.

### What
I changed to comma.

### DIFF
--- a/block_rich_text.go
+++ b/block_rich_text.go
@@ -294,7 +294,7 @@ func NewRichTextSectionLinkElement(url, text string, style *RichTextSectionTextS
 type RichTextSectionTeamElement struct {
 	Type   RichTextSectionElementType `json:"type"`
 	TeamID string                     `json:"team_id"`
-	Style  *RichTextSectionTextStyle  `json:"style.omitempty"`
+	Style  *RichTextSectionTextStyle  `json:"style,omitempty"`
 }
 
 func (r RichTextSectionTeamElement) RichTextSectionElementType() RichTextSectionElementType {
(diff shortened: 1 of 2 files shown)


block_rich_text.go
```diff
@@ -294,7 +294,7 @@ func NewRichTextSectionLinkElement(url, text string, style *RichTextSectionTextS
 type RichTextSectionTeamElement struct {
 	Type   RichTextSectionElementType `json:"type"`
 	TeamID string                     `json:"team_id"`
-	Style  *RichTextSectionTextStyle  `json:"style.omitempty"`
+	Style  *RichTextSectionTextStyle  `json:"style,omitempty"`
```
kanata2, line 297: Thanks 🙏

block_rich_text.go
```diff
@@ -312,64 +312,76 @@ func NewRichTextSectionTeamElement(teamID string, style *RichTextSectionTextStyl
...
 	UsergroupID string                     `json:"usergroup_id"`
+	Style       *RichTextSectionTextStyle  `json:"style,omitempty"`
 }
 
 func (r RichTextSectionUserGroupElement) RichTextSectionElementType() RichTextSectionElementType {
 	return r.Type
 }
 
-func NewRichTextSectionUserGroupElement(usergroupID string) *RichTextSectionUserGroupElement {
+func NewRichTextSectionUserGroupElement(usergroupID string, style *RichTextSectionTextStyle) *RichTextSectionUserGroupElement {
```
kanata2: This breaks backward-compatibility, so if you want to use this immediately, maybe divide this PR into another one.

kanata2: (If there is nothing wrong with including next minor release, you don't divide.)

yokishava: Thanks for the review.
You are right about breaking backward compatibility.  This PR should only be a comma modification of `omitempty`, and the style addition changes should be created as a new separate PR.

yokishava: PR was changed to comma change only.
The title and description of the PR have also been changed.

PR #1153, opened 2022-12-27
fix: ScheduleMessage not returning scheduled_message_id

hussachai:
I'd like to introduce the fix for `ScheduledMessage` API. Currently, it doesn't return scheduled_message_id where this ID is needed for cancelling the scheduled message. My change is the least intrusive I could think of. It doesn't introduce any new field or adding new API. Basically, this change should be backward compatible. It is based on the fact that the API [chat.postMessage](https://api.slack.com/methods/chat.postMessage#examples) doesn't return `scheduled_message_id` and the API [chat.scheduledMessage](https://api.slack.com/methods/chat.scheduleMessage#examples) doesn't return either `message.ts` or `ts`. We just need to check the presence of this field and determine the caller API based on that.

Existing code shouldn't break because the `timestamp` is always an empty string for `ScheduledMessage` API.
Now it will return `scheduled_message_id` in place of `timestamp`. 
`SendMessage` should remain the same.

##### Pull Request Guidelines

##### PR preparation

##### API changes


TopherGopher:
This would be amazing to have - there's no way around actually deleting this message.

jvfrodrigues (approved):
Could we get this up?

winston-stripe:
@hussachai any chance you could fix the merge conflicts that are currently blocking this PR? Would love to see this bug fix merged at some point 😄

hussachai:
> @hussachai any chance you could fix the merge conflicts that are currently blocking this PR? Would love to see this bug fix merged at some point 😄

Absolutely! It’s a simple PR. I can’t believe it’s already been about 2 years. 😄

Adam-Haas:
I really hope this is reviewed soon. A much needed feature

lorenzoaiello (approved):

lorenzoaiello:
> I really hope this is reviewed soon. A much needed feature

Indeed, thank you @hussachai for putting this together!

Issue #1146, opened 2022-12-20
Style is not defined for some RichTextSectionElement


yokishava:
### What happened

The following RichTextSectionElement could not specify a Style.

- RichTextSectionTeamElement
- RichTextSectionDateElement
- RichTextSectionBroadcastElement
- RichTextSectionColorElement

For example, On the Block Kit Builder RichTextSectionDateElement can define format and fallback, but is not defined in Struct.

### Expected behavior

I would like to be able to specify a Style(RichTextSectionTextStyle).

### Versions
- Go:
- slack-go/slack:

kanata2:
@yokishava 
Please give me a sample block kit builder URL 😢

PR #1156, opened 2023-01-06
feat: add a style to a style that does not exist

yokishava:
This PR is the corresponding change to #1146 of issue.
Changes are made to the API, breaking backward compatibility.

Style to fields and arguments.


kanata2:
It seems that this is not implemented in Java SDK.
I would like to consider merging when I can reproduce the actual behavior at hand.

PR #1164, opened 2023-02-05
socketmode example: use embedded api client in socketmode client

kittydoor:
The socketmode example in its current version is confusing for new users. Although a new socketmode client embeds the provided api client into itself, thus inheriting all methods, the example uses the earlier defined api to call said command. This gives the impression that you need to still keep the original client.

One could even go further by creating the api inline with the socketmode client.

To be honest, elaborating a bit more on the socketmode example would be helpful (or at least would've been helpful to me when I first started using it), but that's out of scope for this tiny PR.

Also, accidentally fixed a small formatting error.


kanata2 (approved):
👍

PR #1165, opened 2023-02-05
socketmode remove duplicate debug logging

kittydoor:
`smc.runRequestHandler` and `smc.runMessageReceiver` are both run as goroutines. 

Respectively, when debug is enabled, they log the event:
- `smc.Debugf("Received WebSocket message: %s", message)`
- `smc.Debugln("Incoming WebSocket message:", reencoded)` (in receiveMessagesInto)

One of these should be removed. This PR removes the latter as it re-marshalls the json, and pretty prints (and was the offender that made me investigate this in the first place).

I'm open to other ideas too, as long as it means the end result is no pretty-printed log line, and no duplicate logging of the same data in full.

### DIFF
--- a/socketmode/socket_mode_managed_conn.go
+++ b/socketmode/socket_mode_managed_conn.go
@@ -1,7 +1,6 @@
 package socketmode
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


kanata2 (approved):
LGTM. Thanks!

PR #1173, opened 2023-02-16
Make UpdateUserGroupMembers arguments to be more consistent

### LINKED ISSUE #1172: Change UpdateUserGroupMembers to accept slice of group members
It's really inconvenient api to get members of the group
https://github.com/slack-go/slack/blob/1edf0c7384de52bdeca7a2d6c4ad247d59c683e6/usergroups.go#L271-L283

And then if I need to perform an update - I need to do extra things 
https://github.com/slack-go/slack/blob/1edf0c7384de52bdeca7a2d6c4ad247d59c683e6/usergroups.go#L285-L302

painhardcore:
Change is pretty simple and needed to make flow more straightforward 
Get members from group --> Update group with members 
Get slice --> send slice 
closes #1172

### DIFF
--- a/usergroups.go
+++ b/usergroups.go
@@ -283,16 +283,19 @@ func (api *Client) GetUserGroupMembersContext(ctx context.Context, userGroup str
 }
 
 // UpdateUserGroupMembers will update the members of an existing user group
-func (api *Client) UpdateUserGroupMembers(userGroup string, members string) (UserGroup, error) {
+func (api *Client) UpdateUserGroupMembers(userGroup string, members []string) (UserGroup, error) {
 	return api.UpdateUserGroupMembersContext(context.Background(), userGroup, members)
 }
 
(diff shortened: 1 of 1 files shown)


johanmcos:
I'm not a maintainer, but I think if you want any chance of this being accepted, you're going to need to figure out how to make your change backwards-compatible. 

Perhaps you could use a [variadic function](https://gobyexample.com/variadic-functions) argument for `members`

karnikg:
This perhaps is not a needed/relevant change? I looked at forking the repo to make this change, and realized from the documentation that the actual API needs a comma separated string
```
A comma separated string of encoded user IDs that represent the entire list of users for the user group.
```
https://api.slack.com/methods/usergroups.users.update
Passing a comma separated string instead seemed to work just fine for me (instead of a list)

PR #1177, opened 2023-02-27
Export the Binder type in slacktest

### LINKED ISSUE #1176: Expose the "Binder" type in slacktest
### Description

TL;DR: Please change `binder` to `Binder` in `slacktest/server.go`.

The `binder` type is defined as such in `slacktest/server.go`:

```
type binder func(Customize)
```

This is only used in the signature of `func NewTestServer(custom ...binder) *Server`.

Due to the order of initialization of the handlers, once a handler is initialized, calling the `Handler` method again will be ignored. The `NewTestServer` function initializes a bunch of handlers.

In one of our tests, we wanted to redefine `/conversations.info` and some others, and the only way we could do so was to supply them like so:

```go
s := slacktest.NewTestServer(
  func(c slacktest.Customize) {
    c.Handle("/chat.postEphemeral", chatPostEphemeralHandler(...))
  },
  func(c slacktest.Customize) {
    c.Handle("/conversations.info", conversationsInfoHandler(...))
  },
  func(c slacktest.Customize) {
    c.Handle("/chat.postMessage", chatPostMessageHandler(...))
  },
)
```

However this is not amenable to table-driven tests. We'd like to be able to create our own `[]slacktest.Binder` and pass these to `slacktest.NewTestServer` running in a subtest.

```go
t.Run(testcase.Name, func(t *testing.T) {
  s := slacktest.NewTestServer(testcase.CustomFuncs...)
  // ...
}
```

However, because `binder` is private, there is no way to do this. 

With `Binder` exported we can create `[]slacktest.Binder` as an entry in each row of our table-driven tests and then pass this slice to `NewTestServer` as we would prefer.

kpaulisse:
Fixes https://github.com/slack-go/slack/issues/1176

The `binder` type is defined as such in `slacktest/server.go`:

```
type binder func(Customize)
```

This PR changes `slacktest.binder` to `slacktest.Binder`. The reason for doing this is to be able to define a slice of custom functions to pass to the constructor. This makes things much easier for table-driven tests. Without this change, each entry in the table has to build its own `*slacktest.Server` which is less than convenient.

As far as I can tell these are the only places in the code where `binder` is used explicitly. There are only a few tests in this entire project that actually do these customizations (three tests in `websocket_managed_conn_test.go`). I've explicitly tested that this change has no effect on those tests.

### DIFF
--- a/slacktest/server.go
+++ b/slacktest/server.go
@@ -26,10 +26,10 @@ type Customize interface {
 	Handle(pattern string, handler http.HandlerFunc)
 }
 
-type binder func(Customize)
+type Binder func(Customize)
 
 // NewTestServer returns a slacktest.Server ready to be started
-func NewTestServer(custom ...binder) *Server {
+func NewTestServer(custom ...Binder) *Server {
 	serverChans := newMessageChannels()
 
 	channels := &serverChannels{}


kanata2 (approved):
👍

PR #1179, opened 2023-03-02
Add support for team_id in UserGroup functions

tenyo:
This PR adds support for `team_id` in all of the `usergroups` functions, so they can work in Slack Enterprise Grid with multiple workspaces. For reference see https://api.slack.com/enterprise/apps/changes-apis#methods

I tried to stick to the established pattern of using Options variadic parameters to pass the team_id (together with other optional arguments). I don't think the changes affect existing clients using a single workspace.
I've tested all modified functions against the Slack API in an Enterprise Grid instance.

### DIFF
--- a/usergroups.go
+++ b/usergroups.go
@@ -74,6 +74,10 @@ func (api *Client) CreateUserGroupContext(ctx context.Context, userGroup UserGro
 		values["channels"] = []string{strings.Join(userGroup.Prefs.Channels, ",")}
 	}
 
+	if userGroup.TeamID != "" {
+		values["team_id"] = []string{userGroup.TeamID}
+	}
+
 	response, err := api.userGroupRequest(ctx, "usergroups.create", values)
 	if err != nil {
 		return UserGroup{}, err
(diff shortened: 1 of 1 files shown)


tenyo:
Hi there 👋  How can we get the Github tests to run here?

usergroups.go
```diff
@@ -74,25 +74,66 @@ func (api *Client) CreateUserGroupContext(ctx context.Context, userGroup UserGro
...
+
 	response, err := api.userGroupRequest(ctx, "usergroups.create", values)
 	if err != nil {
 		return UserGroup{}, err
 	}
 	return response.UserGroup, nil
 }
 
+// DisableUserGroupOption options for the DisableUserGroup and EnableUserGroup method calls.
+type DisableUserGroupOption func(*DisableUserGroupParams)
```
nlopes, line 89: 👋 thank you for the PR and apologies it took us this long to get to it.

If you're still up for it, I think a few changes are worth making (feel free to let me know if you're no longer interested in adjusting this PR and I can take care of it).

1. Let's rebase first
2. Either
a) change the name so that it's generic enough to use for both Disable and Enable? Having users use a "DisableUserGroupOption" on an EnableUserGroup method is counterintuitive to me.
b) type alias this to an "EnableUserGroupOption" (and "EnableUserGroupParams" as well)

nlopes (changes requested):

PR #1182, opened 2023-03-16
add support for user_profile_changed callback event

verygoodsoftwarenotvirus:
##### PR preparation

##### API changes
n/a

This change allows for the Slack library to parse incoming callback events of the `user_profile_changed` type. Previously, `parseInnerEvent` threw an error because it didn't have record of this event type.


kanata2 (approved):
👍

PR #1183, opened 2023-03-24
add 2FA type to slack user

imilchev:
Added a missing property to Slack user. Since it is optional (as documented [here](https://api.slack.com/types/user#fields)), I made it optional in the struct too

### DIFF
--- a/users.go
+++ b/users.go
@@ -130,7 +130,7 @@ type User struct {
 	IsAppUser         bool           `json:"is_app_user"`
 	IsInvitedUser     bool           `json:"is_invited_user"`
 	Has2FA            bool           `json:"has_2fa"`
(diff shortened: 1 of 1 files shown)


users.go
```diff
@@ -130,6 +130,7 @@ type User struct {
 	IsAppUser         bool           `json:"is_app_user"`
 	IsInvitedUser     bool           `json:"is_invited_user"`
 	Has2FA            bool           `json:"has_2fa"`
+	TwoFactorType     *string        `json:"two_factor_type,omitempty"`
```
kanata2:
```suggestion
	TwoFactorType     *string        `json:"two_factor_type"`
```

kanata2 (approved):
Thanks!

PR #1190, opened 2023-04-06
Add endpoints for `calls.*` apis and `Type: call` in blockkit

winston-stripe:
Implement the API methods for the Calls API in Slack https://api.slack.com/apis/calls

Implemented methods
- `calls.add` - Indicate a new call has been started
- `calls.end` - Indicate to slack that the call has ended
- `calls.info` - Get information about an ongoing slack call object
- `calls.update` - update call information
- `calls.participants.add`
- `calls.participants.remove`

Additionally, I've added the minimal version of `Block{Type: "call", CallID: string}` which slack recommends/requires be posted back to the channel https://api.slack.com/apis/calls#post_to_channel.

All implemented functionality is publicly documented. There appear to be additional attributes on the `type: call` block, however those appear to be internal values for slack's rendering, so I have left them out. See this gist for specific responses https://gist.github.com/winston-stripe/0cac608bd63b42d73a352be53577f7fd

##### Pull Request Guidelines

##### PR preparation

##### API changes


winston-stripe:
Oddly, I have no idea why the lint step is failing. Downloaded and ran `golangci-lint run` locally and everything ran without a hitch. Maybe some weird bad caching or some unexpected actions interaction? Fwiw, when looking through their docs, I did notice that
> v3.0.0+ requires explicit setup-go installation step prior to using this action: uses: actions/setup-go@v3

Still no idea otherwise esp since master is passing, i'm fully based off of master, and all failing files are other ones I didn't touch here 😕.

winston-stripe:
Thanks @kanata2 for fixing the repo's linting check in #1194! rebased off master and everything is passing now

winston-stripe:
bump: @kanata2 is it possible to get a 👀 on this?

calls.go
```diff
@@ -0,0 +1,206 @@
...
+	return response.Call, response.Err()
+}
+
+// EndCall ends a Call.
+func (api *Client) EndCall(callID string) error {
+	return api.EndCallContext(context.Background(), callID)
+}
+
+// EndCallContext ends a Call.
+func (api *Client) EndCallContext(ctx context.Context, callID string) error {
```
lorenzoaiello: Can you add the ability to pass the [duration](https://api.slack.com/methods/calls.end#arg_duration) optional argument when ending the call?

calls.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	response := &SlackResponse{}
+	if err := api.postMethod(ctx, "calls.end", values, response); err != nil {
+		return err
+	}
+	return response.Err()
+}
+
+// CallAddUsers adds users to a Call.
```
lorenzoaiello: How would you feel about renaming these functions to use `Participant` instead of `User` to more closely align to Slack's official terminology and API methods?

lorenzoaiello (changes requested):
Thanks for the awesome addition @winston-stripe ! I've left just a few minor items for you to review. I'm excited to get this merged and see what folks build with this.

winston-stripe:
ack, finally have time to look back into this. Hopefully get it cleaned up again today. Thanks for the feedback @lorenzoaiello!

winston-stripe:
@lorenzoaiello I updated the params + `s/users/participants/` for the Go side of the API to make it a bit more clear & consistent. LMK your thoughts :)

lorenzoaiello (approved):
Looks good, thank you for the additions!

PR #1198, opened 2023-05-06
add endpoint to socketmode Clinet and OptionAPIURL function

ToshihitoKon:
##### Pull Request Guidelines

##### PR preparation

##### API changes

---
### Change points

- Added an `endpoint` member to SocketMode Client
- Added an `socketmode.OptionAPIURL` function

### Motivation

I'm creating an application that acts as a server for a Slack App.
To do that, we need to point the Endpoint accessed from slack-go/slack away from `https://api.slack.com`.

The slack client provided a way to change the Endpoint URL, while the socketmode client did not. I also found the URL hardcoded in the `openAndDial` function.
This PR prepares the same `OptionAPIURL` for socketmode client as for slack client.

I would like opinions on whether this change should go into slack-go/slack.

### DIFF
--- a/socketmode/client.go
+++ b/socketmode/client.go
@@ -58,6 +58,7 @@ type Client struct {
 	// Dialer.
 	dialer *websocket.Dialer
 
-	debug bool
-	log   ilogger
+	endpoint string
+	debug    bool
+	log      ilogger
 }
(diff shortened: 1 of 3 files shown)


ToshihitoKon:
This change was unnecessary.

I figured out that the websocket URL is taken from the result of the `apps.connections.open` API request and the endpoint of the websocket module is irrelevant.

PR #1201, opened 2023-05-17
docs: Add an example demonstrating correct usage of GetUsersPaginated

adamrothman:
I had a really hard time finding an example of how `(*Client).GetUsersPaginated()` is intended to be used. After debugging my own implementation for the better part of a day, I thought it would be nice to add an official example.

### DIFF
--- a/examples/pagination/pagination.go
+++ b/examples/pagination/pagination.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"context"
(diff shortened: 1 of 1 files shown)


lorenzoaiello (approved):
Thanks for the contribution @adamrothman ! It seems some folks have already benefited from the example and I've validated it locally in my test environment as well.

Issue #1207, opened 2023-06-22
Support link unfurling properties in slack.WebhookMessage


lheiskan:
### Description

Slack allows setting 'unfurl_links' and 'unfurl_media' properties for webhook message payloads to control linnk unfurling behavior of links in the message.  slack.WebhookMessage does not allow setting these properties. Would ne nice if they were supported in the API.

### (Optional) Slack's documentation
https://api.slack.com/reference/messaging/link-unfurling#classic_unfurl

PR #1208, opened 2023-06-22
Add UnfurlLinks and UnfurlMedia to WebhookMessage

lheiskan:
* to support 'classic unfurl' configuration options in webhook messages.
* https://api.slack.com/reference/messaging/link-unfurling#classic_unfurl
* addresses https://github.com/slack-go/slack/issues/1207


lorenzoaiello (approved):
Thank you for your contribution @lheiskan  !

This is a backwards-compatible change that exposed more control in the unfurling behavior when posting to Slack and is documented in the published Messaging Reference for [link-unfurling](https://api.slack.com/reference/messaging/link-unfurling). Test changes are not necessary as the defaults result in no behavioral changes and the current test coverage does not coverage contractual variances.

PR #1220, opened 2023-08-04
Add support for manifests and config tokens

### LINKED ISSUE #1084: Create and configure apps with manifests
### Description
Use manifest's system to create, configure and validate your Slack apps.

### Slack's documentation
- https://api.slack.com/reference/manifests
- https://api.slack.com/methods/apps.manifest.create
- https://api.slack.com/methods/apps.manifest.update
- https://api.slack.com/methods/apps.manifest.delete
- https://api.slack.com/methods/apps.manifest.export
- https://api.slack.com/methods/apps.manifest.validate

stijndcl:
- Add support for `tooling.tokens.rotate` endpoint, and store `configToken` and `configRefreshToken` in the `Client` alongside the other token types
- Add support for manifest endpoints - fixes #1084 

Considering that Slack allows uploading manifests as both JSON and YAML, but the API docs specify that the manifest submitted to this endpoint should be JSON, I added `yaml:` tags to the `Manifest` struct to allow users to marshal their existing YAML manifests.

### DIFF
--- a/examples/manifests/README.md
+++ b/examples/manifests/README.md
@@ -1,0 +1,38 @@
+# Manifest examples
+
+This example tries to show how to interact with the
+new [manifest endpoints](https://api.slack.com/reference/manifests#manifest_apis). These endpoints require a special set
+of tokens: **configuration tokens**. Refer to
+the [relevant documentation](https://api.slack.com/authentication/config-tokens) for how to create these tokens.
+
+For examples on how to use configuration tokens, see the [tokens example](../tokens).
+
+## Usage info
+
+The manifest endpoints allow you to configure your application programmatically instead of manually creating
+a `manifest.yaml` file and uploading it on your Slack application's dashboard.
+
+A manifest should follow a specific structure and has a handful of required fields. These are describe in
+the [manifest documentation](https://api.slack.com/reference/manifests#fields), but Slack additionally returns very
+informative error messages for malformed templates to help you pin down what the issue is. The library itself does not
+attempt to perform any form of validation on your manifest.
+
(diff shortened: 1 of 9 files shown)


parsley42:
Hi @stijndcl , this looks interesting, can you tell me more about how this will be used, and possibly add a subdirectory with a short README to examples/?

stijndcl:
> Hi @stijndcl , this looks interesting, can you tell me more about how this will be used, and possibly add a subdirectory with a short README to examples/?

@parsley42 I can add examples, yes. Will look into it asap. The main goal of this is to integrate the new endpoints from Slack, because I feel an SDK should cover all available endpoints.

In terms of use cases, you can use these to dynamically configure your Slack app. The current way is manually creating a `manifest` file and uploading it on the dashboard. These endpoints let you do that programmatically.

If you want more information about how these work or what they are for, the official Slack documentation is at your service: https://api.slack.com/reference/manifests#manifest_apis

parsley42:
@stijndcl Ah, yes, that would be great! I'm the author of [gopherbot](https://github.com/lnxjedi/gopherbot), and setting up a robot means editing and uploading a manifest - so it might be a nice feature for new users if this could be automated. However, I do this via the web ui - so, there must be another step for creating credentials that can use this endpoint, which would be helpful to have a README and example for. Thanks!

stijndcl:
@parsley42 there are now two examples. The manifest endpoints are really trivial so it didn't make sense to me to include an example for every single method. One should suffice.

I added necessary information about how to handle configuration tokens as well.

parsley42 (approved):
Thanks for the examples. LGTM.

PR #1224, opened 2023-08-14
add `omitempty` to ConfirmationBlockObject.Deny

kamil-gwozdz:
This PR adds `omitempty` to `ConfirmationBlockObject.Deny`.

It's not a required field and without `omitempty` a `null` ends up being generated in the json payload, which is not accepted by the schema. This means that, with the current version of the client, it's not possible to omit the deny button.


parsley42 (approved):
LGTM

PR #1225, opened 2023-09-10
Add is_enterprise_install flag for org wide installation support

hussachai:
Introducing the inclusion of the 'is_enterprise_install' parameter in commands, interactions, and oauth2 responses. This flag is essential for facilitating organization-wide installations within the enterprise grid environment. Without this flag, the authentication token mapping process becomes unreliable, as it necessitates attempts with both the enterprise ID and team ID, potentially resulting in unexpected outcomes.

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/interactions.go
+++ b/interactions.go
@@ -33,29 +33,30 @@ const (
 
 // InteractionCallback is sent from slack when a user interactions with a button or dialog.
 type InteractionCallback struct {
-	Type            InteractionType         `json:"type"`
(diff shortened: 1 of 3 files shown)


parsley42 (approved):
LGTM

parsley42:
@lololozhkin Any objections? Care to add a review?

parsley42:
Sorry slow to get this in; appears to be more complete than #1255

Issue #967, opened 2021-08-22
Support the new OpenID endpoints


abatilo:
### Description
Slack has added new endpoints such as `/openid.connect.token` for doing an OpenID based sign in flow. It'd be nice to have these endpoints also be in this library. Thanks!

### (Optional) Slack's documentation
https://api.slack.com/authentication/sign-in-with-slack
https://api.slack.com/methods/openid.connect.token

PR #1242, opened 2023-11-06
Openid connect

danielmichaels:
This PR adds support for [Sign in with Slack](https://api.slack.com/authentication/sign-in-with-slack), in particular the [openid.connect.token](https://api.slack.com/methods/openid.connect.token) endpoint.

I've tried to keep consistency with the other methods found in `oauth.go` by providing context and non-context aware functions. 

Two functions were also incorrectly commented which have also been addressed in this PR.

`make pr-prep` passed.

Related: #967


parsley42 (approved):
Appears useful / not harmful

PR #1254, opened 2024-01-24
Add multiline plain text input

obed-sj:
Adding missing arguments on `NewPlainTextInputBlockElement` for Multiline and min_length


obed-sj:
Did not mean to create a PR here

PR #1259, opened 2024-02-07
Fix: always add context to errors during block serde

yunginnanet:
Maintains consistency with existing code:

https://github.com/slack-go/slack/blob/86cd1bdb0575308dd7cc8345cf15e7b724bfafbf/block_conv.go#L213-L215

Adds critical context to errors during the many ~headaches~ joys of developing with the slack block kit.

### DIFF
--- a/block_conv.go
+++ b/block_conv.go
@@ -2,7 +2,6 @@ package slack
 
 import (
 	"encoding/json"
-	"errors"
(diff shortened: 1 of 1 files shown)


parsley42 (approved):
Looks simple/useful enough.

PR #1264, opened 2024-02-16
Fix: parse is_enterprise_install string

yunginnanet:
`Cause: (*fmt.wrapError)(0xc004b8f3c0)(parsing slash command: json: cannot unmarshal string into Go struct field SlashCommand.is_enterprise_install of type bool),`

sometimes we get `"is_enterprise_install":"false"`, sometimes we get `"is_enterprise_install": false`

add `,string`: and now both work


yunginnanet:
errr, actually it won't be that simple

i'll look at this later - needs an issue - can't do it rn

PR #1266, opened 2024-02-20
Parse string or boolean for SlashCommand.IsEnterpriseInstall

### LINKED ISSUE #1265: Release 0.21.4 breaks slash commands in socket mode
### What happened

As of version 0.21.4, Slash commands do not work correctly in socket mode (resulting in "dispatch failed" in the client). This is a regression since version 0.21.3.

### Expected behavior

Slash commands should work correctly in all versions.

### Steps to reproduce

When using Socket mode on a non-enterprise instance, this is the payload when typing `/mycommand foobar` (with lots of redactions):

```json
{
    "envelope_id": "e16988f4-ac45-4b4a-93b4-1851a6f77e6d",
    "payload": {
        "token": "XXX",
        "team_id": "T0000000000",
        "team_domain": "my-workspace",
        "channel_id": "C000000000000",
        "channel_name": "my-channel",
        "user_id": "U0000000000",
        "user_name": "My name here",
        "command": "/mycommand",
        "text": "foobar",
        "api_app_id": "A000000000",
        "is_enterprise_install": "false",
        "response_url": "https://hooks.slack.com/commands/A000000000/00000/XXXXX",
        "trigger_id": "00000.00000.00000"
    },
    "type": "slash_commands",
    "accepts_response_payload": true
}
```

The problem here is that https://github.com/slack-go/slack/pull/1255 adds a new `IsEnterpriseInstall` field with type boolean, whereas in the payload this is a string (e.g. `"false"`). Therefore the parsing of the JSON fails and the slash command handlers don't ever get invoked.

### Versions
- Go: 1.22.0
- slack-go/slack: 0.21.4

kpaulisse:
This PR adds a custom JSON parser for the `SlashCommand` struct to work correctly when `"is_enterprise_install"` comes in as either a string or a boolean. As noted in the issue, in socket mode, certain (but not all 🤷) payloads have `"is_enterprise_install"` as a quoted boolean (e.g. `"false"`) whereas others have it as an actual boolean (e.g. `false`).

This doesn't update the `IsEnterpriseInstall` field in `SlashCommand` itself, so this change should be transparent to anyone who might have started using that newly added field.

Fixes https://github.com/slack-go/slack/issues/1265
Prior fix attempt: https://github.com/slack-go/slack/pull/1264

Testing:
1. Added unit test covering happy paths and error cases
2. My application works correctly pointing at this branch

### DIFF
--- a/slash.go
+++ b/slash.go
@@ -1,7 +1,10 @@
 package slack
 
 import (
+	"encoding/json"
+	"fmt"
 	"net/http"
+	"strconv"
 )
 
 // SlashCommand contains information about a request of the slash command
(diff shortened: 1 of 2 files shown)


parsley42 (approved):
Thanks for writing the tests, LGTM.

PR #1273, opened 2024-03-15
Block Object: Validate text min and max length

Chagui-:
The validate block wasn't validating the text size. This was generating a confusing `invalid_block` error when we were sending a message that was longer than the allowed max value, even though we were calling validate() before sending.

Adds:
- Text field min size validation of 1
- Text field max size validation of 3000
- Tests for both cases


parsley42 (approved):
I don't see any issue.

PR #1275, opened 2024-04-10
expose slackevents.ReactionEvent

kjelle:
Hello.

This PR is related to https://github.com/slack-go/slack/pull/1095 and https://github.com/matthew-graves/slack/commit/a293ed46be13cdf1972630d22de6c44bfd991eb8. 

Please accept this PR to expose `slackevents.ReactionEvent`

### DIFF
--- a/slackevents/inner_events.go
+++ b/slackevents/inner_events.go
@@ -312,7 +312,7 @@ type pinEvent struct {
 	HasPins        bool   `json:"has_pins,omitempty"`
 }
 
-type reactionEvent struct {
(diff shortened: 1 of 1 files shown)


lorenzoaiello (changes requested):
Thanks for the help @kjelle , can you please add a corresponding test for each event type we are now exposing (`ReactionAddedEvent` and ReactionRemovedEvent`)? You can find the other event types tested in `inner_events_tests.go` in the same directory as the file you modified.

PR #1278, opened 2024-04-18
Add missing message option for link_names

robbydyer:
This adds a missing optional parameter to the `chat.postMessage` API (https://api.slack.com/methods/chat.postMessage#arg_link_names)


parsley42 (approved):
Looks harmless

PR #1279, opened 2024-04-18
feat: Add Convenience Methods to Block Elements

obed-sj:
Adds some convenience methods to block elements to easily add functionality


lorenzoaiello (approved):
Thanks for the contribution @obed-sj !

This is a pretty straightforward addition that is aligned to how the other block components have helper methods for manipulating the element structs.

PR #1283, opened 2024-04-26
fix!: Expose `team_id` parameter for use with org-wide app

### LINKED ISSUE #1282: bug: Missing `team_id` param
### What happened

Not all API calls [required](https://api.slack.com/enterprise/apps/reference#list) to have `team_id` param actually have it in the SDK.

### Expected behavior

All API calls [required](https://api.slack.com/enterprise/apps/reference#list) to have `team_id` param actually have it in the SDK.

### Steps to reproduce

`files.list` API (as `Client.GetFilesContext`) as an example

### Versions
- slack-go/slack: `v0.12.5`

candiduslynx:
Fixes https://github.com/slack-go/slack/issues/1282

### DIFF
--- a/bots.go
+++ b/bots.go
@@ -35,19 +35,28 @@ func (api *Client) botRequest(ctx context.Context, path string, values url.Value
 	return response, nil
 }
 
+type GetBotInfoParameters struct {
+	Bot    string
+	TeamID string
+}
+
 // GetBotInfo will retrieve the complete bot information
-func (api *Client) GetBotInfo(bot string) (*Bot, error) {
-	return api.GetBotInfoContext(context.Background(), bot)
+func (api *Client) GetBotInfo(parameters GetBotInfoParameters) (*Bot, error) {
+	return api.GetBotInfoContext(context.Background(), parameters)
 }
 
 // GetBotInfoContext will retrieve the complete bot information using a custom context
-func (api *Client) GetBotInfoContext(ctx context.Context, bot string) (*Bot, error) {
+func (api *Client) GetBotInfoContext(ctx context.Context, parameters GetBotInfoParameters) (*Bot, error) {
 	values := url.Values{
 		"token": {api.token},
 	}
 
-	if bot != "" {
-		values.Add("bot", bot)
+	if parameters.Bot != "" {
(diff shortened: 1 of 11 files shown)


candiduslynx:
Hi @parsley42!

I've noticed that you reviewed several PRs before, so tagging you here.

This PR mainly goes through the list mentioned in the issue & adds `team_id` param setting.

parsley42:
Hi @candiduslynx - I'm retiring from being the de facto maintainer of this project; slack-go/slack needs maintainers, please email me at the address on my GitHub page if interested, thanks.

candiduslynx (commented):
@parsley42 the tests have passed, so I'd like another set of eyes to merge this

bots.go
```diff
@@ -35,19 +35,28 @@ func (api *Client) botRequest(ctx context.Context, path string, values url.Value
...
 
+type GetBotInfoParameters struct {
+	Bot    string
+	TeamID string
+}
+
 // GetBotInfo will retrieve the complete bot information
-func (api *Client) GetBotInfo(bot string) (*Bot, error) {
-	return api.GetBotInfoContext(context.Background(), bot)
+func (api *Client) GetBotInfo(parameters GetBotInfoParameters) (*Bot, error) {
```
parsley42, line 44: The only thing I want to point out is that this is a breaking change in a library. I can't think of a better way to do this, so it's worth thinking about updating the version.

parsley42 (commented):
I'd be interested in your thinking on updating the version, but otherwise this looks reasonable to me.

bots.go
candiduslynx: I think it's fine to do a breaking change while the version is `v0.x.y`.
Once the library hits `v1` that'll be another thing.

Additionally, I'm thinking of overhauling the params as the inconsistent use or url params bugs be a bit (& that'll also be breaking, so I'd reserve a version bump for that).

If that's the only thing could this PR be merged & released?

parsley42: Everything else looks fine to me, sure.

candiduslynx: I don't have the merge rights, though

parsley42 (approved):
LGTM

PR #1286, opened 2024-04-30
chore: enable to upload file without channel param

nakamasato:
##### Pull Request Guidelines

##### PR preparation

##### API changes

###### Ref
- https://api.slack.com/methods/files.completeUploadExternal

### DIFF
--- a/files.go
+++ b/files.go
@@ -517,7 +517,9 @@ func (api *Client) completeUploadExternal(ctx context.Context, fileID string, pa
 	values := url.Values{
 		"token":      {api.token},
 		"files":      {string(requestBytes)},
-		"channel_id": {params.channel},
+	}
+	if params.channel != "" {
+		values.Add("channels", params.channel)
 	}
 
 	if params.initialComment != "" {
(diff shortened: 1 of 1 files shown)


0xdeafcafe:
Anything I can do to help push this through? Would be really useful for us.

nakamasato:
@0xdeafcafe Thanks!I wanted to upload a file without sending out to the channel but after roughly checking the API, i'm not very sure if it's possible from the api perspective and kept my pr draft. I saw you already made your own PR. Let's see there!

0xdeafcafe:
> @0xdeafcafe Thanks!I wanted to upload a file without sending out to the channel but after roughly checking the API, i'm not very sure if it's possible from the api perspective and kept my pr draft. I saw you already made your own PR. Let's see there!

We're using a fork of the repo with my PR on production at the moment, so I can confirm it works thankfully.

lorenzoaiello:
Thanks for helping get the ball started @nakamasato - I'm closing this in favor of #1293

PR #1291, opened 2024-05-24
Fix UploadFileV2 when using text content

### LINKED ISSUE #1290: UploadFileV2 uploads incorrectly when using text `Content` parameter
### What happened

When calling `UploadFileV2()` with a string passed to the `Content` parameter, the file is incorrectly uploaded. This seems to be because old `files.upload` endpoint allowed text content to be passed as a POST parameter, but the new upload URL interprets the unexpected `content=` parameter as the start of the file content, and doesn't correctly URL-decode it.

Screenshot 2024-05-23 at 8 23 05 PM

### Expected behavior
Screenshot 2024-05-23 at 8 25 09 PM

#### reproducible code
```
package main

import (
	"os"

	"github.com/slack-go/slack"
)

func main() {
	token := os.Getenv("SLACK_AUTH_TOKEN")
	channelId := os.Getenv("SLACK_CHANNEL_ID")
	client := slack.New(token, slack.OptionDebug(true))

	fileContent := "This is test file content"
	fileSize := len(fileContent)

	fileParams := slack.UploadFileV2Parameters{
		Channel:        channelId,
		Content:        fileContent,
		Filename:       "file.txt",
		FileSize:       fileSize,
		InitialComment: "Test file",
		Title:          "Test file",
	}
	client.UploadFileV2(fileParams)
}
```

This seems to only be an issue with the `Content` parameter, as converting the string to a `Reader` fixes the problem:

```
reader := strings.NewReader(fileContent)

fileParams := slack.UploadFileV2Parameters{
	Channel:        channelId,
	Reader:         reader,
	Filename:       "file.txt",
	FileSize:       fileSize,
	InitialComment: "Test file",
	Title:          "Test file",
}
client.UploadFileV2(fileParams)
```

### Versions
- Go: `go version go1.20.7 darwin/arm64`
- slack-go/slack: `v0.13.0`

calebmckay:
Fixes #1290.

`postWithMultipartResponse` works better for uploading this way, per the [Slack docs](https://api.slack.com/messaging/files#uploading_files).

> POST the contents of your file to the URL returned in step 1. This can be done by sending the raw bytes, or **_can be a multipart form ended request_**. If all went well, Slack will respond with an HTTP 200.

### DIFF
--- a/files.go
+++ b/files.go
@@ -504,9 +504,8 @@ func (api *Client) getUploadURLExternal(ctx context.Context, params getUploadURL
 func (api *Client) uploadToURL(ctx context.Context, params uploadToURLParameters) (err error) {
 	values := url.Values{}
 	if params.Content != "" {
-		values.Add("content", params.Content)
-		values.Add("token", api.token)
-		err = postForm(ctx, api.httpclient, params.UploadURL, values, nil, api)
+		contentReader := strings.NewReader(params.Content)
+		err = postWithMultipartResponse(ctx, api.httpclient, params.UploadURL, params.Filename, "file", api.token, values, contentReader, nil, api)
 	} else if params.File != "" {
 		err = postLocalWithMultipartResponse(ctx, api.httpclient, params.UploadURL, params.File, "file", api.token, values, nil, api)
 	} else if params.Reader != nil {


candiduslynx (approved):

kalavt:
@calebmckay can we merge?

PR #1293, opened 2024-06-13
chore: make channel optional on `UploadFileV2`

0xdeafcafe:
##### Pull Request Guidelines

##### PR preparation

##### API changes

This change is non-breaking, and will allow users to upload files using the new `v2` endpoint that don't have channel id, so they won't automatically be posted. This is supported by Slack and needed if users want to attach the files to a post vs have them posted in the same thread/channel later.

None.

### DIFF
--- a/files.go
+++ b/files.go
@@ -522,11 +522,13 @@ func (api *Client) completeUploadExternal(ctx context.Context, fileID string, pa
 		return nil, err
 	}
 	values := url.Values{
-		"token":      {api.token},
-		"files":      {string(requestBytes)},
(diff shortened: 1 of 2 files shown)


lorenzoaiello (approved):
Thank you for your contribution @0xdeafcafe ! 

This is a backwards-compatible change that provides more flexibility in uploading files to Slack and is aligned with the published API Specification for [files.upload](https://api.slack.com/methods/files.upload) which also specifies `channel_id` to be an optional argument. Test coverage is sufficient to address the minor change implemented.

PR #1296, opened 2024-07-10
Fix UploadFileV2 to return File instead of FileSummary

vaijab:
By returning `FileSummary` instead of a `File` object, in most cases, we have to make another call to get file info, which requires `files:read` permission scope. This in fact breaks many existing apps.

This change makes the Go sdk somewhat more in-line with other SDKs.

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/files.go
+++ b/files.go
@@ -197,7 +197,7 @@ type completeUploadExternalParameters struct {
 
 type completeUploadExternalResponse struct {
 	SlackResponse
-	Files []FileSummary `json:"files"`
+	Files []File `json:"files"`
 }
 
 type fileResponseFull struct {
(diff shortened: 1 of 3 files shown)


lorenzoaiello:
Thanks for opening a pull request @vaijab 

I don't conceptually have an issue with returning more information from the `UploadFileV2` function if it was available, but the `FileSummary` struct accurately represents the response from the [files.completeUploadExternal](https://api.slack.com/methods/files.completeUploadExternal) API method which is the last call made in the [new upload flow](https://api.slack.com/messaging/files#uploading_files).

The other API Method called in the same `UploadFileV2` function ([files.getUploadURLExternal](https://api.slack.com/methods/files.getUploadURLExternal)) also doesn't provide any unique information worth retaining.

The Slack documentation for the `file.getUploadURLExternal` has an information banner in the `Usage info` section that states:

> If you have migrated to the v2 method, note that this method requires both files:write and files:read scopes. If your existing apps have only the files:write scope for uploading files, you must add files:read, then re-install the app to issue an updated token. Refer to [files.upload v2 in WebClient/AsyncWebClient](https://github.com/slackapi/python-slack-sdk/releases/tag/v3.19.0) for more information.

This was subsequently retracted in [v3.23.0](https://github.com/slackapi/python-slack-sdk/releases/tag/v3.23.0) but only because [the internal calls to `files.info` were eliminated](https://github.com/slackapi/python-slack-sdk/pull/1408) meaning that `files:read` was no longer required to use the function in the SDK instead requiring it to be explicitly called but not fundamentally changing how the Slack API behaves in this respect.

Unfortunately, I don't think it's going to be possible to circumvent the need for the `files:read` scope to get the information you are after (without knowing exactly what you are after, but knowing you want more than is currently returned).

PR #1307, opened 2024-07-19
feat(slackevents): Support metadata in MessageEvent

rfratto:
Support for receiving metadata when reading conversation history was first added in PR #1083. This commit extends metadata support to message events when connected to the Slack WebSocket API.

This aligns with the structure mentioned on Slack's documentation for [Using message metadata](https://api.slack.com/metadata/using#events), and my local testing showed that the metadata does indeed appear properly with this field added.


lorenzoaiello (approved):
Verified supporting documentation for validation. The code change itself is also very minor and is backwards compatible.

PR #1309, opened 2024-08-09
Support publishing a messge to a specific thread

Nikolo:
Support publishing a messge to a specific thread https://api.slack.com/interactivity/handling#publishing_in_thread

From slack interactivity documentation:
Publishing responses in thread 
If you want to publish a message to a specific thread, you'll need to include an attribute response_type and set its value to in_channel. Then, to specify the thread, include a thread_ts.

Also, be sure to set replace_original to false or you'll overwrite the message you're wanting to respond to!


lorenzoaiello (approved):
Apologies for the delay, and thank you for your contribution.

This is a non-breaking change that conforms to Slack's public documentation.

PR #1310, opened 2024-08-10
chore: replace ioutil with io or os package

nakamasato:
## why

> Deprecated: As of Go 1.16, the same functionality is now provided by package [io](https://pkg.go.dev/io) or package [os](https://pkg.go.dev/os), and those implementations should be preferred in new code.

Ref: https://pkg.go.dev/io/ioutil

## what

- `io/ioutil` -> `io`
	- `io.ReadAll`
	- `io.NopCloser`
- `io/ioutil` -> `os`
	- `os.CreateTemp`

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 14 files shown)


lorenzoaiello (approved):
Thanks for the contribution @nakamasato !

PR #1316, opened 2024-08-21
docs: fix deprecated comment for UploadFile and UploadFileContext

shogo82148:
The [UploadFile](https://pkg.go.dev/github.com/slack-go/slack#Client.UploadFile) and [UploadFileContext](https://pkg.go.dev/github.com/slack-go/slack#Client.UploadFileContext) have been deprecated, but pkgsite is not aware of it:

image

This pull request fixes it.
pkgsite will mark UploadFile and UploadFileContext as DEPRECATED:

image

And some tools will warn use of UploadFile and UploadFileContext.

See https://go.dev/wiki/Deprecated for more detail.

### DIFF
--- a/files.go
+++ b/files.go
@@ -355,14 +355,16 @@ func (api *Client) ListFilesContext(ctx context.Context, params ListFilesParamet
 }
 
 // UploadFile uploads a file.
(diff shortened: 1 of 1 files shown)


lorenzoaiello (approved):
Thank you for catching and fixing it! 🙇

PR #1317, opened 2024-08-26
fix: Add required `format` field to rich text date blocks

luke-josh:
As per [block kit docs](https://api.slack.com/reference/block-kit/blocks#date-element-type), the date element requires a format string to be included. Currently, submitting a block kit with this element results in a slack API error.

Also added the two optional fields `url` and `fallback` for posterity.

##### PR preparation

##### API changes


lorenzoaiello (approved):
Apologies for the delay, and thank you for your contribution!

This pull request fixes rich text date blocks by including the `format` field as documented in the public block kit documentation. Additionally, it adds the ability to set the `url` and `fallback` fields which are also publicly documented.

Note: This does include breaking changes to the `NewRichTextSectionDateElement` function.

PR #1319, opened 2024-09-09
feat: Add support for unicode parameter in emoji type of rich text blocks

YutoKashiwagi:
This PR adds support for the `unicode` parameter to the `RichTextSectionEmojiElement` struct for rich text blocks. While this parameter is not officially documented in Slack's API, it is present in the JSON payload of actual Slack messages and represents the Unicode code point of the emoji.
https://api.slack.com/reference/block-kit/blocks#emoji-element-type

For example, a rich text block with an emoji can include the unicode field like this:

```json
"blocks": [
    {
      "type": "rich_text",
      "block_id": "xxxxx",
      "elements": [
        {
          "type": "rich_text_section",
          "elements": [
            {
              "type": "emoji",
              "name": "+1",
              "unicode": "1f44d"
            }
          ]
        }
      ]
    }
  ]
```

The unicode parameter behaves similarly to the skin-tone parameter, which is also undocumented but has already been included in the structure. This PR aligns the handling of unicode in the same way to ensure emojis are fully supported in Slack message payloads.

Please review, and feel free to provide feedback if any adjustments are needed. Thank you!

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/block_rich_text.go
+++ b/block_rich_text.go
@@ -341,6 +341,7 @@ type RichTextSectionEmojiElement struct {
 	Type     RichTextSectionElementType `json:"type"`
 	Name     string                     `json:"name"`
 	SkinTone int                        `json:"skin_tone"`
+	Unicode  string                     `json:"unicode,omitempty"`
 	Style    *RichTextSectionTextStyle  `json:"style,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


lorenzoaiello (approved):
Thank you for the contribution @YutoKashiwagi !

The Slack documentation now also reflects the `unicode` field so this is no longer un-documented.

Issue #1276, opened 2024-04-11
Not possible to set initial value of RichTextInputBlockElement due to wrong type


matcornic:
Hello maintainers,

First, thanks a lot for this lib, it's very useful for us.
I encountered a bug: I can't set an initial value for the  RichTextInputBlockElement due the type being a string and not a Rich text object

### What happened

I tried to use this code, which is compiling on :

```
	elem := slack.NewRichTextInputBlockElement(nil, "test")
	elem.InitialValue = "test"
	elemInput := slack.NewInputBlock("", slack.NewTextBlockObject("plain_text", "Test:", false, false), nil, elemInput)
```

But got an `invalid_block` error on Slack while checking the generated json on the Block kit builder

https://app.slack.com/block-kit-builder/T02F6HSJU2H#%7B%22blocks%22:%5B%7B%22type%22:%22header%22,%22text%22:%7B%22type%22:%22plain_text%22,%22text%22:%22Test%22%7D%7D,%7B%22type%22:%22input%22,%22label%22:%7B%22type%22:%22plain_text%22,%22text%22:%22RulTestes:%22%7D,%22element%22:%7B%22type%22:%22rich_text_input%22,%22action_id%22:%22test%22,%22initial_value%22:%22test%22%7D%7D%5D%7D

![image](https://github.com/slack-go/slack/assets/3520706/8904a99c-c155-4af5-8009-94c79215082c)

Seeing the documentation, it expects a Rich text object https://api.slack.com/reference/block-kit/block-elements#rich_text_input

### Versions
- Go: 1.20
- slack-go/slack: v0.12.5

reaganrazon:
I am also encountering this problem. @matcornic How did you get around it? I need to set an initial value, and in the RichTextInputBlockElement, "InitialValue" is defined as a string, but this will not pass my validation.

Manjish:
I have opened a PR to fix this issue. [PR LINK](https://github.com/slack-go/slack/pull/1320)
For the workaround, I have forked the repo, changed the data type and module then used the new package in the project

lorenzoaiello:
I believe this was fixed in #1320, please follow up if you are still experiencing the issue.

zevzoldan:
I'm able to set an initial value, and while it loads, Slack returns an `internal_error` when using `slack_client.views_publish(user_id=user_id, view={"type": "home", "blocks": blocks})`

```
[...]

PR #1320, opened 2024-09-13
fix: Updated RichTextInputBlockElement InitialValue data type

Manjish:
##### Pull Request Guidelines

##### PR preparation

Fix for [issue #1276](https://github.com/slack-go/slack/issues/1276)
Updated the datatype of RichTextInputBlockElement InitialValue from string to *RichTextBlock

### DIFF
--- a/block_element.go
+++ b/block_element.go
@@ -527,7 +527,7 @@ type RichTextInputBlockElement struct {
 	Type                 MessageElementType    `json:"type"`
 	ActionID             string                `json:"action_id,omitempty"`
 	Placeholder          *TextBlockObject      `json:"placeholder,omitempty"`
-	InitialValue         string                `json:"initial_value,omitempty"`
+	InitialValue         *RichTextBlock                `json:"initial_value,omitempty"`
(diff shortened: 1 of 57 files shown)


lorenzoaiello (changes requested):
@Manjish , thank you for the contribution!

It looks like there are some files that were perhaps mistakenly committed that change a lot of the imports to the forked github paths. Can you please fix those?

Manjish:
@lorenzoaiello 
Thanks for pointing it out. I changed to test in my local.
I have reverted the changes. 
Thank you 🙇

Manjish:
> @Manjish , thank you for the contribution!
> 
> It looks like there are some files that were perhaps mistakenly committed that change a lot of the imports to the forked github paths. Can you please fix those?

@lorenzoaiello 
Thanks for pointing it out. I changed to test in my local.
I have reverted the changes.
Thank you 🙇

lorenzoaiello (approved):
Thanks for a quick follow up!

PR #1322, opened 2024-09-17
Implement PublishViewContextV2 to keep hash optional

ishitamundhra:
##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/views.go
+++ b/views.go
@@ -137,12 +137,6 @@ type openViewRequest struct {
 	View      ModalViewRequest `json:"view"`
 }
 
-type publishViewRequest struct {
-	UserID string             `json:"user_id"`
-	View   HomeTabViewRequest `json:"view"`
-	Hash   *string            `json:"hash,omitempty"`
-}
-
 type pushViewRequest struct {
 	TriggerID string           `json:"trigger_id"`
 	View      ModalViewRequest `json:"view"`
--- a/views.go
+++ b/views.go
@@ -110,6 +110,12 @@ type ModalViewRequest struct {
 	ExternalID      string           `json:"external_id,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


lorenzoaiello:
@ishitamundhra , thanks for opening the PR!

I appreciate you wanting to make this backwards compatible, but I'd like to try and limit additional functions for this and instead bite the bullet once and transition the function to use an input struct instead so that additional parameters can be added in the future without breaking backwards compatibility. How would you feel about going down that path for this PR?

ishitamundhra:
> @ishitamundhra , thanks for opening the PR!
> 
> I appreciate you wanting to make this backwards compatible, but I'd like to try and limit additional functions for this and instead bite the bullet once and transition the function to use an input struct instead so that additional parameters can be added in the future without breaking backwards compatibility. How would you feel about going down that path for this PR?

Hey @lorenzoaiello. Makes complete sense. I've made the changes to the current diff—can you take a look and let me know if it's in line with what you had in mind?

lorenzoaiello:
@ishitamundhra , apologies for any ambiguity in my previous post. Thank you for adding the input struct, I was actually proposing not creating a `V2` function but creating the breaking change on the current function to transition to an input struct. Does that make sense?

ishitamundhra:
> @ishitamundhra , apologies for any ambiguity in my previous post. Thank you for adding the input struct, I was actually proposing not creating a `V2` function but creating the breaking change on the current function to transition to an input struct. Does that make sense?

Hey @lorenzoaiello, thanks for the feedbacak. Was hesistant to make the breaking change in the first place. Have updated the diff based on your comment, let me know if this feels good.

nlopes (approved):

PR #1330, opened 2024-10-06
feat: Add support for external_limited option of inviteShared

bamo:
Expose the ability to override the [external_limited option](https://api.slack.com/methods/conversations.inviteShared#arg_external_limited) for inviteShared.

Adding the param to all the InviteSharedEmailsToConversation, etc. methods would be a breaking change to those callers, so I opted instead to expose the underlying helper (renamed to InviteSharedToConversation).

I feel like the convenience methods (InviteSharedEmailsToConversation/InviteSharedUserIDsToConversation) are not actually that much more convenient than just using the helper, and I think we can eventually remove them in favor of having people call InviteSharedToConversation directly.  But that's a future thing.

Although it's slightly inconvenient for the caller to use *bool for ExternalLimited, the two alternatives I considered are, I think worse:
 - Include ExternalLimited as a bool in the InviteSharedParams.  I dislike this way because it gives the SDK user of InviteSharedToConversation a different default behavior from inviteShared, since the default value in the API is true.
 - Add a bool like NonExternalLimited to InviteSharedParams.  This way the defaulting is consistent with the API if it's not specified; however, the InviteSharedParams no longer mirror the API args, which I think is confusing.

### DIFF
--- a/conversation.go
+++ b/conversation.go
@@ -329,42 +329,67 @@ func (api *Client) InviteUsersToConversationContext(ctx context.Context, channel
 }
 
 // InviteSharedEmailsToConversation invites users to a shared channels by email.
-// For more details, see InviteSharedEmailsToConversationContext documentation.
+// For more details, see InviteSharedToConversationContext documentation.
 func (api *Client) InviteSharedEmailsToConversation(channelID string, emails ...string) (string, bool, error) {
-	return api.inviteSharedToConversationHelper(context.Background(), channelID, emails, nil)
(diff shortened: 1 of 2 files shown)


lorenzoaiello (approved):
Thanks for the contribution and excellent PR description @bamo ! I appreciate you choosing the path of least resistance for existing users.

> I feel like the convenience methods (InviteSharedEmailsToConversation/InviteSharedUserIDsToConversation) are not actually that much more convenient than just using the helper, and I think we can eventually remove them in favor of having people call InviteSharedToConversation directly. But that's a future thing.

I agree.

PR #1341, opened 2024-10-31
fix: Don't require `skin_tone` for rich text emoji element

calebmckay:
In a rich text `emoji` element, the `skin_tone` value is optional, and when not provided will use the default skin tone. When unmarshalling from JSON, this parameter defaults to an invalid value of 0 (`skin_tone` values range from 2 to 6), resulting in an "invalid_blocks" error when sending a message. This PR allows the `skin_tone` element to be omitted when unmarshalling from JSON.


calebmckay:
Working on adding test cases now, to make sure both marshalling and unmarshalling work as expected.

lorenzoaiello (approved):
Thanks for the contribution, this looks good - and I appreciate the backwards compatibility.

PR #1349, opened 2024-11-14
feat: Adding Workflow Trigger WebAPI Methods

lorenzoaiello:
This pull request adds support for Web API Methods related to Workflow Triggers:

- [workflows.triggers.permissions.add](https://api.slack.com/methods/workflows.triggers.permissions.add)
- [workflows.triggers.permissions.list](https://api.slack.com/methods/workflows.triggers.permissions.list)
- [workflows.triggers.permissions.remove](https://api.slack.com/methods/workflows.triggers.permissions.remove)
- [workflows.triggers.permissions.set](https://api.slack.com/methods/workflows.triggers.permissions.set)

It also adds a helper method for posting JSON.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -26,6 +26,12 @@ type SlackResponse struct {
 	ResponseMetadata ResponseMetadata `json:"response_metadata"`
 }
 
+type ResponsePointer struct {
+	Ok               *bool            `json:"ok,omitempty"`
+	Error            *string          `json:"error,omitempty"`
+	ResponseMetadata ResponseMetadata `json:"response_metadata,omitempty"`
+}
+
 func (t SlackResponse) Err() error {
 	if t.Ok {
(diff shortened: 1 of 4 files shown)


misc.go
```diff
@@ -26,6 +26,12 @@ type SlackResponse struct {
 	ResponseMetadata ResponseMetadata `json:"response_metadata"`
 }
 
+type ResponsePointer struct {
```
nlopes: Can we have this PR without this?

I'd love to keep `SlackResponse` for introducing this feature.

I think I understand the rationale for having these as `omitempty` but if we introduce this change, I'd rather do it in its own PR and change the `SlackResponse` instead of a new struct.

What's your take @lorenzoaiello ?

nlopes (changes requested):

Issue #1352, opened 2024-11-22
Parse conversation_host_id to Channel object


fahmizulhasymi:
### Description

To get "host" information from shared channel, so it can be used to separate list of channel accessible by user or bots.

### (Optional) Slack's documentation

The attributes classified as [other-attributes](https://api.slack.com/types/conversation#other-attributes) that indicates the "host" of the shared channel

PR #1353, opened 2024-11-22
chore: Add parse channel context and host for shared channel

fahmizulhasymi:
Resolving issue https://github.com/slack-go/slack/issues/1352 that currently unable to parse the host information on shared channel.

With this addition, user can know the owner of the shared channel.

### DIFF
--- a/conversation_test.go
+++ b/conversation_test.go
@@ -165,6 +165,12 @@ func TestCreateSharedChannel(t *testing.T) {
 	channel.LastRead = "1401383885.000061"
 	channel.UnreadCount = 0
 	channel.UnreadCountDisplay = 0
+	channel.IsShared = true
+	channel.IsExtShared = true
+	channel.ContextTeamID = "T1ABCD2E12"
+	channel.ConversationHostID = "T07XY8FPJ5C"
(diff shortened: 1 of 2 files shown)


conversation_test.go
```diff
@@ -82,6 +82,92 @@ func TestCreateSimpleChannel(t *testing.T) {
...
+	channel.IsChannel = true
+	channel.Created = JSONTime(1360782804)
+	channel.Creator = "U024BE7LH"
+	channel.IsArchived = false
+	channel.IsGeneral = false
+	channel.IsMember = true
+	channel.LastRead = "1401383885.000061"
+	channel.UnreadCount = 0
+	channel.UnreadCountDisplay = 0
+	assertSharedChannel(t, channel)
```
calebmckay, line 174: This test fails. Looks like this was copied/pasted from `TestCreateSimpleChannel` without adding the extra fields for the shared channel.

calebmckay (changes requested):

conversation_test.go
fahmizulhasymi: Thanks, miss that one.

calebmckay (approved):

nlopes (approved):

nlopes:
Thank you!

PR #1354, opened 2024-11-27
fix: Allow `emoji` value in plain text blocks to be null

### LINKED ISSUE #1332: Unable to disable emojis in plain text blocks
### What happened

Creating a `TextBlockObject` with the `Emoji` field set to `false` **does not** disable rendering of emojis in that text block.

The reason this happens is because of the `omitempty` JSON-tag:

```go
type TextBlockObject struct {
	Type     string `json:"type"`
	Text     string `json:"text"`
	Emoji    bool   `json:"emoji,omitempty"` // <-- this one
	Verbatim bool   `json:"verbatim,omitempty"`
}
```

This causes the field to be removed from the payload when encoded to JSON (if the field is `false`) as seen in this snippet: https://go.dev/play/p/63tqgJJ6Nnt.

The required JSON payload for creating a plain text object without support for emojis is:

```json
{
	"type": "section",
	"text": {
		"type": "plain_text",
		"text": ":+1:",
		"emoji": false
	}
}
```

This is likely an issue for some other fields with the `omitempty` tag as well, where you explicitly need to set a value to `false` in the JSON payload for the Slack API.

### Expected behavior

Creating a `TextBlockObject` with the `Emoji` field set to `false` **should** disable rendering of emojis in that text block.

### Steps to reproduce

```go
emoji, verbatim := false, false
slack.New("some token").PostMessage(
	"channel ID",
	slack.MsgOptionBlocks(
		slack.NewSectionBlock(
			slack.NewTextBlockObject("plain_text", ":+1:", emoji, verbatim), 
			nil, 
			nil,
		),
	),
)
```

### Versions
- Go: `go version go1.23.1 linux/amd64`
- slack-go/slack: `v0.14.0`

calebmckay:
Fixes #1332.

This PR allows setting null values for the `emoji` field of plain text blocks, which was previously incorrectly marshalling to JSON due to a "false" value being omitted. I opted not to change the signature of `NewTextBlockObject`, since the null value is only expected when unmarshalling, and the user should know which value they need to use when creating the block.


nlopes:
This will go into a "breaking changes" batch.

nlopes:
@calebmckay I finally took a look at this today and I'm not keen on making people depend on another package to interact with this one.

The best approach would be to make `Emoji bool` become `Emoji *bool` and then we can have:
- `nil` -> omitted
- `false` -> `false`
- `true` -> `true`

And no package introduced. I think this would be preferable. Let me know if this isn't up your alley and I can put a different PR to fix the issue with this approach but also happy to have you change this PR instead.

calebmckay:
Understandable. I worked on this a while back so I don't remember exactly, but I think I opted for the null package because I ran into some issue with the bool pointers. I'm willing to give it another shot though.

calebmckay:
Got it working with `*bool`, so this is ready for re-review.

nlopes (approved):

sheldonhull:
I'm on v0.17.0-rc4 and don't see this change persisted in it, the omitempty is not there.
Is this supposed to be included in the latest release candidate @nlopes? 

https://github.com/slack-go/slack/blob/3353adb92b294c4fdd5a5ff73aac1e9c8d8e9888/block_object.go#L119-L126


correction: I see what you did with the pointer, I guess this broke for me on: 					`slack.NewTextBlockObject(slack.MarkdownType, message, false, false)` as it doesn't except a pointer. It then breaks the message.

https://github.com/slack-go/slack/blob/3353adb92b294c4fdd5a5ff73aac1e9c8d8e9888/block_object.go#L163-L171

So I'm seeing this behavior on all my block oriented messages with slack.NewTextBlock


```go
	helpBlocks := []slack.Block{
		slack.NewSectionBlock(
			slack.NewTextBlockObject(
				slack.MarkdownType,
				myactivitymessage,
				false,
				false,
			),
			nil,
			nil,
		),
	}
```

I don't see how to fix as i have to pass true or false to NewTextBlockObject and can't pass a nil.

Issue #1358, opened 2024-12-05
Add GetBlockID() method to slack.Block interface.


takuo:
### Description

It makes easy to find a block by BlockID from BlockSet since I believe that all  MessageBlockType has BlockID.
I will create PR if the request is acceptable.

### usage example 

```go
	var ic slack.InteractionCallback
	b, _ := io.ReadAll(req.Body)
	json.Unmarshal(b, &ic)

	var block slack.Block
	if i := slices.IndexFunc(ic.View.Blocks.BlockSet, func(blk slack.Block) bool {
		return blk.GetBlockID() == "input-block-age"
	}); i >= 0 {
		block = ic.View.Blocks.BlockSet[i]
	}
```

PR #1359, opened 2024-12-07
Add ID() to the Block interface

samstarling:
At incident.io, we often iterate over the contents of Slack views, and often want to add the block ID to various bits of observability (logs, traces). Doing this is quite difficult, because the `Block` interface doesn't allow easy access to the ID — we have to cast each possible kind of block to pull this out.

This PR introduces an `ID()` function to the `Block` interface, making this kind of thing much easier.

I added some tests along the way, and also noticed that `CallBlock` doesn't allow the ID to be set. This doesn't affect us, but you might wish to fix that.

### DIFF
--- a/block.go
+++ b/block.go
@@ -26,6 +26,7 @@ const (
 // to ensure consistency between blocks.
 type Block interface {
 	BlockType() MessageBlockType
+	ID() string
 }
 
 // Blocks is a convenience struct defined to allow dynamic unmarshalling of
--- a/block_action.go
+++ b/block_action.go
@@ -14,6 +14,11 @@ func (s ActionBlock) BlockType() MessageBlockType {
 	return s.Type
 }
 
+// ID returns the ID of the block
+func (s ActionBlock) ID() string {
+	return s.BlockID
+}
+
 // NewActionBlock returns a new instance of an Action Block
 func NewActionBlock(blockID string, elements ...BlockElement) *ActionBlock {
 	return &ActionBlock{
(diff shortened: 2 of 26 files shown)


calebmckay (approved):
Fixes #1358.

This definitely seems like a good feature to have when parsing a BlockSet from a JSON payload. I'm not sure how often we'd want to set a BlockID manually from the `New*Block` functions, but I guess the main problem is that it's not currently consistent across block types.

samstarling:
@calebmckay Nice, thanks! I hadn't seen that issue. Can I ask what the path to getting this reviewed/approved is? Just want to make sure I've not missed anything. But yes, you're right — it's useful to (for example) get a payload from Slack and be able to inspect the block IDs without having to do a bunch of type conversion.

calebmckay:
It looks like @lorenzoaiello is the primary active maintainer right now, so they'll need to review and merge.

samstarling:
Hey @parsley42, I was wondering, are you the current maintainer here?

parsley42:
No, you want @lorenzoaiello

nlopes (approved):

nlopes:
Although technically this can be a breaking change, I think people should be relying on what we provide so am going to consider this as part of a non-breaking change batch.

samstarling:
@nlopes I've just pushed a fix for the linting issues, thanks for flagging!

jamesjarvis:
It appears this has been a bit of a breaking change 😢  the [TextBlockObject](https://github.com/slack-go/slack/blob/965251aa80cbcdae9e44c96b80ef04bae2f3c515/block_object.go#L122) cannot have a block ID (as slack does not provide one), yet is used in this context for looping over blocks :/

PR #1364, opened 2025-01-06
add enable_section option for UserGroup API

### LINKED ISSUE #1363: Support enable section
### Description

Currenty, CreateUserGroup(Cotnext) cannot pass the enable_section argument, so we can't handle it. I want to handle this param.

https://api.slack.com/methods/usergroups.create

sivchari:
close #1363 

##### Pull Request Guidelines

##### PR preparation

- [x] exposes a previously private type, const, method, etc.
- [x] is it application specific (caching, retry logic, rate limiting, etc)
- [x] is it performance related.

##### API changes

### DIFF
--- a/usergroups.go
+++ b/usergroups.go
@@ -3,6 +3,7 @@ package slack
 import (
 	"context"
 	"net/url"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


usergroups.go
```diff
@@ -50,20 +51,45 @@ func (api *Client) userGroupRequest(ctx context.Context, path string, values url
 	return response, response.Err()
 }
 
+// CreateUserGroupParams contains arguments for CreateUserGroup method call
+type CreateUserGroupParams struct {
```
nlopes: Can we make this private and same with `EnableSection` to keep it consistent with the pattern we already have? 

See [chat.go](https://github.com/slack-go/slack/blob/34f9a10b71d70bdcd1b9456a6d91805aca8000bf/chat.go#L398).

nlopes (changes requested):

usergroups.go
sivchari: done.

nlopes (approved):

PR #1370, opened 2025-01-28
Additional field to the File structure

rntk:
Slack has a feature called 'Send emails to Slack' - https://slack.com/intl/en-gb/help/articles/206819278-Send-emails-to-Slack

When emails are sent to this channel, they become Slack files with additional fields containing sender and receiver information. 

I added two new fields to properly support the unmarshaling of file metadata.


nlopes (approved):

nlopes:
Thank you!

PR #1374, opened 2025-02-14
Configure Dependabot to maintain GitHub Actions

sgerrand:
💁 When browsing this repository earlier, I saw that external actions are referenced using commit SHAs. This is useful as a way of ensuring that the version of a specific workflow doesn't drift unexpectedly but it also means that they need to be maintained (whether actively or not).

This change configures @dependabot to do the hard work of looking after those external action references and updating them as needed.


sgerrand:
👋 @nlopes, requesting your 👀 on this small change.

nlopes (approved):

sgerrand:
Thanks! I can't merge this so feel free to do so.

Issue #1346, opened 2024-11-08
Expose the file upload endpoints instead of bundling it into one uploadFileV2 function


assembly-deepali:
### Description
There are cases where I want to upload the files to slack, and not immediately send it. My example is I want to send a message with the correct attachments within the message a user specifies (and not two separate messages). I want two endpoints support: specifically I want to be able to upload them to slack as an external url, and create my own file block to send it within a channel.

sheldonhull:
I'm guessing this is what I need as well to be able to upload the file and then post with attachments like action buttons. I was trying to find out how to do this, but I think the current approach doesn't expose this to me to work with in a separate post.

Issue #1288, opened 2024-05-06
Upload Files V2: Ability to upload multiple files in one message


rbreeze:
### Description
With V2 File uploading, Slack added the ability to upload multiple files to one message to its Python SDK. Would be great to have it here.

### Slack's documentation
https://github.com/slackapi/python-slack-sdk/releases/tag/v3.19.0

parsley42:
Hi @rbreeze  - I'm retiring from being the de facto maintainer of this project; slack-go/slack needs maintainers, please email me at the address on my GitHub page if interested, thanks.

nlopes:
This should be fixed now. Let me know if you think it isn't.

PR #1376, opened 2025-02-19
Expose private File Upload V2 methods to support multiple file uploads in a single message

YutoKashiwagi:
Hi, thank you for maintaining slack-go!

Currently, `UploadFileV2Context` works by sequentially calling three private methods in the following flow:

1. getUploadURLExternal (retrieves the upload URL)
2. uploadToURL (uploads the file to that URL)
3. completeUploadExternal (finalizes the upload)

Because these methods are private, developers can only upload one file per message via `UploadFileV2Context`. However, the Slack API itself supports uploading multiple files in a single message by running these steps for each file and then calling completeUploadURLExternal for all file IDs together.

To unlock this functionality while still using slack-go, I’ve made the following changes in this PR:

- Made `getUploadURLExternal` and `completeUploadExternal` public
  - This allows developers to orchestrate multiple file uploads (Steps 1–3 repeated for each file) and finalize them all in a single message.
- `uploadToURL` remains private for now. This method is simply a regular POST request handler, not part of Slack’s official API. Since making getUploadURLExternal and completeUploadExternal public already addresses the main use case of uploading multiple files, this might be sufficient.
  - However, if you feel uploadToURL should also be made public for consistency, please let me know and I can include that change in this PR.
  

Thank you for considering this proposal. Being able to handle multi-file uploads in a single Slack message through slack-go would be extremely helpful. If you have any suggestions or an alternative approach, I’d love to hear them!

# Related Issues

- https://github.com/slack-go/slack/issues/1346
- https://github.com/slack-go/slack/issues/1288

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/files.go
+++ b/files.go
@@ -510,10 +510,10 @@ func (api *Client) ShareFilePublicURLContext(ctx context.Context, fileID string)
 // Slack API docs: https://api.slack.com/methods/files.getUploadURLExternal
 func (api *Client) GetUploadURLExternalContext(ctx context.Context, params GetUploadURLExternalParameters) (*GetUploadURLExternalResponse, error) {
 	if params.FileName == "" {
-		return nil, fmt.Errorf("GetUploadURLExternalContext: fileName cannot be empty")
+		return nil, fmt.Errorf("files.getUploadURLExternal: fileName cannot be empty")
 	}
 	if params.FileSize == 0 {
-		return nil, fmt.Errorf("GetUploadURLExternalContext: fileSize cannot be 0")
+		return nil, fmt.Errorf("files.getUploadURLExternal: fileSize cannot be 0")
 	}
 
 	values := url.Values{
--- a/files.go
+++ b/files.go
@@ -173,14 +173,14 @@ type UploadFileV2Parameters struct {
 	SnippetText     string
 }
 
-type getUploadURLExternalParameters struct {
-	altText     string
-	fileSize    int
-	fileName    string
-	snippetText string
+type GetUploadURLExternalParameters struct {
+	AltText     string
+	FileSize    int
+	FileName    string
+	SnippetText string
 }
 
-type getUploadURLExternalResponse struct {
+type GetUploadURLExternalResponse struct {
 	UploadURL string `json:"upload_url"`
 	FileID    string `json:"file_id"`
 	SlackResponse
@@ -506,20 +506,28 @@ func (api *Client) ShareFilePublicURLContext(ctx context.Context, fileID string)
 	return &response.File, response.Comments, &response.Paging, nil
 }
 
-// getUploadURLExternal gets a URL and fileID from slack which can later be used to upload a file.
-func (api *Client) getUploadURLExternal(ctx context.Context, params getUploadURLExternalParameters) (*getUploadURLExternalResponse, error) {
+// GetUploadURLExternalContext gets a URL and fileID from slack which can later be used to upload a file.
+// Slack API docs: https://api.slack.com/methods/files.getUploadURLExternal
+func (api *Client) GetUploadURLExternalContext(ctx context.Context, params GetUploadURLExternalParameters) (*GetUploadURLExternalResponse, error) {
+	if params.FileName == "" {
+		return nil, fmt.Errorf("GetUploadURLExternalContext: fileName cannot be empty")
+	}
+	if params.FileSize == 0 {
+		return nil, fmt.Errorf("GetUploadURLExternalContext: fileSize cannot be 0")
+	}
+
 	values := url.Values{
 		"token":    {api.token},
-		"filename": {params.fileName},
-		"length":   {strconv.Itoa(params.fileSize)},
+		"filename": {params.FileName},
+		"length":   {strconv.Itoa(params.FileSize)},
 	}
-	if params.altText != "" {
-		values.Add("initial_comment", params.altText)
+	if params.AltText != "" {
+		values.Add("initial_comment", params.AltText)
 	}
-	if params.snippetText != "" {
-		values.Add("thread_ts", params.snippetText)
+	if params.SnippetText != "" {
+		values.Add("thread_ts", params.SnippetText)
 	}
-	response := &getUploadURLExternalResponse{}
+	response := &GetUploadURLExternalResponse{}
 	err := api.postMethod(ctx, "files.getUploadURLExternal", values, response)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


files_test.go
```diff
@@ -282,3 +284,264 @@ func TestUploadFileV2(t *testing.T) {
...
+func (m *mockGetUploadURLExternalHttpClient) Do(req *http.Request) (*http.Response, error) {
+	if req.URL.Path != "files.getUploadURLExternal" {
+		return nil, fmt.Errorf("invalid path: %s", req.URL.Path)
+	}
+
+	return &http.Response{
+		StatusCode: m.ResponseStatus,
+		Body:       io.NopCloser(bytes.NewBuffer(m.ResponseBody)),
+	}, nil
+}
```
YutoKashiwagi, line 302: I considered following the existing once.Do(startServer) approach for testing, but found that simply mocking Client.httpclient would be sufficient. Therefore, I decided to mock it instead.

files.go
```diff
@@ -615,11 +623,14 @@ func (api *Client) UploadFileV2Context(ctx context.Context, params UploadFileV2P
...
-	c, err := api.completeUploadExternal(ctx, u.FileID, completeUploadExternalParameters{
-		title:           params.Title,
-		channel:         params.Channel,
-		initialComment:  params.InitialComment,
-		threadTimestamp: params.ThreadTimestamp,
+	c, err := api.CompleteUploadExternalContext(ctx, CompleteUploadExternalParameters{
+		Files: []FileSummary{{
+			ID:    u.FileID,
+			Title: params.Title,
+		}},
```
YutoKashiwagi, line 630: For CompleteUploadURLExternal, I've not only made it public but also updated its signature so that multiple fileID parameters (with IDs and titles) can be specified, aligning it more closely with Slack’s official API.

sheldonhull:
thanks for doing this. I just was revisiting a way to grab the urls so I could publish a single "gallery style" message and this probably will fix much of that by providing back the upload urls back. cheers 🙇

files.go
```diff
@@ -506,20 +506,28 @@ func (api *Client) ShareFilePublicURLContext(ctx context.Context, fileID string)
 	return &response.File, response.Comments, &response.Paging, nil
 }
 
-// getUploadURLExternal gets a URL and fileID from slack which can later be used to upload a file.
-func (api *Client) getUploadURLExternal(ctx context.Context, params getUploadURLExternalParameters) (*getUploadURLExternalResponse, error) {
+// GetUploadURLExternalContext gets a URL and fileID from slack which can later be used to upload a file.
+// Slack API docs: https://api.slack.com/methods/files.getUploadURLExternal
+func (api *Client) GetUploadURLExternalContext(ctx context.Context, params GetUploadURLExternalParameters) (*GetUploadURLExternalResponse, error) {
+	if params.FileName == "" {
+		return nil, fmt.Errorf("GetUploadURLExternalContext: fileName cannot be empty")
```
nlopes: Do you mind removing `GetUploadURLExternalContext` from the error? Happy for the message to indicate where it's at but this format doesn't quite fit with everything else.

I am aware we already have `file.upload.v2:` in `fmt.Errorf` in this file so happy if you want to change to be sort of that format but what I'd really prefer is to not have the name of the function in the error returned.

nlopes (changes requested):
@YutoKashiwagi thank you so much for sending this and providing a well reasoned approach here.

Also very smart to match with the ability to send multiple files per the Slack API as part of opening this up to users of the library.

There's a tiny change I think is worth doing and once that's out of the way I will ✅ this.

files.go
YutoKashiwagi: Thank you. I’ve updated it to include the API name instead of the function name.

https://github.com/slack-go/slack/pull/1376/commits/9a7fd29f8ccb6698f14e20a2f418d3e27012b439

files.go
```diff
@@ -506,20 +506,28 @@ func (api *Client) ShareFilePublicURLContext(ctx context.Context, fileID string)
 	return &response.File, response.Comments, &response.Paging, nil
 }
 
-// getUploadURLExternal gets a URL and fileID from slack which can later be used to upload a file.
-func (api *Client) getUploadURLExternal(ctx context.Context, params getUploadURLExternalParameters) (*getUploadURLExternalResponse, error) {
+// GetUploadURLExternalContext gets a URL and fileID from slack which can later be used to upload a file.
+// Slack API docs: https://api.slack.com/methods/files.getUploadURLExternal
+func (api *Client) GetUploadURLExternalContext(ctx context.Context, params GetUploadURLExternalParameters) (*GetUploadURLExternalResponse, error) {
+	if params.FileName == "" {
+		return nil, fmt.Errorf("files.getUploadURLExternal: fileName cannot be empty")
```
nlopes: I think because these are called before the API call, we can just state the name of the param that is required.

```suggestion
		return nil, fmt.Errorf("FileName cannot be empty")
```

files.go
```diff
@@ -506,20 +506,28 @@ func (api *Client) ShareFilePublicURLContext(ctx context.Context, fileID string)
...
+	}
+	if params.FileSize == 0 {
+		return nil, fmt.Errorf("files.getUploadURLExternal: fileSize cannot be 0")
```
nlopes:
```suggestion
		return nil, fmt.Errorf("FileSize cannot be 0")
```

nlopes (changes requested):

nlopes (approved):

files_test.go
```diff
@@ -282,3 +284,264 @@ func TestUploadFileV2(t *testing.T) {
...
+			wantSlackResponse: []byte(`{"ok":false,"error":"errored"}`),
+			wantErr:           fmt.Errorf("errored"),
+		},
+		{
+			title: "Testing with invalid parameters: empty file name",
+			params: GetUploadURLExternalParameters{
+				FileName: "",
+				FileSize: 10,
+			},
+			wantErr: fmt.Errorf("files.getUploadURLExternal: fileName cannot be empty"),
```
nlopes:
```suggestion
			wantErr: fmt.Errorf("FileName cannot be empty"),
```

files_test.go
```diff
@@ -282,3 +284,264 @@ func TestUploadFileV2(t *testing.T) {
...
+		},
+		{
+			title: "Testing with invalid parameters: file size 0",
+			params: GetUploadURLExternalParameters{
+				FileName: "test.txt",
+				FileSize: 0,
+			},
+			wantErr: fmt.Errorf("files.getUploadURLExternal: fileSize cannot be 0"),
```
nlopes:
```suggestion
			wantErr: fmt.Errorf("FileSize cannot be 0"),
```

nlopes (changes requested):
I missed these so adding them here and will resolve in a sec.

nlopes (approved):

sheldonhull:
thank you for this! I'm not seeing a release containing it, so do you have a general timeframe @nlopes you plan on cutting a release? If I missed something on this already being said, my apologies 🙇🏻

nlopes:
> , so do you have a general timeframe @nlopes you plan on cutting a release

I intend to cut a breaking change release this coming weekend, stay tuned 🙏

sheldonhull:
Quick follow-up. I'm trying to use this to handle multiple generated images being uploaded at once, but I think the part I'm missing for this to be useful is `uploadToURL` which is private. 

Is there further changes I'm missing for `uploadToURL` being visible? I think as I'm working through my scenario to upload all at once into a single message it won't work without this.

CleanShot 2025-04-29 at 14 17 57


If this is intended to stay private, I can do the steps myself, but with all the release changes in preview, I wanted to follow-up in the context where change was made.

PR #1380, opened 2025-02-25
Additional field to the File structure

rntk:
add Headers field to support headers of an email file attachment.

Example: 
```
"headers": {
    "date": "Thu, 23 Jan 2025 18:09:25 +0100",
    "in_reply_to": null,
    "reply_to": null,
    "message_id": "<email_provider_message_id>"
},
```

related to https://github.com/slack-go/slack/pull/1370

### DIFF
--- a/files.go
+++ b/files.go
@@ -90,10 +90,11 @@ type File struct {
 	IsStarred       bool     `json:"is_starred"`
 	Shares          Share    `json:"shares"`
 
-	Subject string              `json:"subject"`
-	To      []EmailFileUserInfo `json:"to"`
-	From    []EmailFileUserInfo `json:"from"`
-	Cc      []EmailFileUserInfo `json:"cc"`
+	Subject string                 `json:"subject"`
+	To      []EmailFileUserInfo    `json:"to"`
+	From    []EmailFileUserInfo    `json:"from"`
+	Cc      []EmailFileUserInfo    `json:"cc"`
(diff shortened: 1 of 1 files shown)


files.go
```diff
@@ -90,10 +90,11 @@ type File struct {
...
 
-	Subject string              `json:"subject"`
-	To      []EmailFileUserInfo `json:"to"`
-	From    []EmailFileUserInfo `json:"from"`
-	Cc      []EmailFileUserInfo `json:"cc"`
+	Subject string                 `json:"subject"`
+	To      []EmailFileUserInfo    `json:"to"`
+	From    []EmailFileUserInfo    `json:"from"`
+	Cc      []EmailFileUserInfo    `json:"cc"`
+	Headers map[string]interface{} `json:"headers"`
```
nlopes: I think it's probably worth doing a Headers type instead which contains:

- `date`
- `in_reply_to`
- `message_id`
- `reply_to`


I know a map[string]interface{} will just catch everything but I'd rather not.

nlopes (changes requested):

files.go
rntk: Fixed. I have verified the type in the Java SDK at https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-model/src/main/java/com/slack/api/model/File.java#L303. It appears that a struct with only four fields may suffice

nlopes (approved):

PR #1383, opened 2025-03-09
Return a StatusCodeError when a workspace's message limit is exceeded

greysteil:
Currently, this library throws a `strconv.NumError` when attempting to post messages to a Slack workspace that has exceeded its message limit. It does so because Slack returns a 429 without a `Retry-After` header for those requests.

This PR improves the situation by instead returning a `StatusCodeError`. I couldn't see any obvious way to provide a more detailed error without creating a new error class. To only additional information returned from Slack for these errors is a body that reads `message_limit_exceeded`.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -306,7 +306,7 @@ func timerReset(t *time.Timer, d time.Duration) {
 }
 
 func checkStatusCode(resp *http.Response, d Debug) error {
-	if resp.StatusCode == http.StatusTooManyRequests {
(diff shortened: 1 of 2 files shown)


nlopes (approved):
Thanks @greysteil! This should go out with 0.17 which hopefully I'll release soon.

nlopes:
@greysteil I think you'll need to change the testing endpoint as `websocket2` is already in use by another test.

greysteil:
@nlopes - updated!

PR #1385, opened 2025-03-15
Add rich_text_value into BlockAction

borisputerka:
Hello, when submitting modal, we would like to parse contents of `RichTextInput` block from `viewState`. Not really sure whether there is a way now to do it, but it appears to me that it is missing.


nlopes (approved):

nlopes:
Thank you @borisputerka 🙏

vuongvm812:
Could you tell me when this will be released? My project is depending on this feature.

PR #1391, opened 2025-03-22
Fix message event struct

nlopes:
This has been the source of many bugs.

Slack's docs say that a 'message event' contains a 'message' which is the same type as other representations of a message in slack. However, the library has previously represented this as a nested MessageEvent which is recursively nested. That creates lots of bugs, as you can never know if you're looking at the outer MessageEvent (which contains the event information) or the inner MessageEvent (which contains the message itself).

Instead, we can use our canonical Message instead.

Note that we also have to add a `root` onto Message, as this is populated for thread_broadcast events (for legacy reasons).

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -17,6 +17,10 @@ type Message struct {
 	Msg
 	SubMessage      *Msg `json:"message,omitempty"`
 	PreviousMessage *Msg `json:"previous_message,omitempty"`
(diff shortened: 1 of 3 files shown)


nlopes:
I've tried a few types of events, this seems to be populated correctly.

It's a pretty big change so I'm likely to do an rc release before the prod release.

PR #1393, opened 2025-03-22
feat: add '&debug_reconnects=true' to websocket url

### LINKED ISSUE #1379: Add debug_reconnects to socketmode client
### Description
Hello Team,
thanks for this beautiful project! It would be nice to have an option to add `&debug_reconnects=true` to the WebSocket URL. Here is how it described in the docs: 

> One helpful trick: you can append &debug_reconnects=true to your WebSocket URL when you connect to it in order to make the connection time significantly shorter (360 seconds). That way, you can test and debug reconnects without waiting around.

What do you think?

https://api.slack.com/apis/socket-mode

nlopes:
According to the documentation at https://api.slack.com/apis/socket-mode you can append &debug_reconnects=true to your WebSocket URL when you connect to it in order to make the connection time significantly shorter (360 seconds). That way, you can test and debug reconnects without waiting around.

We're adding this whenever debug mode is enabled in the socketmode client.

Closes #1379.

### DIFF
--- a/examples/socketmode/socketmode.go
+++ b/examples/socketmode/socketmode.go
@@ -141,6 +141,8 @@ func main() {
 				}
 
 				client.Ack(*evt.Request, payload)
+			case socketmode.EventTypeHello:
+				client.Debugf("Hello received!")
 			default:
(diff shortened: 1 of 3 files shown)


AyumuKasuga:
Thank you!

PR #1401, opened 2025-03-23
Add some missing admin.conversations endpoints

samstarling:
This PR introduces three new methods for:

* `admin.conversations.setTeams`
* `admin.conversations.convertToPrivate`
* `admin.conversations.convertToPublic`

I've left some comments in-line below.

### DIFF
--- a/admin_conversations.go
+++ b/admin_conversations.go
@@ -1,0 +1,76 @@
+package slack
+
+import (
+	"context"
+	"net/url"
+	"strconv"
+	"strings"
+)
+
+// Set the workspaces in an Enterprise Grid organisation that connect to a public or
+// private channel.
+// See: https://api.slack.com/methods/admin.conversations.setTeams
+func (api *Client) AdminConversationsSetTeams(ctx context.Context, channelID string, orgChannel *bool, targetTeamIDs *[]string, teamID *string) error {
+	values := url.Values{
+		"token":      {api.token},
+		"channel_id": {channelID},
+	}
+
+	if orgChannel != nil {
+		values.Add("org_channel", strconv.FormatBool(*orgChannel))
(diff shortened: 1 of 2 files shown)


admin_conversations.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if teamID != nil {
+		values.Add("team_id", *teamID)
+	}
+
+	response := &SlackResponse{}
+	err := api.postMethod(ctx, "admin.conversations.setTeams", values, response)
+	if err != nil {
+		return err
+	}
```
samstarling, line 44: In `admin.go`, other functions use `api.adminRequest` — however, that assumes a prefix of `https://%s.slack.com/api/users.admin.%s?t=%d`, as defined in the (badly named, imo) `WEBAPIURLFormat`. I think it's OK to just use `postMethod` here.

admin_conversations.go
```diff
@@ -0,0 +1,76 @@
...
+	"context"
+	"net/url"
+	"strconv"
+	"strings"
+)
+
+// Set the workspaces in an Enterprise Grid organisation that connect to a public or
+// private channel.
+// See: https://api.slack.com/methods/admin.conversations.setTeams
+func (api *Client) AdminConversationsSetTeams(ctx context.Context, channelID string, orgChannel *bool, targetTeamIDs *[]string, teamID *string) error {
```
samstarling: I've named these exactly the same as the endpoint — and because they don't have non-`ctx` equivalents, I have no included `...Context` at the end of the function name.

nlopes:
Would love to have some tests here. I know `admin.go` doesn't have any but would be neat to have tests on this one.

nlopes (approved):

PR #1403, opened 2025-03-23
Handle nil being passed in to newTextParser

samstarling:
The `dst` variable was flagged as being unused here. In `newJSONParser`, we check if `dst == nil`: this change copies that behaviour in `newTextParser`.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -336,6 +336,10 @@ func newJSONParser(dst interface{}) responseParser {
 
 func newTextParser(dst interface{}) responseParser {
 	return func(resp *http.Response) error {
+		if dst == nil {
+			return nil
+		}
+
 		b, err := io.ReadAll(resp.Body)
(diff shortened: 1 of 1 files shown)


nlopes (approved):

misc.go
```diff
@@ -336,6 +336,10 @@ func newJSONParser(dst interface{}) responseParser {
 
 func newTextParser(dst interface{}) responseParser {
 	return func(resp *http.Response) error {
+		if dst == nil {
```
nlopes, line 339: I think the difference, and I'm not quite sure I grok if there will be issues here, is that in `newJSONParser` we must use `dst` but in this one we don't.

So _technically_ (from the pov of this function), this changes the behaviour and expectation here.

I still think this is ok - there's no call in the codebase I can find that means `dst` will ever be nil here.

PR #1404, opened 2025-03-23
Tidy up arguments to _search function

samstarling:
The `search.all`, `search.files` and `search.messages` endpoints all take the same parameters — so the booleans being passed here are redundant: they are not used for anything, nor can they be. This change removes them, and tidies up some redundant boolean logic in the handling of `params.Highlight`.

### DIFF
--- a/search.go
+++ b/search.go
@@ -89,7 +89,7 @@ func NewSearchParameters() SearchParameters {
 	}
 }
 
-func (api *Client) _search(ctx context.Context, path, query string, params SearchParameters, files, messages bool) (response *searchResponseFull, error error) {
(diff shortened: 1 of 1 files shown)


search.go
```diff
@@ -103,7 +103,7 @@ func (api *Client) _search(ctx context.Context, path, query string, params Searc
 	if params.SortDirection != DEFAULT_SEARCH_SORT_DIR {
 		values.Add("sort_dir", params.SortDirection)
 	}
-	if params.Highlight != DEFAULT_SEARCH_HIGHLIGHT {
+	if params.Highlight {
```
samstarling, line 106: `DEFAULT_SEARCH_HIGHLIGHT` is `false`, so this is saying `if params.Highlight != false`, which is the same as `if params.Highlight`.

nlopes (approved):

PR #1405, opened 2025-03-23
Tidy up go.mod, add a linting check to prevent this

samstarling:
I noticed `go mod tidy` produced a diff, so I have tidied it up. I've also added a step to the "lint" GitHub actions which verifies that `go mod tidy` does not result in any changes.


nlopes (approved):
Lovely!

PR #1406, opened 2025-03-23
Fix issues with flaky tests

samstarling:
You couldn't run `go test -count=100 -failfast` because the use of `http.HandleFunc` meant that tests were sharing global state — to get around this, I've introduced `startServer()` which uses `mux` to isolate each test. This threw up some interesting cases, including some tests which relied on test data set up by previous tests.

To test this, you can run:

```
go test -shuffle=on -count=10 -failfast
```

### DIFF
--- a/apps_test.go
+++ b/apps_test.go
@@ -7,8 +7,10 @@ import (
 )
 
 func TestListEventAuthorizations(t *testing.T) {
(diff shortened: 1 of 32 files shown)


samstarling:
OK, something funky going on with this, I'll take another pass soon!

nlopes:
Let me know when ready for a review.

PR #1412, opened 2025-04-23
RichTextSectionLinkElement Text omitempty

bassettb:
The `text` field in a `RichTextSectionLinkElement` is optional.  Sending an empty string results in a Slack error.  The field needs to be omitted if it is unused.

https://api.slack.com/reference/block-kit/blocks#link-element-type


nlopes:
Thank you 🙏

nlopes (approved):

Issue #1434, opened 2025-05-29
Adapt CompleteUploadExternalParameters to accept a `blocks` element


dalyndalton:
### Description

Currently the `UploadFileV2Context` function nor the `CompleteUploadExternalContext` support passing a list of blocks.  The slack documentation also specifies that this field is to be ignored if InitialComment is provided.
 
### (Optional) Slack's documentation
https://api.slack.com/methods/files.completeUploadExternal#arg_blocks

PR #1435, opened 2025-05-29
feat: Add missing Blocks to UploadFileContextV2

dalyndalton:
see https://github.com/slack-go/slack/issues/1434

This adds the missing optional dependency for Blocks to the upload file method, allowing users to use rich blocks and buttons during the file upload process.

Confirmed that new tests are passing, as well as the Blocks functioning correctly in the app I'm currently developing.

### DIFF
--- a/files.go
+++ b/files.go
@@ -175,6 +175,7 @@ type UploadFileV2Parameters struct {
 	Filename        string
 	Title           string
 	InitialComment  string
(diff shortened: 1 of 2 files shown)


nlopes (approved):

nlopes:
Thank you @dalyndalton

PR #1439, opened 2025-06-04
support Data Access API

nktks:
- support [Data Access API](https://api.slack.com/docs/apps/data-access-api#event-listener)
  - added https://api.slack.com/methods/assistant.search.context api
  - added `assistant_thread` attribute to Event API `message`, `app_mention` envets https://api.slack.com/docs/apps/data-access-api#event-listener
- We are still testing this changes for `Agent & AI Apps` enabled app.

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/examples/socketmode_handler/socketmode_handler.go
+++ b/examples/socketmode_handler/socketmode_handler.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"os"
@@ -56,6 +57,8 @@ func main() {
 
 	// Handle a specific event from EventsAPI
 	socketmodeHandler.HandleEvents(slackevents.AppMention, middlewareAppMentionEvent)
+	socketmodeHandler.HandleEvents(slackevents.Message, middlewareMessageEvent)
+	socketmodeHandler.HandleEvents(slackevents.AssistantThreadStarted, middlewareAssistantThreadStartedEvent)
 
 	//\\ EventTypeInteractive //\\
 	// Handle all Interactive Events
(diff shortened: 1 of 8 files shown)


nlopes:
@nktks I'm not going to review this until you move it into ready for review - just wanted to call that out explicitly. Thanks for adding this.

nlopes:
@nktks just checking in on this - no pressure at all but want to make sure you didn't just leave it at Draft by mistake :pray:

nktks:
@nlopes Sorry for the confusion.
I've requested for the Data Access API to Slack, but it hasn't been approved yet for long time, so I'm still waiting....
Because of this, I haven't been able to test it....

nlopes:
~Ah if that's the part that's missing, I can test it for you, no problem at all.~

Realised that developer sandbox environments don't have this enabled :(

nktks:
Yes.
https://api.slack.com/docs/apps/data-access-api seems to be limited access stage.
My company has joined Slack Enterprise Grid and I requested from https://api.slack.com/forms/data_access_api

nktks:
@nlopes Hello.
Our app was approved and I could test thorough https://github.com/slack-go/slack/pull/1439/files#diff-03269bab64cd2cd7188b80de22221ee1321f6ee37f70d909a40e67205b724c24 .

So I changed as ready for review this PR.
Could you please take a look when you have time?

examples/socketmode_handler/socketmode_handler.go
```diff
@@ -56,6 +57,8 @@ func main() {
 
 	// Handle a specific event from EventsAPI
 	socketmodeHandler.HandleEvents(slackevents.AppMention, middlewareAppMentionEvent)
+	socketmodeHandler.HandleEvents(slackevents.Message, middlewareMessageEvent)
```
nlopes: I would prefer if this was an example on its own. That's to make sure that people that don't have this API yet, won't be affected if they want to run this example (basically it will save me answering queries about why this example might not be working).

nlopes (changes requested):

examples/socketmode_handler/socketmode_handler.go
nktks: Thanks for your review.
I divided the exapmle.

[d4221e7](https://github.com/slack-go/slack/pull/1439/commits/d4221e7659559f291d92ac9321f7609c50223eb0)
[e044f7a](https://github.com/slack-go/slack/pull/1439/commits/e044f7a80713e466893a0d2b4d44dd79660114bd)

nlopes (approved):

PR #1444, opened 2025-06-08
Support Slack Files in ImageBlockElement

### LINKED ISSUE #1438: Support `slack_file` in ImageBlockElement
### Description
`UploadFileV2Context`/`UploadFileV2` (which replace the deprecated `UploadFileContext` / `UploadFile`) only return [Slack File IDs](https://api.slack.com/reference/block-kit/composition-objects#slack_file), but the Slack Go SDK currently does not support their in Image Block Elements.

Note: the SDK already does support them in Image Blocks since #1311, so this shouldn't be a crazy lift.

https://api.slack.com/reference/messaging/blocks#image

superturkey650:
Supporting slack_file in ImageBlockElement: 1438
- Add support for slack_file to ImageBlockElement
- update tests involving ImageBlockElement where ImageURL is set as a function arg
- fix support for slack_file in ImageBlock
- update tests involving ImageBlock where ImageURL is set as a function arg
- Update examples for both ImageBlock and ImageBlockElement

This does change the function signature of both NewImageBlockElement and NewImageBlock. If you would instead prefer slack-file specific versions of each of these functions so that the existing function signature can remain the same, let me know.

### DIFF
--- a/block_element.go
+++ b/block_element.go
@@ -117,9 +117,10 @@ func (s UnknownBlockElement) ElementType() MessageElementType {
 //
 // More Information: https://api.slack.com/reference/messaging/block-elements#image
 type ImageBlockElement struct {
-	Type     MessageElementType `json:"type"`
-	ImageURL string             `json:"image_url"`
-	AltText  string             `json:"alt_text"`
+	Type      MessageElementType `json:"type"`
+	ImageURL  string             `json:"image_url"`
+	AltText   string             `json:"alt_text"`
+	SlackFile *SlackFileObject   `json:"slack_file,omitempty"`
 }
 
 // ElementType returns the type of the Element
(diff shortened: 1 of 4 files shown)


nlopes:
Thank you @superturkey650. Would you mind adjusting the following:
- add a title
- rebase and squash the commits
- keep the signature for now and add a slack file specific (but add a TODO with the words BREAKING CHANGE so that when I do a breaking change release, I can clean this up)

I think at this stage, I'm not going for another BREAKING CHANGE release unless I really need to.

superturkey650:
Added a title, squashed the commits, and reverted the breaking change in favor of a new function just for slack files.

nlopes (approved):

PR #1461, opened 2025-07-09
Make it possible to omit fields in the manifest Settings struct

### LINKED ISSUE #1460: It's Impossible to Omit the EventSubscriptions and Interactivity fields in Manifest's Settings Struct
### What happened

Currently the Settings struct in manifest.go defines both `EventSubscriptions` and `Interactivity` as struct values rather than pointers, which means omitempty will never take effect.

https://github.com/slack-go/slack/blob/e29b7e304998262f189737ff254c7b539db7fa60/manifests.go#L189-L195

This results in an empty `EventSubscriptions` or `Interactivity` object being passed to Slack as an empty JSON object that is impossible to get rid of and requires you to work around it by doing things like setting an `EventSubscriptions` object even when you're not using oauth scopes that would require these fields.

### Expected behavior

These should be pointers and be omitted when serializing to JSON

### Steps to Reproduce

- Create a Slack app without scopes for `EventSubscriptions`
- Create a `slack.Manifest{}` object without setting `EventSubscriptions`
- Call `CreateManifestContext(ctx, manifestStruct, your config token)`
- See the error claiming you have an invalid manifest because your `EventSubscriptions`.RequestUrl is invalid
- Curse quietly
- Print the object in logging and see the struct definition
- Open this issue
- Open a PR to fix it.

### Versions
- Go: 1.24
- slack-go/slack: v0.17.3

pmarkowsky:
This PR changes the `Interactivity` and `EventSubscriptions` fields in the manifest.go `Settings`  struct to pointers so they can be omitted if they're left empty.

This fixes #1460

### DIFF
--- a/manifests.go
+++ b/manifests.go
@@ -187,11 +187,11 @@ type Display struct {
 
 // Settings is a group of settings corresponding to the Settings section of the app config pages.
 type Settings struct {
-	AllowedIPAddressRanges []string           `json:"allowed_ip_address_ranges,omitempty" yaml:"allowed_ip_address_ranges,omitempty"`
-	EventSubscriptions     EventSubscriptions `json:"event_subscriptions,omitempty" yaml:"event_subscriptions,omitempty"`
-	Interactivity          Interactivity      `json:"interactivity,omitempty" yaml:"interactivity,omitempty"`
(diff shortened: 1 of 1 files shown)


nlopes:
Thank you!

nlopes (approved):

PR #1483, opened 2025-10-01
Add support for title argument in CreateChannelCanvas method

### LINKED ISSUE #1482: Add support for title parameter in conversations.canvases.create method
### Description
[conversations.canvases.create](https://docs.slack.dev/reference/methods/conversations.canvases.create/) method now supports passing `title` argument. Add support for the argument to the `CreateChannelCanvas` method.

https://docs.slack.dev/reference/methods/conversations.canvases.create/

jarospisak-unity:
This PR adds support for `title` argument in the `CreateChannelCanvas` method.

Closes #1482.

### DIFF
--- a/conversation.go
+++ b/conversation.go
@@ -1033,24 +1033,55 @@ func (api *Client) MarkConversationContext(ctx context.Context, channel, ts stri
 	return response.Err()
 }
 
+// createChannelCanvasParams contains arguments for CreateChannelCanvas method call.
+type createChannelCanvasParams struct {
+	title           string
+	documentContent *DocumentContent
+}
+
+// CreateChannelCanvasOption options for the CreateChannelCanvas method call.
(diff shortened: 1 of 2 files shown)


nlopes:
Hi @jarospisak-unity. As it stands this is a breaking change and I'd rather not add breaking changes if they can be avoided at all.

One way is simply to add a method that allows to pass options. This pattern already exists in other parts of the codebase. Happy for you to give it a go if you're ok to do so, alternatively, I'll come back to this soon-ish.

```
func (api *Client) CreateChannelCanvasWithOptions(channel string, options
  ...CreateChannelCanvasOption) (string, error) {
      return api.CreateChannelCanvasWithOptionsContext(context.Background(), channel,
  options...)
  }
[...]
```

jarospisak-unity:
@nlopes thanks for the feedback, I wasn't aware of that pattern. I'll refactor this pull request.

PR #1499, opened 2025-11-14
feat: add support for workflow_button as a block element

nlopes:
[Workflow Button Element](https://docs.slack.dev/reference/block-kit/block-elements/workflow-button-element)


block_element_test.go
```diff
@@ -361,3 +361,115 @@ func TestIconButtonJSONMarshalling(t *testing.T) {
 	assert.Equal(t, "delete_button_1", unmarshalled.ActionID)
 	assert.Equal(t, "delete_item", unmarshalled.Value)
 }
+
+func TestNewWorkflowButtonBlockElement(t *testing.T) {
```
nlopes, line 365: Comment for myself: Claude coded! Need to do a review on these tests to make sure they aren't silly.

Issue #1485, opened 2025-10-24
[Feature Request] Support Streaming messages


yongzhang:
### (Optional) Slack's documentation
Ref: https://docs.slack.dev/changelog/2025/10/7/chat-streaming/

Thanks.

dnck:
@nlopes I would nominate myself for work on this feature if you don't already have it in your current scope of work. Happy to lend a hand in this awesome project.

nlopes:
Happy for you to do so. I intend to work on it over the weekend so if you get there first go for it. If not, I'll let you know here before I start work on it.

PR #1506, opened 2025-12-02
Adds Chat Streaming API Support

k-k:
Implements Slack's new [Chat Streaming APIs](https://docs.slack.dev/changelog/2025/10/7/chat-streaming/) for #1485.

I was looking to use these new APIs for a project and when looking through `chat.go`, this actually seemed pretty straight forward, following the existing conventions. 

`MsgOptionMarkdownText` was the one piece that felt a bit odd since this was intentionally handled somewhat differently for `PostMessage`, etc - so, happy to make changes here. The difference and why I included it as I did was that the [Streaming APIs](https://docs.slack.dev/reference/methods/chat.appendStream#arguments) _only_ supports the `markdown_text` attribute.

### DIFF
--- a/chat_test.go
+++ b/chat_test.go
@@ -493,7 +493,6 @@ func TestStartStream(t *testing.T) {
 				rw.Write(response)
 			})
 
-			// Fixed: removed timestamp parameter, pass via MsgOptionTS
 			_, _, _ = api.StartStream("CXXX", test.opt...)
 		})
 	}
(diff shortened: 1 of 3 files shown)


chat_test.go
```diff
@@ -425,3 +425,207 @@ func TestUpdateMessage(t *testing.T) {
...
+					Channel:   "CXXX",
+					Timestamp: "1234567890.123456",
+					SlackResponse: SlackResponse{
+						Ok: true,
+					},
+				})
+				rw.Write(response)
+			})
+
+			// Fixed: removed timestamp parameter, pass via MsgOptionTS
```
nlopes: ✂️

nlopes (approved):

PR #1519, opened 2026-01-29
feat: Add `focus_on_load` support for remaining block elements

### LINKED ISSUE #1518: Add support for `focus_on_load` for missing elements
### Description

👋 There's partial support for the `focus_on_load` parameter already implemented for some block elements.

These elements already implement support:
* `DateTimePickerBlockElement`
* `EmailTextInputBlockElement`
* `URLTextInputBlockElement`
* `RichTextInputBlockElement`

These additional block elements also have support for this, but it isn't implemented yet:
* `DatePickerBlockElement` ([docs](https://docs.slack.dev/reference/block-kit/block-elements/date-picker-element))
* `TimePickerBlockElement` ([docs](https://docs.slack.dev/reference/block-kit/block-elements/time-picker-element))
* `PlainTextInputBlockElement` ([docs](https://docs.slack.dev/reference/block-kit/block-elements/plain-text-input-element))
* `CheckboxGroupsBlockElement` ([docs](https://docs.slack.dev/reference/block-kit/block-elements/checkboxes-element))
* `RadioButtonsBlockElement` ([docs](https://docs.slack.dev/reference/block-kit/block-elements/radio-button-group-element))
* `NumberInputBlockElement` ([docs](https://docs.slack.dev/reference/block-kit/block-elements/number-input-element))
* `SelectBlockElement` ([docs](https://docs.slack.dev/reference/block-kit/block-elements/select-menu-element))
* `MultiSelectBlockElement` ([docs](https://docs.slack.dev/reference/block-kit/block-elements/multi-select-menu-element))

CGA1123:
This adds `focus_on_load` to the remaining Slack BlockKit Block Elements that support it:
* `{static,external,users,conversations,channels}_select`
* `multi_{static,external,users,conversations,channels}_select`
* `datepicker`
* `timepicker`
* `plain_text_input`
* `checkboxes`
* `radio_buttons`
* `number_input`

Closes: #1518

### DIFF
--- a/block_element.go
+++ b/block_element.go
@@ -248,6 +248,7 @@ type SelectBlockElement struct {
 	Filter                       *SelectBlockElementFilter `json:"filter,omitempty"`
 	MinQueryLength               *int                      `json:"min_query_length,omitempty"`
 	Confirm                      *ConfirmationBlockObject  `json:"confirm,omitempty"`
+	FocusOnLoad                  bool                      `json:"focus_on_load,omitempty"`
 }
 
 // SelectBlockElementFilter allows to filter select element conversation options by type.
@@ -339,6 +340,7 @@ type MultiSelectBlockElement struct {
 	MinQueryLength       *int                      `json:"min_query_length,omitempty"`
 	MaxSelectedItems     *int                      `json:"max_selected_items,omitempty"`
 	Confirm              *ConfirmationBlockObject  `json:"confirm,omitempty"`
+	FocusOnLoad          bool                      `json:"focus_on_load,omitempty"`
 }
 
 // ElementType returns the type of the Element
@@ -459,6 +461,7 @@ type DatePickerBlockElement struct {
 	Placeholder *TextBlockObject         `json:"placeholder,omitempty"`
 	InitialDate string                   `json:"initial_date,omitempty"`
 	Confirm     *ConfirmationBlockObject `json:"confirm,omitempty"`
+	FocusOnLoad bool                     `json:"focus_on_load,omitempty"`
 }
 
 // ElementType returns the type of the Element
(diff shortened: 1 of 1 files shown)


block_element.go
```diff
@@ -248,6 +248,7 @@ type SelectBlockElement struct {
 	Filter                       *SelectBlockElementFilter `json:"filter,omitempty"`
 	MinQueryLength               *int                      `json:"min_query_length,omitempty"`
 	Confirm                      *ConfirmationBlockObject  `json:"confirm,omitempty"`
+	FocusOnLoad                  bool                      `json:"focus_on_load,omitempty"`
```
CGA1123, line 251: Added these as `bool` rather than `*bool`.

Although these are always optional fields in for the Slack API, the existing support already uses `bool` for this field. Doesn't seem worth adding that confusion or breaking existing usages!

nlopes: > Although these are always optional fields in for the Slack API, the existing support already uses bool for this field. Doesn't seem worth adding that confusion or breaking existing usages!

I'm not sure I completely follow this bit. Would you mind explaining a bit more?

At a glance this PR seems totally fine.

CGA1123: Because we use `bool` + `omitempty` we can only model the "unspecified" and `true` state for `focus_on_load`. It's impossible to send an explicit `false`. Which is completely fine as long as Slack always keeps `false` as the default for `bool` fields in their APIs (which is the case for `focus_on_load`).

Just is one of those things that might cause weird random annoying issues in the future if it's accidentally replicated to another similar but not really _that_ similar case in the future 😄

nlopes: Ah gotcha, I see what you meant now. 🙏

PR #1520, opened 2026-01-31
feat(admin): Implement missing admin.roles.* API method:

nicolas-carlier-ls:
Implements support for missing of admin.roles.* API methods

- admin.roles.listAssignments
- admin.roles.addAssignments
- admin.roles.removeAssignments

##### Pull Request Guidelines

##### PR preparation

##### API changes

### DIFF
--- a/admin_roles.go
+++ b/admin_roles.go
@@ -66,9 +66,11 @@ func (api *Client) AdminRolesAddAssignments(ctx context.Context, params AdminRol
 
 // AdminRolesListAssignmentsParams contains arguments for AdminRolesListAssignments method call.
 type AdminRolesListAssignmentsParams struct {
-	RoleIDs []string
-	Limit   int
-	Cursor  string
+	RoleIDs       []string
+	EntityIDs     []string
+	Limit         int
+	Cursor        string
+	SortDirection string
 }
 
 // RoleAssignment represents a single role assignment.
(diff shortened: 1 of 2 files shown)


admin_roles.go
```diff
@@ -0,0 +1,156 @@
...
+	"strconv"
+	"strings"
+)
+
+// AdminRolesAddAssignmentsParams contains arguments for AdminRolesAddAssignments method call.
+type AdminRolesAddAssignmentsParams struct {
+	RoleID        string
+	EntityIDs     []string
+	UserIDs       []string
+	DateEffective int64
```
nlopes: Is this documented anywhere? None of the official libraries have it.

I also can't find this in the docs so would be great to understand where I can read about it (if it exists anywhere).

admin_roles.go
```diff
@@ -0,0 +1,156 @@
...
+	err := api.postMethod(ctx, "admin.roles.addAssignments", values, response)
+	if err != nil {
+		return nil, err
+	}
+
+	return response, response.Err()
+}
+
+// AdminRolesListAssignmentsParams contains arguments for AdminRolesListAssignments method call.
+type AdminRolesListAssignmentsParams struct {
```
nlopes: This should have EntityIDs and SortDir as well.

admin_roles.go
```diff
@@ -0,0 +1,156 @@
...
+	Limit   int
+	Cursor  string
+}
+
+// RoleAssignment represents a single role assignment.
+type RoleAssignment struct {
+	RoleID      string `json:"role_id"`
+	EntityID    string `json:"entity_id,omitempty"`
+	UserID      string `json:"user_id,omitempty"`
+	DateCreated int64  `json:"date_created,omitempty"`
```
nlopes: This is date_create (per docs and official library); really hoping the typo is here not on their docs/libraries 😆

nlopes (changes requested):
There are a few tweaks needed before I merge this in. @nicolas-carlier-ls if you prefer, I can adjust this and ship, alternatively I'll leave the tweaks to you.

admin_roles.go
nlopes: I've removed it to merge this - can always add later.

nlopes (approved):

PR #1533, opened 2026-02-22
chore: use golangci-lint in Makefile

olegbespalov:
## What?

Aligns linter with CI.

Use the binary when present in PATH; otherwise, run via “go run” so no install is required.

## Why?

So local lint matches CI, and contributors catch the same issues before pushing, without needing to install golangci-lint.


nlopes (approved):
Thank you!

PR #1537, opened 2026-03-18
Add WorkflowID to message

clayallsopp:
Hey team, this adds the `workflow_id` to the message struct.

I can't find this field documented anywhere, but it is on the messages we see that originate from Workflows - let me know if there's something different/else you'd need!

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -87,6 +87,7 @@ type Msg struct {
 	Username   string      `json:"username,omitempty"`
 	Icons      *Icon       `json:"icons,omitempty"`
 	BotProfile *BotProfile `json:"bot_profile,omitempty"`
+	WorkflowID string      `json:"workflow_id,omitempty"`
 
 	// channel_join, group_join
 	Inviter string `json:"inviter,omitempty"`
--- a/messages_test.go
+++ b/messages_test.go
@@ -285,6 +285,31 @@ var botMessage = `{
     "ts": "1358877455.000010"
 }`
 
+var workflowBotMessage = `{
+    "type": "message",
+    "subtype": "bot_message",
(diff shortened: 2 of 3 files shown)


nlopes:
Hey @clayallsopp, thanks for sending. 

> I can't find this field documented anywhere

Honestly, me neither. I tend to avoid adding elements if none of the slack docs + canonical libraries from slack have it unless I know clearly why they weren't added/should be added.

Is this one that you need as part of some specific need? If so, would love to understand your use case. Adding something is (usually) a non-breaking change but removing is so want to make sure what I add is valuable (especially if undocumented)

clayallsopp:
> Hey @clayallsopp, thanks for sending.
> 
> > I can't find this field documented anywhere
> 
> Honestly, me neither. I tend to avoid adding elements if none of the slack docs + canonical libraries from slack have it unless I know clearly why they weren't added/should be added.
> 
> Is this one that you need as part of some specific need? If so, would love to understand your use case. Adding something is (usually) a non-breaking change but removing is so want to make sure what I add is valuable (especially if undocumented)

@nlopes In short, we have logic for some internal bots that requires approval on a per-workflow basis for visibility/auditing (related to workflows that interact with AI systems). It seems like the Bot IDs are not consistent across invocations of a Workflow, so those are not reliable.

As an alternative - is possible to expose more raw API results to enable some access to these types of fields? as in, the raw JSON maps rather than the structs?

nlopes:
Gotcha. I am going to add this. (I'm going to follow-up with a PR for adding TriggerID as well in a bit.)

nlopes (approved):

PR #1547, opened 2026-04-06
fix: move attachment and block serialization to formSender

moabukar:
## What

Fixes the `FIXME` in `MsgOptionAttachments` - attachments and blocks were being serialized twice (once for the JSON path, once for the form path). Same issue existed in `MsgOptionBlocks`.

## How

Moved form-path serialization into `formSender.BuildRequestContext` instead of doing it eagerly in the option functions. Now each sender handles its own serialization:
- `responseURLSender` → reads typed struct fields (unchanged)
- `formSender` → marshals into form values at send time

## Worth noting

- `UnsafeApplyMsgOptions` returns `config.values` directly — after this change, `attachments`/`blocks` won't be in there since marshaling is deferred. That function is documented as unsupported though.
- `MsgOptionMetadata` has the same pattern — left it out to keep scope small, happy to follow up.

## Tests

All existing tests pass. Covered by `TestPostMessage` (attachments, blocks, empty blocks) and `TestPostMessageWithBlocksWhenMsgOptionResponseURLApplied`.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -375,13 +375,15 @@ func (t sendConfig) BuildRequestContext(ctx context.Context, token, channelID st
 			deleteOriginal:  t.deleteOriginal,
 		}.BuildRequestContext(ctx)
 	default:
-		return formSender{endpoint: t.endpoint, values: t.values}.BuildRequestContext(ctx)
+		return formSender{endpoint: t.endpoint, values: t.values, attachments: t.attachments, blocks: t.blocks}.BuildRequestContext(ctx)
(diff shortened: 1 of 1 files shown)


nlopes (approved):
Looks good, thank you 🙏

moabukar:
> Looks good, thank you 🙏 

Appreciated! Wanted to catchup, do you have 5/10 mins to spare please? my email is devopsbymo@gmail.com @nlopes

PR #1565, opened 2026-06-11
feat: unmarshal is_ext_shared_channel field in EventsAPIEvent

### LINKED ISSUE #1564: Expose IsExtSharedChannel on EventsAPIEvent
https://docs.slack.dev/apis/events-api/#callback-field lists this as a field on outer events, but the `EventsAPIEvent` object doesn't include it. It'd be useful to have this field on the struct for logic around ext channels.

I'll make a PR for this!

jjiang-stripe:
Fixes https://github.com/slack-go/slack/issues/1564

Adds `IsExtSharedChannel` to the `EventsAPIEvent` struct so it gets unmarshalled from JSON. This is a field on the outer event as described by https://docs.slack.dev/apis/events-api/#callback-field and I would like to be able to write some logic around this field after calling `slackevents.ParseEvent`.

### DIFF
--- a/slackevents/outer_events.go
+++ b/slackevents/outer_events.go
@@ -8,13 +8,14 @@ import (
 
 // EventsAPIEvent is the base EventsAPIEvent
 type EventsAPIEvent struct {
-	Token        string `json:"token"`
-	TeamID       string `json:"team_id"`
-	Type         string `json:"type"`
-	APIAppID     string `json:"api_app_id"`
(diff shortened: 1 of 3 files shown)


nlopes:
This is incomplete as it's missing from the EventCallback. I'm going to merge and follow with a cherry picked commit that fixes it fully as it doesn't seem I can update your branch.

